PR #2006, opened 2013-09-15
Implement the python datasource in the python bindings

nmichaud:
This is mostly copied from the python plugin. The benefit to this approach is that the datasource object can be constructed in the python script and passed into the mapnik api. So, for example, in the concentric-circles example at the bottom of https://github.com/mapnik/mapnik/wiki/Python-Plugin, you can do this instead:

```python
if __name__ == '__main__':
    m = mapnik.Map(640, 320)

    m.background = mapnik.Color('white')
    s = mapnik.Style()
    r = mapnik.Rule()
    r.symbols.append(mapnik.LineSymbolizer())
    s.rules.append(r)
    m.append_style('point_style',s)

    ######## This part is different
    layer = mapnik.Layer('python')
    datasource = CircleDatasource(centre_x=4, centre_y=50)
    layer.datasource = mapnik.Python(datasource)
    mapnik.render(m, mapnik.Image(512, 512))
    ...
    datasource.centre_x = 10
    mapnik.render(m, mapnik.Image(512, 512))
    ...
```

This means that later on you can change the values of centre_x and centre_y on the instance of CircleDatasource and they will be used the next time the map is rendered (this is useful in UIs with mapping widgets).

### DIFF
--- a/bindings/python/mapnik/__init__.py
+++ b/bindings/python/mapnik/__init__.py
@@ -605,17 +605,6 @@ def Osm(**keywords):
     keywords['type'] = 'osm'
     return CreateDatasource(keywords)
 
-def Python(**keywords):
-    """Create a Python Datasource.
-
-    >>> from mapnik import Python, PythonDatasource
-    >>> datasource = Python('PythonDataSource')
-    >>> lyr = Layer('Python datasource')
-    >>> lyr.datasource = datasource
-    """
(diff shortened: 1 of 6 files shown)


springmeyer:
closing, now that python bindings are standalone this might be more reasonable to do, let's track at https://github.com/mapnik/python-mapnik/issues/6.

PR #2203, opened 2014-04-04
Use a std::vector for vertex_vector

### LINKED ISSUE #2200: Extremely high memory usage with big input polygons
I'm seeing up to 18GB of virtual memory used to render tiles with polygons whose total in-memory size is ~200MB in PostGIS.

The memory usage I saw via node-mapnik when about 16 tile requests were sent, but still 200*16 would give 3.2GB...

Consider this a placeholder ticket for now, will update later with more info, once I have a simplified testcase. \cc @springmeyer

strk:
Greatly reduces memory usage with large collections of small
component geometries. Closes #2200.


springmeyer:
Closing: we now using std::vector after merging #2739

Issue #2227, opened 2014-04-28
Behavior of minimum-distance for linear text placements


hollinger:
For linear text placements, minimum-distance used to check distance between repeated labels, i.e. any label with the same text as the one being placed. This behavior was changed during the harfbuzz related work on placement_finder. Now minimum-distance simply checks distance from all other labels.

I am not sure if this change was intentional or not, but the new behavior of minimum-distance for linear text placements is at least more consistent with that of point text placements and shield symbolizer. Checking repeated labels was also a useful capability, however.

Rather than changing minimum-distance behavior back to how it was before, I think it might be beneficial to have a separate attribute to handle repeated labels. Any thoughts about adding a new attribute for repeat-distance?

PR #2331, opened 2014-08-03
Repeat distance for labels

hollinger:
This is a potential solution to problem with minimum-distance and text symbolizer, as mentioned in mapnik/mapnik#2227. A repeat-distance attribute has been added to text properties, which can be applied to both text and shield symbolizer. It is a minimum distance to check only against labels with the same text in the collision detector, while minimum-distance will still check against all other labels. The attributes minimum-distance and repeat-distance can be used independently or in combination with each other. Additional test cases are included.

Note this also impacts group symbolizer, which uses mimimum-distance and repeat-distance in the same way as text and shield symbolizers. Previously, group symbolizer treated the minimum-distance attribute like repeat-distance.


DinleyGitHub:
In the mapnik v3.0, we used the "repeat-distance" to control the repeated lables of  line was nor effective，still dense locally.
I've set "repeat-distance" to max but still no effect, same goes for "minimum-distance".
QQ截图20220425191348

PR #2459, opened 2014-09-30
Add "clip" and "grid" support to PostGIS input plugin

strk:
For "clip", uses ST_ClipToBox2D function from upcoming PostGIS 2.2.

Adds parameters "clip_geometries" and "grid_geometries".
Also adds max resolution parameter for simplify and grid and
min resolution parameter for clip.


strk:
So,  #2481 and #2482 superceed this, this one can be closed in favour of those two others (canit please be moved on?)

PR #2482, opened 2014-10-02
Add geometry_display_expression parameter

strk:
It can be used to run arbitrary functions to preprocess the geometry
to be rendered based on pixel resolution and/or bounding box.

Includes testcase


strk:
Ping... This change is CI-green, has a testcase and after 5 years still sounds like a good idea to me ?
@springmeyer do you still work on Mapnik ? :)

Issue #952, opened 2011-11-16
merge markers and point symbolizer


springmeyer:
A single symbolizer should be able to render any type of file (png or svg), support any kind of placement (centroids or along lines), and should support points being dynamically plotted with a few dynamic options (like fill, stroke, and width). I think - its a lot of functionality to pack in one place, but designed well it should be manageable.

Currently point and markers symbolizer's have grown alongside each other are confusingly similar and different:

Markers supports:
- placing only svg files or dynamically drawing circle's/ellipses
- supports advanced ability to place points along lines (with special collision avoidance)

Points symbolizer supports:
- support reading from png or svg files
- have more traditional placement options targeted at just points and polygons: centroid,point_on_surface/interior

springmeyer:
Next step on this is to solve #2511. After that I think it makes sense to try to merge them. And likely along the way we'll find other issues like #2511.

PR #2549, opened 2014-10-27
Support MarkersSymbolizer in GroupSymbolizer

hollinger:
The motivation of these changes is to support markers symbolizer inside of group symbolizer, so that point symbolizer can be removed without breaking group symbolizer tests. This is a requirement for #952.

Some refactoring of marker handling was needed to avoid a lot of code duplication for marker rendering in group symbolizer. The refactor also helped to eliminate some existing code duplication between renderers for the marker dispatch types.

I also added a new test case to test different types of markers (ellipse, vector, raster, etc) within group symbolizer.


springmeyer:
merged in #2642

Issue #2292, opened 2014-07-03
Hextree PNG quant too aggressive


springmeyer:
Take this 24bit png:

![](https://cloud.githubusercontent.com/assets/126952/3046521/e017f4b4-e129-11e3-8b62-e9d30b78744b.png)

And run it through Mapnik using hextree quant:

``` sh
python -c "import mapnik as m;m.Image.open('input.png').save('out.png','png8:m=h')"
```

and the small blue line in the upper right drops out:

![out](https://cloud.githubusercontent.com/assets/20300/3472098/af1a55a8-02c8-11e4-8966-252cd2446a1a.png)

Leading to an artifact between tiles like:

![](https://camo.githubusercontent.com/ce1cfb5cfe16049f9f27ae624c92ef4028a2221c/68747470733a2f2f696d672e736b697463682e636f6d2f32303132303232392d716b3131387965707166727866686d626b6d66326a326d31622e6a7067)

Another testcase is ![](https://camo.githubusercontent.com/437d9e9eb1fe319aadcf9795dfb049fbe9fb58b5/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f313132313439392f3939353739302f30356139643038322d303963342d313165332d383535302d3537366539373634373639362e706e67) from #1987

springmeyer:
noting this is not fixed by #2671 since it is a slightly different problem as @artemp mentions (https://github.com/mapnik/mapnik/issues/2292#issuecomment-47950465)

springmeyer:
@rcoup - pinging you here as I'm curious whether you are also aiming to fix this or not? I'm recalling now that we held off on this change specifically due to filesize bloat so I'm not sure if it is viable. But I'd welcome attention if you think it is viable.

rcoup:
I'd like to fix it - issues like this tend to trip up our safety asserts like "you shouldn't have rendered a blank image here but you somehow did".

Feels like if you have fewer unique colours *before* hextree-encoding maybe we should throttle back the quantizing aggressiveness? Since PNG compression should work well if there's mostly solid colours in a tile.

Though from a quick read of the code I'm not 100% on why my test image from #2599 (with 2 colours and 1px per colour) doesn't disappear yet the test image from here does - implies maybe the tree creation or cost reduction functions have some edge cases?

/cc @sarametz

springmeyer:
[...]

Issue #1987, opened 2013-08-20
Some images broken with PNG hextree


Andrey-VI:
Found that Mapnik produces broken tiles with `png8:m=h` option enabled in certain cases. Illustrations are better than words.
1 octree:
![_3o](https://f.cloud.github.com/assets/1121499/995085/9ba5805c-09b7-11e3-8bd4-9278b7fc47a8.png)
1 hextree:
![_3h](https://f.cloud.github.com/assets/1121499/995087/a6ae6f2c-09b7-11e3-9cf9-4e32dee0a9aa.png)

2 octree:
![_2o](https://f.cloud.github.com/assets/1121499/995099/c9b2ec32-09b7-11e3-918c-e71037b355bc.png)
2 hextree:
![_2h](https://f.cloud.github.com/assets/1121499/995104/d0c95204-09b7-11e3-915f-63fd062d27bd.png)

3 octree:
![_1o](https://f.cloud.github.com/assets/1121499/995105/d68e3b46-09b7-11e3-84e0-e7c04aeef771.png)
3 hextree:
![_1h](https://f.cloud.github.com/assets/1121499/995107/dc3120c2-09b7-11e3-9505-7e51c736d4f5.png)

I'm aware of https://github.com/mapnik/mapnik/wiki/OutputFormats#hextree-details, but the results are worse that I expected. Actually, I want to use one option (`png8:m=h`) for all our maps.

springmeyer:
confirmed this is still present and is a dupe of #2292. Let's track at #2292.

Issue #1097, opened 2012-02-22
test potential hextree edge conditions


springmeyer:
Post #1087, we should develop better test cases for possible hextree and other image encoding issues like:

 * 1x1 image that results in an empty palette
 * 16x16 image with every pixel a different color
 * default to png32 if palette is empty?
 * consider removing delete/new at https://github.com/mapnik/mapnik/blob/master/include/mapnik/hextree.hpp#L330-331

springmeyer:
test script above no longer creates blank images after #2671, so closing this issue. #2292 stays open as a slightly different issue.

Issue #2068, opened 2013-11-07
hextree::quantize crashes on 2x2 image


springmeyer:
If you try rendering a 2x2 image using the testcase from #2067 then hextree crashes:

```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libmapnik.dylib                 0x00000001054f5d48 mapnik::hextree<mapnik::rgba, mapnik::RGBAPolicy>::quantize(unsigned int) const + 472
1   libmapnik.dylib                 0x00000001054fa6c0 void mapnik::save_as_png8<std::ostream, mapnik::ImageData<unsigned int>, mapnik::hextree<mapnik::rgba, mapnik::RGBAPolicy> >(std::ostream&, mapnik::ImageData<unsigned int> const&, mapnik::hextree<mapnik::rgba, mapnik::RGBAPolicy> const&, std::vector<mapnik::rgb, std::allocator<mapnik::rgb> > const&, std::vector<unsigned int, std::allocator<unsigned int> > const&, mapnik::png_options const&) + 624
2   libmapnik.dylib                 0x00000001054f8669 void mapnik::save_as_png8_hex<std::ostream, mapnik::ImageData<unsigned int> >(std::ostream&, mapnik::ImageData<unsigned int> const&, mapnik::png_options const&) + 793
3   libmapnik.dylib                 0x00000001054ebf39 void mapnik::save_to_stream<mapnik::ImageData<unsigned int> >(mapnik::ImageData<unsigned int> const&, std::ostream&, std::string const&) + 857
4   libmapnik.dylib                 0x00000001054eb939 void mapnik::save_to_file<mapnik::ImageData<unsigned int> >(mapnik::ImageData<unsigned int> const&, std::string const&, std::string const&) + 105
5   _mapnik.so                      0x0000000104e99298 render_to_file1(mapnik::Map const&, std::string const&, std::string const&) + 296
```

springmeyer:
noting: this never landed in master either. But I'm not seeing the crash anymore at e04b0617e795a6f5c988bc5c7c27cf9439e149da

```
python -c "import mapnik;im=mapnik.Image(1,1);im.background=mapnik.Color('green');im.save('1x1.png')"
```

PR #2599, opened 2014-12-07
png8 hextree/octree encoding produces blank output

rcoup:
Test case:
* png32 image that has 2x non-transparent pixels
* encoding as png8 with hextree results in a fully transparent image
* encoding as png8 with octree results in an image with 1x coloured pixel, 1x black pixel
* expected: output images resemble the input image

May be related to #1987, #1097, #2292


craigds:
This line appears to be the cause of the hextree issues at least:

https://github.com/mapnik/mapnik/blob/2.3.x/include/mapnik/hextree.hpp#L411

Replacing that with:

    if (itr->count != 0)

... I see the correct output.


test script:

    import mapnik

    img = mapnik.Image(256, 256)
    img.background = mapnik.Color("transparent")

    img.set_pixel(250, 250, mapnik.Color(255, 1, 1, 255))
    img.set_pixel(251, 250, mapnik.Color(1, 1, 255, 255))

    img.save("test-pr2599-1.png","png")
    img.save("test-pr2599-2.png","png8:m=h")
    img.save("test-pr2599-3.png","png8:m=h:g=1.0")
    img.save("test-pr2599-4.png","png8:m=o")

Looks like this was previously attempted in https://github.com/mapnik/mapnik/commit/03250869b642cf9de71d02e72abf0048fe7c8855 but was rejected for some reason?

springmeyer:
@craigds the reason it never landed was 1) uncertainty it was needed (because the crash we were trying to avoid was no longer replicable https://github.com/mapnik/mapnik/issues/2068), 2) concern (untested) that it would slow down encoding (needs testing), and 3) unclear it was worth it in that we'd need to update a lot of visual tests for no huge benefit.

So, if you can demonstrate no perf hit, you think its a good change, and review/update the visual tests then I think it would make sense to get into master.

craigds:
I can't demonstrate no perf hit. On the contrary, I can demonstrate that some images are much bigger and also take longer to write.

[This script](https://gist.github.com/craigds/d0c7ba50426ee159322a) shows the worst case (all pixels are random, so the palette is filled up).

with current trunk code:

> 7.714s for 20000 trials
> 97 bytes

with `if (itr->count != 0)`:

> 12.408s for 20000 trials
> 1345 bytes

Then feeling dejected at the bad performance, I noticed that the output image is actually completely empty for the current trunk, since *all* the palette entries are stripped. So .. for any images that this makes bigger/slower, it also fixes problems in.

rcoup:
@springmeyer any further thoughts here? If you're ok with changing it we can trawl through the visual tests and update them. What's the best approach for testing performance in this area?

springmeyer:
> So it seems likely that for any images that this makes bigger/slower, it also fixes problems in.

Interesting.

> If you think it'd be helpful, I suppose I could look into making this a configurable parameter, i.e. "png8:m=h,n=3".

If you think the only perf hit is with images it "fixes" then ideally we don't expose yet another parameter.

> What's the best approach for testing performance in this area?

The visual tests all use png8 - what about using them as a real-world case? Do they run measurably slower with the change? Also would be good to take a look at the timings in https://github.com/mapnik/mapnik/blob/master/tests/python_tests/png_encoding_test.py and make sure they don't slow down.

springmeyer:
closed as #2671 landed.

Issue #2405, opened 2014-09-03
Use unbuffered extents in raster datasource plugins


springmeyer:
We should consider doing this. Testing this patch:

```
diff --git a/plugins/input/gdal/gdal_featureset.cpp b/plugins/input/gdal/gdal_featureset.cpp
index 8252ac5..d33724b 100644
--- a/plugins/input/gdal/gdal_featureset.cpp
+++ b/plugins/input/gdal/gdal_featureset.cpp
@@ -117,7 +117,7 @@ feature_ptr gdal_featureset::get_feature(mapnik::query const& q)
     */

     view_transform t(raster_width_, raster_height_, raster_extent_, 0, 0);
-    box2d<double> intersect = raster_extent_.intersect(q.get_bbox());
+    box2d<double> intersect = raster_extent_.intersect(q.get_unbuffered_bbox());
     box2d<double> box = t.forward(intersect);

     //size of resized output pixel in source image domain
diff --git a/plugins/input/pgraster/pgraster_datasource.cpp b/plugins/input/pgraster/pgraster_datasource.cpp
index 4983dfe..3a8e92b 100644
--- a/plugins/input/pgraster/pgraster_datasource.cpp
+++ b/plugins/input/pgraster/pgraster_datasource.cpp
@@ -806,7 +806,7 @@ featureset_ptr pgraster_datasource::features_with_context(query const& q,process
 #endif


-    box2d<double> const& box = q.get_bbox();
+    box2d<double> const& box = q.get_unbuffered_bbox();
     double scale_denom = q.scale_denominator();

     CnxPool_ptr pool = ConnectionManager::instance().getPool(creator_.id());
diff --git a/plugins/input/raster/raster_datasource.cpp b/plugins/input/raster/raster_datasource.cpp
index 13d996e..32ffbad 100644
--- a/plugins/input/raster/raster_datasource.cpp
+++ b/plugins/input/raster/raster_datasource.cpp
@@ -174,7 +174,7 @@ layer_descriptor raster_datasource::get_descriptor() const
 featureset_ptr raster_datasource::features(query const& q) const
 {
     mapnik::view_transform t(width_, height_, extent_, 0, 0);
-    mapnik::box2d<double> intersect = extent_.intersect(q.get_bbox());
+    mapnik::box2d<double> intersect = extent_.intersect(q.get_unbuffered_bbox());
     mapnik::box2d<double> ext = t.forward(intersect);

[...]

PR #2615, opened 2014-12-18
Limit size of raster buffer

rafatower:
Right now the pgraster_datasource can feed the raster symbolizer with
tiles that are first scaled up then cropped and that might result in the
symbolizer trying to allocate insane amounts of memory to just scale up
a few pixels.

This is also the source of some problems such as the bad_alloc when
raster_width * raster_height overflow and tries a mem alloc with a
negative value.

### DIFF
--- a/src/agg/process_raster_symbolizer.cpp
+++ b/src/agg/process_raster_symbolizer.cpp
@@ -43,6 +43,8 @@
 #include "agg_rendering_buffer.h"
 #include "agg_pixfmt_rgba.h"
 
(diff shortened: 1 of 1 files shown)


springmeyer:
Make sure you are not setting a map or layer buffer-size. For rasters it should be set to zero until we move on #2405

rafatower:
Thanks for pointing that out. I'm closing this PR for the moment and I'll take a closer look at the buffer settings.

Issue #2374, opened 2014-08-21
rotation of markers upside down, line placement


talaj:
Is there a way how to prevent rotation of markers upside down?

See "[c]" marker on this image:

![jan_git-20140821-114455-iw65mkyf](https://cloud.githubusercontent.com/assets/1950911/3995240/228290be-2925-11e4-8997-34edbaee6956.png)

I would like to keep markers rotate in path direction but prevent them rotate more than 90°. TextSymbolizer does similar thing.

This issue seems to be related: #1030

In the case this is not possible with current mapnik, how could it be implemented the best way?

talaj:
Closed by https://github.com/mapnik/mapnik/pull/2621.

Issue #1030, opened 2012-01-16
Rotation in marker symbolizer for linestrings


jacarma:
I set marker symbols for a linestring but I got some symbols upside down (see image).

Is there a way to get the markers right or to make them not to adjust rotation to the linestring ?

Thank you

<img src="https://correo.prodevelop.es/descarga/jacarma/marker_symbolizer_for_linestrings.png"></img>

talaj:
Solved by https://github.com/mapnik/mapnik/pull/2621.

PR #2621, opened 2015-01-06
marker direction

talaj:
This is attempt to bring `upright` property to markers symbolizer which should solve https://github.com/mapnik/mapnik/issues/2374 and https://github.com/mapnik/mapnik/issues/1030.

I was thinking about reusing upright property for text but there is almost empty intersection between values for text and markers. It is also useful to set default values for text and markers separately.


talaj:
Rebased. I have changed naming and added `up` and `down` values inspired by https://github.com/mapnik/mapnik/issues/2635:
markers-upright

springmeyer:
@talaj - looks fantastic and a great contribution. What do you think about

 - auto-down behavior as well (so the inverse of `auto`)?
 - I just merged group symbolizer support for markers - can you rebase again and make sure your changes play well with groups?
 - Would you be up for tackling #2635 as well so we can keep markers and text in sync as much as possible. Now that markers are supported in groups the support for placing text and markers together will be very commonly used so I think trying to keep implementations and properties close will be beneficial.

talaj:
* I have added `auto-down` value.
* It's rebased.
* Yes, I can take a look on https://github.com/mapnik/mapnik/issues/2635. If we can look at markers as a special case of text with just one character, then markers and text can have exactly the same upright properties.

talaj:
@springmeyer - Is it ok?
markers-upright

springmeyer:
@talaj renderings look fantastic. Can you explain what left/right/left_only/right_only would do? I've never fully understood those options for text.

Also, I'm starting to wonder if we should use a better name than `marker-upright`? I wonder if `marker-direction` or `marker-orientation` would be better?

If we do settled on better naming/design for markers then I think we could implement the new name for text as well and then deprecate `text-upright`.

talaj:
@springmeyer 
This is reference image for test I have created for https://github.com/mapnik/mapnik/issues/2635. It should explain how all `upright` options work. Those lines go clockwise.
text-upright-800-800-1 0-agg

If there is better name for this functionality then I will use it.

talaj:
So I introduced `orientation` property instead of `upright` for markers since the functionality is not only about setting markers upright. It behaves almost exactly the same as `upright` for text, except of:
* Default value is `right` instead of `auto`.
* There are two more values `up` and `down`.
marker-orientation-1152-384-1 0-agg-reference

springmeyer:
I propose `marker-direction` because `text-orientation` property also/already exists and it accepts degrees rather than being an enum. What do you think?

Over this is looking great - glad to see travis is green - thanks for the hard work.

talaj:
I agree with `marker-direction`.

talaj:
Rebased with name `marker-direction`. Created pull request in mapnik-reference: https://github.com/mapnik/mapnik-reference/pull/96.

talaj:
Any chance to merge this?

artemp:
@talaj - thanks!

Issue #2633, opened 2015-01-12
Removing Image_32 from scope


flippmoke:
I am working on moving all the image_32 functions into general methods here is the short list of what I am targeting. I will make a feature/ branch for each change and will do a pull request into the release/image_data_any once it is ready.

If I put a ? at the end, I am not quite sure so would like to spark some discussion with @springmeyer and @artemp 

- [x] is_solid in [python_image](https://github.com/mapnik/mapnik/blob/release/image_data_any/bindings/python/mapnik_image.cpp#L122) and [python_view](https://github.com/mapnik/mapnik/blob/release/image_data_any/bindings/python/mapnik_image.cpp#L122) - Create new method in image_util.hpp
- [x] [painted](https://github.com/mapnik/mapnik/blob/release/image_data_any/include/mapnik/graphics.hpp#L58) - move to new "image"?
- [x] [set_background](https://github.com/mapnik/mapnik/blob/release/image_data_any/src/graphics.cpp#L115) - Not sure yet
- [x] [premultiply](https://github.com/mapnik/mapnik/blob/release/image_data_any/src/graphics.cpp#L126) - Move out to image_util, can be then used in mapnik-vector-tile
- [x] [demultiply](https://github.com/mapnik/mapnik/blob/release/image_data_any/src/graphics.cpp#L134) - Move out to image_util, can be then used in mapnik-vector-tile
- [x] [set_rectangle_alpha2](https://github.com/mapnik/mapnik/blob/release/image_data_any/include/mapnik/graphics.hpp#L223) - Remove and replace with composite src_over
- [x] [set_rectangle_alpha](https://github.com/mapnik/mapnik/blob/release/image_data_any/include/mapnik/graphics.hpp#L179)  - delete?
- [x] [merge_rectangle](https://github.com/mapnik/mapnik/blob/release/image_data_any/include/mapnik/graphics.hpp#L271) - ?
- [x] [composite_pixel](https://github.com/mapnik/mapnik/blob/release/image_data_any/src/graphics.cpp#L142) - ?
- [x] [set_rectangle](https://github.com/mapnik/mapnik/blob/release/image_data_any/src/graphics.cpp#L142) - ? 
- [x] [set_alpha](https://github.com/mapnik/mapnik/blob/release/image_data_any/src/graphics.cpp#L93) - ?
- [x] [set_color_to_alpha](https://github.com/mapnik/mapnik/blob/release/image_data_any/src/graphics.cpp#L74) - ?
[...]

PR #2640, opened 2015-01-14
Feature/premultiply demultiply move

flippmoke:
Some tests are failing currently for this, but wanted to go ahead and create the pull request so that @artemp and @springmeyer could start to review as there are a lot of changes.

Ref #2633


flippmoke:
I pushed a few more updates to this, however, still have some failing tests. I have left in some debug code in image_compositing.cpp that throws exceptions if the src or dst variables are not premultiplied, but as per @springmeyer 's request, I will have them removed prior to finishing this pull request.

springmeyer:
This is excellent work. Nice choice on throwing in the composite function for now. I actually even put in an abort to make it easy to get a backtrace of where non-premultiplied pixels were comping from. Will push a few fixes in a moment. All tests are passing for me now after 364fb49.

PR #2643, opened 2015-01-15
Removed the set_rectangle_alpha code from image_32. Added new blend usin...

flippmoke:
...g composite to python bindings to replace it by doing src_over with composite.

Ref #2633


bindings/python/mapnik_image.cpp
```diff
@@ -189,9 +189,11 @@ std::shared_ptr<image_32> frombuffer(PyObject * obj)
 }
 
 
-void blend (image_32 & im, unsigned x, unsigned y, image_32 const& im2, float opacity)
+void blend (image_32 & im, unsigned x, unsigned y, image_32 & im2, float opacity)
 {
-    im.set_rectangle_alpha2(im2.data(),x,y,opacity);
+    mapnik::premultiply_alpha(im.data());
+    mapnik::premultiply_alpha(im2.data());
```
springmeyer: is there test coverage for `mapnik.Image.blend` in tests/python_tests ? If not let's remove the function because I think there is test coverage for `mapnik.Image.composite`. If you remove add a note to https://github.com/mapnik/mapnik/wiki/Api-changes-between-v2.3-and-v3.0

PR #2645, opened 2015-01-16
Moved set_alpha to src/image_util.cpp and out of image_32. Added a boole...

flippmoke:
...an response to premultiply_alpha and demultiply_alpha so that if it is changed in set_alpha and other locations like the python bindings for composite that the image can be restored to its original state. Removed blend from python bindings.

Ref #2633

### DIFF
--- a/bindings/python/mapnik_image.cpp
+++ b/bindings/python/mapnik_image.cpp
@@ -36,6 +36,7 @@
 
 // mapnik
 #include <mapnik/graphics.hpp>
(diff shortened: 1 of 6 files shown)


flippmoke:
Not quite yet ready to merge as I still have to do set_color_to_alpha and set_grayscale_to_alpha, but went ahead and pushed branch and created pull request.

flippmoke:
Finished the changes, going to go ahead and merge.

Issue #2648, opened 2015-01-20
Bad reprojection of antimeridian query bbox


craigds:
Maps are hard in New Zealand.

We've found a bug rendering an image of a dataset covering both mainland NZ (west of the antimeridian) and the Chatham Islands (east of the antimeridian.)

Some data near the antimeridian is missing from the rendered image, due to a badly calculated query bbox.

The bug happens when all of these are true:
 * the map is in [NZTM](http://spatialreference.org/ref/epsg/nzgd2000-new-zealand-transverse-mercator-2000/).
 * the layer is in WGS84
 * the map's `maximum-extent` in NZTM covers the antimeridian
 * there is data quite close to the antmeridian (within a few degrees of +/- 180)

It seems the logic for calculating the query bbox in this instance is:

 * take render extent in NZTM (covers a big part of the world, since this a low zoom level)
 * crop it to the map's `maximum-extent`, and potentially buffers it a little (0px in this case)
 * interpolates the extent in NZTM into a 20-point box.

so far so good - see screenshot: 
![parcel-rendering-before-reprojection](https://cloud.githubusercontent.com/assets/32112/5825721/732c8a88-a152-11e4-881b-8ee59b515783.png)

Then:

 * reprojects each of the 20 points into WGS84
 * takes the bounding-box of the 20 reprojected points.
 * Because of the antimeridian split, some of the points are on the left of the world and some are on the right. However it happens that there's no points exactly at 180, so the query bounds only goes to -177 and +178ish, thus missing some data on the east cape of NZ.

![parcel-rendering-after-projection](https://cloud.githubusercontent.com/assets/32112/5825762/b2c39632-a152-11e4-853a-ddab2efe9174.png)

Here's a [fairly minimal testcase](https://drive.google.com/file/d/0B39m5rsOl0o_WDY5VUU2UDA0Szg/view?usp=sharing) (a few points) which demonstrates the render gap across the antimeridian.

I suspect a fix might involve calculating cross-products to determine winding order of the 20-point query box. If the winding order changes when reprojected, the box has hit the antimeridian. Then you could snap the X values outward to +/- 180.

A related performance issue: in this case the query box covers the whole width of the world, so potentially is going to be quite slow. Snapping the X values outward to fix the rendering gap won't fix that performance issue. However an alternative would involve splitting the box into multiple boxes, and that'd probably be much harder.

PR #2657, opened 2015-01-27
Handle bounds reprojections which cross the anti-meridian

rcoup:
Fixes the bug identified in #2648 with the simplest solution.

When doing an envelope-points reprojection *to* a geographic coordinate system, generate the points in clockwise order, and after projection check the points are still in clockwise order. If not, expand the resulting bounds to include the world width (±180°). This won't perform any worse for most cases (and we avoid even doing the is-clockwise calc for non-geographic CS), but will be correct.

When mapnik supports multiple boxes for querying, or a concept of geographic Box2d, we can revisit and easily make it better.


springmeyer:
Want to add some visual tests with small but representative sample data? This would be a nice redundancy and help give a visual sense of problems if this were to ever regress.

rcoup:
@springmeyer I added a visual test based on @craigds' data. Any idea why cairo would still show the error after the fix - does it do additional/different bounds handling somewhere? agg/grid are fine.

If you're free to jump on IRC for a couple of minutes it'd be handy :)

rcoup:
Outcome of test weirdness: visual test comparator is not finding any differences between (clearly different) broken & reference images for cairo, but the renderer is generating the right output.

springmeyer:
@rcoup replicated and fixed the cairo visual test oddity (should now fail when it should): https://github.com/mapnik/mapnik/commit/d38b3bb2cf7f84e871bed897d43d0142eaff7d5c

Issue #2662, opened 2015-01-27
Master + Travis == Green


springmeyer:
This is an issue to:

 - Announce that we're finally green again on Travis :tada: (https://travis-ci.org/mapnik/mapnik/builds/48451510)
 - Document how to use the `bootstrap.sh` as an easy way to ensure visual tests pass locally for you
 - Give some background about why we've been seeing red for a while.

![screen shot 2015-01-27 at 10 24 48 am](https://cloud.githubusercontent.com/assets/20300/5924216/c7207c88-a60e-11e4-82c0-421483286d5d.png)

## tl;dr Green is good

You can start using the new `bootstrap.sh` script to match the dependencies you use to the same dependencies other devs use and travis builds with. So the "new" way to set up Mapnik is:

```sh
cp config.py config.py.backup
source bootstrap.sh # downloads deps into ./mason_packages and generates new config.py
./configure && make
make test
```

This will download binaries of all the most recent releases of Mapnik dependencies using https://github.com/mapbox/mason. I'll be writing more docs on how Mason and `bootstrap.sh` works. But for now if anything goes wrong and you need to fully rebuild do:

```sh
rm -rf mason_packages
rm -rf .mason
cp config.py config.py.backup
source bootstrap.sh
./configure && make
make test
```

Now, if you prefer to manage your own installs of dependencies and not use `bootstrap.sh`: I completely understand. In that case, if you don't want to use the bootstrap.sh for your main development, at least it now offers you [the ability to quickly see what version of deps @springmeyer has used to generate the expected visual tests](https://github.com/mapnik/mapnik/blob/eb598f443582f05346704abefa09769933284654/bootstrap.sh#L74-L93).

## Background: Why tests have been failing

Overall:

 - We've been building out the visual test suite more
 - We want to pay attention and care about sometimes very slight visual differences
 - And yet we want our tests to fail only if they should :)

Tough goals to balance because differing dependency versions can cause minor (or rarely major) visual differences.  The main issues causing slight differences (visual or even not visual) have been freetype, gdal, and cairo but also image libraries (jpeg-turbo vs libjpeg, etc) and/or a combination of those dependencies being different on a given machine than the machine the visual tests were generated on (likely @springmeyer's machine). For a long time we've coped with this issue by:

  - @springmeyer generates the expected test images using the latest releases of all deps.
[...]

Issue #2029, opened 2013-10-04
Optimize PNG encoding when no transparency is detected in final image


springmeyer:
As a followup to #2028, #1629, and #875, we should take a look at whether automatically detecting images with 100% 256 for the alpha channel could be encoded faster. Given `png8:m=h` is now default (in Mapnik >= 2.3.x) this takes on more meaning because we may be able to save some CPU time by treating fully opaque images differently in the png8/hextree code path

PR #2671, opened 2015-01-29
Png8 encoding bug

sarametz:
Builds on PR #2599 
##### Hextree
Fixes bug where encoding a png32 with 2 non transparent pixels as png8 with hextree results in a fully transparent image
This caused a number of small changes to the visual test images
##### Octree
Fixes bug where encoding a png32 with 2 non transparent pixels as png8 with octree results in an image with 1x coloured pixel, 1x black pixel 
Images with at least one pixel with alpha less than 1 have their colours jumbled, as black was last in the palette rather than first. Two new tests have been added to show this.   
May be related to #1987, #1097, #2292


rcoup:
@springmeyer @sarametz manually checked all the images here, but Travis is obviously running something slightly different. She's going to use #2662 to see if she can get a matching set of test failures/updates. And then we need to do the performance analysis.

springmeyer:
How about you leave the image updates to me (so don't commit those changes for now). Then I'll confirm they pass and commit after merging.

rcoup:
ok, so just drop the last commit from the PR and go back to 	f38d2ee? Easy As.

For benchmarks, how about "fastest time of 10x visual-test runs" for old vs new? And a file-size check via something like:
```bash
find /tmp/mapnik-visual-images -name "*.png" -printf '%s\n' | awk '{sum += $1} END{print sum}'
```

springmeyer:
yes and yes

sarametz:
| Without Changes | With Changes
---------- | ------------ | -------------
**Fastest Time** | 100.151s | 101.956s
**Average Time** | 101.748s | 103.668s
**Slowest Time** | 104.071s | 108.847s
**File Size Check** | 41190970 | 41380993   

Time: (min | mean | max) of all 1626 visual tests run 10x
Size check command: `find /tmp/mapnik-visual-images/ -name "*.png" -type f -exec ls -l {} \; | awk '{sum += $5} END {print sum}'`

springmeyer:
Thanks for this detailed summary. I think ~1% is an acceptable increase given this fixes bugs. And hopefully we can make up the time other places like #2029. So I think its ready to merge and I will merge it once I have a chance to review and update the test images.

springmeyer:
reminder to myself: also should try running `tests/python_tests/image_encoding_speed_test.py` to see if this detects any speed diffs

springmeyer:
I applied the patch and made this change:

```diff
diff --git a/tests/python_tests/image_encoding_speed_test.py b/tests/python_tests/image_encoding_speed_test.py
index 9ba1748..999236d 100644
--- a/tests/python_tests/image_encoding_speed_test.py
+++ b/tests/python_tests/image_encoding_speed_test.py
@@ -72,7 +72,7 @@ def do_encoding():
         avg = (sum(set)/len(set))*1000
         name = func.__name__ + ' ' + format
         results[name] = [min_,avg,elapsed*1000,name,len(func())]
-        sortable[name] = [min_]
+        sortable[name] = [name]
 
     if 'blank' in tiles:
         def blank():
```

So that I could easily and quickly compare the output of that test before/after and could not see a significant change in perf (up to 3-4 ms change is just noise), so merging this now. Thanks!

PR #2682, opened 2015-02-04
fix enum word separator

talaj:
I have produced a bit of confusion by mixing up underscore and dash word separators in enum options. The `vertex_first` and `vertex_last` are bearable, but I mixed up both separators in [single enum](https://github.com/mapnik/mapnik/blob/1fcbdc3e98de893ce0b603865f57f261437fb377/src/symbolizer_enumerations.cpp#L169-L175) for `text-upright`.

I think it should be all unified to underscore. What do you think? I would also provide a fix to https://github.com/mapnik/mapnik-reference and wiki.

I'm sorry for that.


springmeyer:
I'm torn on this: I prefer hyphen (`-`) in all cases because that is was we consistently use for property names in XML. We only use underscores for programmatic API's like python where you cannot do `sym.property-name = "foo"`. So I'm inclined to fix `vertex_last` and `vertex_first` to use dashes. I also see `miter_revert` is the other oddball. I'm inclined to move to `-` and try to add back compatibility in the XML loading to avoid breaking maps already using the few properties with `_`. What do you think?

talaj:
OK, I also prefer hyphens. They are now used for `vertex-first`, `vertex-last` and `auto-down`. The reason why I decided to change them to underscore is to retain backward compatibility without introducing more complexity to the code.

I have remade this pull request, take a look please.

talaj:
[This](https://github.com/mapnik/mapnik/blob/5361d21beb1c8ad6c6e1816993fc853754fb7653/src/load_map.cpp#L1274-L1278) might be related.

include/mapnik/enumeration.hpp
```diff
@@ -176,9 +177,11 @@ class MAPNIK_DECL enumeration {
      * */
     void from_string(std::string const& str)
     {
+        std::string str_copy(str);
+        std::replace(str_copy.begin(), str_copy.end(), '_', '-');
         for (unsigned i = 0; i < THE_MAX; ++i)
```
springmeyer: ^^ this is for back compatibility? If so then could you add a note about removal at Mapnik 4.x?

talaj: Sure, added in 5c65dadd72cb6a59deec984e0babd94fc58ccf10.

PR #2686, opened 2015-02-05
Gdal.input dataset rasterio improvements

rouault:
2 commits on top of the gdal-keep-dataset-open branch to use GDALDataset::RasterIO()

### DIFF
--- a/plugins/input/gdal/gdal_datasource.cpp
+++ b/plugins/input/gdal/gdal_datasource.cpp
@@ -45,37 +45,9 @@ using mapnik::layer_descriptor;
 using mapnik::datasource_exception;
 
 
-/*
- * Opens a GDALDataset and returns a pointer to it.
- * Caller is responsible for calling GDALClose on it
- */
-inline GDALDataset* gdal_datasource::open_dataset() const
-{
-    MAPNIK_LOG_DEBUG(gdal) << "gdal_datasource: Opening " << dataset_name_;
-
-    GDALDataset *dataset;
-#if GDAL_VERSION_NUM >= 1600
-    if (shared_dataset_)
-    {
-        dataset = reinterpret_cast<GDALDataset*>(GDALOpenShared((dataset_name_).c_str(), GA_ReadOnly));
-    }
-    else
-#endif
-    {
-        dataset = reinterpret_cast<GDALDataset*>(GDALOpen((dataset_name_).c_str(), GA_ReadOnly));
-    }
-
-    if (! dataset)
-    {
-        throw datasource_exception(CPLGetLastErrorMsg());
-    }
-
-    return dataset;
-}
-
-
 gdal_datasource::gdal_datasource(parameters const& params)
     : datasource(params),
+      dataset_(nullptr),
       desc_(gdal_datasource::name(), "utf-8"),
       nodata_value_(params.get<double>("nodata")),
       nodata_tolerance_(*params.get<double>("nodata_tolerance",1e-12))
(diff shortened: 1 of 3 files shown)


springmeyer:
Excellent, thanks! Looks like travis builds are failing only because this leads to a few visual test differences. I assume the differences are improvements but I'll check them out locally when I have a moment tomorrow.

rouault:
I wouldn't expect it to cause differences in output, at least with TIFF image. For things like ECW or JPEG2000 it might be possible. The other possibility is that there's just a bug...
Unfortunately the mapnik tests don't pass, even without my changes. With or without them I get the following errors : https://gist.github.com/rouault/0cc6d983cf7e7583d653

rouault:
looking at https://travis-ci.org/mapnik/mapnik/jobs/49647112 I guess it might be due to an issue (or change of behaviour) due to https://github.com/rouault/mapnik/commit/efdca26f2d058ee362cfab24789d1ddddac935a7. I see in .travis.yml there's a "git clone https://github.com/mapbox/mapnik-test-data tests/data/mapnik-test-data". Doing that right now to be able to reproduce locally (hoping that those tests pass locally without my changes)

springmeyer:
Thanks for sharing your test output - those all look expected in that they are of two kind 1) we have a visual test that depends on very recent harfbuzz build with font-feature-settings support and 2) a lot of minor visual differences because I recently reduced the tolerances in pixels to 0 so any difference in libpng or libjpeg (for example) will trigger test failures. My plan is to try to alleviate this by detecting know versions of libpng, libjpeg, libfreetype, libcairo, and libgdal in order to disable specific tests and keep the thresholds low. But this may not be the best strategy. How do you handle these kind of issues in gdal autotests? More details at https://github.com/mapnik/mapnik/issues/2662.

rouault:
I'll try upgrading a few mapnik dependencies with bootstrap.sh.
Regarding GDAL, we are probably less prone to issues with external dependencies (and there are less contributors, and fewer ones actually running tests...). For JPEG, we have versions of checksums on datasets for libjpeg6 and libjpeg8. But occasionnaly when people use another version of a dependency than the one usually used, they report failures. No magic solution. And otherwise what make tests pass on Travis is a recommended dependency...

springmeyer:
Built with your latest changes and see the same failures as on travis. Looks like something to do with alpha? and only impact a few of many tests: https://springmeyer.s3.amazonaws.com/visual-test-results/comparison.html

springmeyer:
> I'll try upgrading a few mapnik dependencies with bootstrap.sh.

It might be easier just to note the [versioned used](https://github.com/mapnik/mapnik/blob/master/bootstrap.sh#L38-L59) and build them yourself. That is what I would recommend rather than using Mason. Because Mason only currently supports static libraries this can make it painful for configuring GDAL predictably against mason-build libpng/libjpeg/etc. Notably the libtool .la files store the wrong paths (of the remote travis build machines) plus GDAL's configure checks, as you know, largely assume shared libraries rather than static. That said, if you are already down the path and not worried about fighting a bit with static libraries, here are all the tricks I use to get GDAL to build and link against mason libs. Its sort of crazy, but I am crazy :) https://github.com/mapbox/mason/blob/gdal-1.11.1-big-pants/script.sh#L20-L89

rouault:
OK, I finally managed to get an almost green 'make test' locally. After having applied the fix  https://github.com/rouault/mapnik/commit/7c44b2412e29111c827a3a88233e9ce9afd7feae for the failures I introduced previously , only 4 failures left : https://gist.github.com/rouault/39a84f0fc4c4c0db5799 . From a quick look, they seem related to the use of GDAL 2.0dev. I haven't dug further on that.

For reference, I ended up upgrading to (not sure to have captured all dependencies and built everything in the right order...) :
- boost 1.57 (depends on icu)
- cairo 1.12.18 (depends on freetype)
- freetype 2.5.4 (depends on harfbuzz)
- icu 54.1
- harfbuzz 2cd5323
- libjpeg-turbo-1.4.0
- libpng-1.6.16
- libwebp-0.4.2
- pixman 0.32.6
- cairo 1.12.18

springmeyer:
Great. Yes, those dep versions look good. I build freetype `--with-harfbuzz=no` to avoid the circular dep but that should not matter in terms of visual tests. I think your hunch is correct that the remaining ones are due to minor (maybe unimportant) differences with gdal 2.0. I would be happy to move the bootstrap to GDAL 2.0 if that would make things easier.

springmeyer:
Merging now. Thanks again for these improvements. Will be testing the code more this week.

Issue #2729, opened 2015-03-13
Shields disappear on avoid-edges


charterchap:
I have a <GroupSymbolizer> at the top level for my shields. I'm having a problem with clipping on tile boundaries. Adding clip="false" at the GroupSymbolizer level takes care of ~70% of the cases (random guess). If I try adding some other attributes to help fix the remaining problem shields, the shields just don't render at all. For example either of avoid-edges="true" or minimum-padding="x" will cause no shields to be rendered.

charterchap:
To elaborate:
If I modify the test style group-symbolizer-line-2.xml line 78
`<GroupSymbolizer start-column="1" num-columns="1" placement="line" avoid-edges="false" repeat-distance="20" spacing="1" repeat-key="foo">`
to be
`<GroupSymbolizer start-column="1" num-columns="1" placement="line" avoid-egdes="true" repeat-distance="20" spacing="1" repeat-key="foo">`
all the shields disappear.
The same things will happen on 
`<GroupSymbolizer start-column="1" num-columns="1" placement="line" minimum-padding="15" repeat-distance="20" spacing="1" repeat-key="foo">`
Original images looks like:
![group-symbolizer-line-3-512-512-1 0-cairo-reference_avoidedgesfalse](https://cloud.githubusercontent.com/assets/10038860/6674328/7af31136-cbde-11e4-908b-ecca0fe2a18f.png)


missing shields:
![group-symbolizer-line-3-512-512-1 0-cairo-reference_avoidedgestrue](https://cloud.githubusercontent.com/assets/10038860/6674295/521922e6-cbde-11e4-8e5b-5285ed9d6841.png)

springmeyer:
closed by #2732

PR #2732, opened 2015-03-18
Fix GroupSymbolizer placement options

zerebubuth:
As originally explained in #2729, adding the options `avoid-edges` or `minimum-padding` to GroupSymbolizer cause all instances to disappear.

It turns out this is because the bounding box used for checking the edges was the extent in world-space, not pixel-space. This PR fixes that, and adds a couple of extra visual tests to ensure the correct behaviour of these options.


springmeyer:
awesome, thanks @zerebubuth - merging.

PR #2739, opened 2015-03-23
New geometry storage and API

springmeyer:
Long planned, this branch makes a faster, more flexible internal geometry API a reality.

This unlocks:

  - separate storage of polygon exterior rings and interior rings to allow for more robust clipping of parts.
  - enforces consistent winding order per OGC spec (exterior rings are CCW, interior CW)
  - reduced memory consumption for layers with many points
  - ability to adapt mapnik geometries to boost::geometry operations (in a zero-copy way)
  - ability to have i/o grammars for json/wkt work on geometries rather than paths for better efficiency and simplier code
  - much, much more...

References:

  - Closes #2612, #2203, #2078, #2151 
  - Builds upon #2688, #2149
  - Refs #1699, #2152, https://github.com/mapbox/mapnik-vector-tile/issues/59

PR #2763, opened 2015-04-17
Test polygon interior

talaj:
Results of tests 1 and 3 are suboptimal.


springmeyer:
great, merging now. These are especially valuable because I think interior placement support is commented/broken in #2739 and these tests will catch that if it is the case.

PR #2765, opened 2015-04-20
postgis: prefix table name with schema

talaj:
Schema prefix was probably forgotten in 073e46aad8f53015e201a59942bbc5c5ffbb54f3.


springmeyer:
Yes, nice catch. Do you think it would be possible to add a test for this too?

talaj:
Would be nice, but it seems postgis tests are not running in Travis anyway. Does anyone use them?

springmeyer:
The travis linux jobs should be testing postgres. This was working previously (on linux), so thanks for noticing its broken. I see the cause is something to do with the postgis programs:

```
Notice: skipping pgraster tests (raster2pgsql)
Notice: skipping postgis tests (shp2pgsql)
```

PR #2769, opened 2015-04-22
Update Boost minimum version for `is_valid`

zerebubuth:
The minimum Boost version which contains the header `boost/geometry/algorithms/is_valid.hpp`, used in `include/mapnik/geometry_is_valid.hpp` appears to be 1.56. This updates the configure (SCons) script to require that as the minimum version and the install instructions.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -58,7 +58,7 @@ Build system dependencies are:
 Mapnik Core depends on:
 
  * Boost
-    - >= 1.47 is required and >= 1.56 recommended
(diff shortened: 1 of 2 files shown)


springmeyer:
Ah, thanks for catching this. We could also just disable this functionality for older boost - I'm not aware of `is_valid` being critical for anything yet.

zerebubuth:
@springmeyer okay - I'll work up another patch to disable `is_valid` based on boost version.

PR #2770, opened 2015-04-23
Disable missing geometry functions in older Boost versions.

zerebubuth:
The functions `is_valid` and `is_simple` and their associated headers are only present in versions of Boost >= 1.56, and will cause compilation errors with previous versions of Boost. This masks them with preprocessor macros to allow the compilation to complete, at the cost of some (currently not widely used) functionality.

Refs #2769.


springmeyer:
Thanks!

PR #2783, opened 2015-04-29
fix infinite loop

talaj:
Mapnik end up in infinite loop when text `placement="line"` is used on other geometry than `line_string`.

### DIFF
--- a/src/text/symbolizer_helpers.cpp
+++ b/src/text/symbolizer_helpers.cpp
@@ -342,9 +342,10 @@ bool text_symbolizer_helper::next_line_placement() const
                 geo_itr_ = geometries_to_process_.erase(geo_itr_);
                 return true;
             }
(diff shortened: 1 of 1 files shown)


src/text/symbolizer_helpers.cpp
```diff
@@ -342,9 +342,10 @@ bool text_symbolizer_helper::next_line_placement() const
                 geo_itr_ = geometries_to_process_.erase(geo_itr_);
                 return true;
             }
-            // No placement for this geometry. Keep it in geometries_to_process_ for next try.
-            ++geo_itr_;
```
talaj: Never called when geometry != `line_string`.

artemp: @talaj - good catch, cheers!

talaj:
Shouldn't `placement="line"` also work for polygons?

artemp:
@talaj - yes, placing text along polygon boundaries

PR #2796, opened 2015-05-07
new visual test runner

talaj:
To do:
* Almost all visual tests are failing now because I'm comparing original quantized png8 reference images with in-memory `image_rgba8` actual images. My plan is not to write any files to disk unless there is diff > 0. Therefore I would like to use png32 for reference images instead of png8. What do you think? Data size shouldn't be an issue nowadays.
* Grid rendering is missing.
* Reporting is not that fancy, no colors.
* No tolerance settings. Threshold is always 0 now. 

What is new:
* Better parametrization from shell:
```
 $ test/visual/run -h
visual test runner:
  -h [ --help ]                         produce usage message
  -v [ --verbose ]                      verbose output
  -o [ --overwrite ]                    overwrite reference image
  -j [ --jobs ] arg (=1)                number of paralel threads
  --styles-dir arg (=test/data-visual/styles)
                                        directory with styles
  --images-dir arg (=test/data-visual/images)
                                        directory with reference images
  --output-dir arg (=/tmp/mapnik-visual-images)
                                        directory for output files
  -u [ --uuid-subdir ]                  write output files to subdirectory with
                                        unique name
  --styles arg                          selected styles to test
  --fonts arg (=fonts)                  font search path
  --plugins arg (=plugins/input)        input plugins search path
```
* Parallel processing. Rendering runs on the main thread if number of jobs is one.

### DIFF
--- a/test/build.py
+++ b/test/build.py
@@ -23,6 +23,9 @@ else:
         test_env.PrependUnique(CPPPATH=test_env['CAIRO_CPPPATHS'])
         test_env.Append(CPPDEFINES = '-DHAVE_CAIRO')
     test_env.PrependUnique(CPPPATH=['./'])
+    if test_env['PLATFORM'] == 'Linux':
+        test_env['LINKFLAGS'].append('-pthread')
+    test_env.AppendUnique(LIBS='boost_program_options%s' % env['BOOST_APPEND'])
     test_env_local = test_env.Clone()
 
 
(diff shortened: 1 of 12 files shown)


springmeyer:
Yay, thank you for this huge lift!

> My plan is not to write any files to disk unless there is diff > 0. Therefore I would like to use png32 for reference images instead of png8. What do you think?

:+1: 

> Grid rendering is missing.

No problem, we can add this later.

> Reporting is not that fancy, no colors.

okay.

> No tolerance settings. Threshold is always 0 now.

Great. With mason packages we want to use 0 anyway. Can adjust/add back later if needed.

springmeyer:
Please merge whenever you are ready, and then:

 - commit changes needed to https://github.com/mapnik/test-data
 - update mapnik core's submodule to grab the latest test-data

Note: this will break the python bindings, since that test runner will need an update. But this is okay: the C++ test runner is more critical to have working right now and we can adapt the python running as needed later this week.

talaj:
@springmeyer Can you please grant me write permissions to test-data?

talaj:
I have added option `--uuid-subdir`, which will place generated data into subdirectory with unique name, e.g. `/tmp/mapnik-visual-images/74d3d470-ea0b-4bd8-8242-bc014977a7af/`.

springmeyer:
> @springmeyer Can you please grant me write permissions to test-data?

@talaj done. Sorry I missed that earlier.

flippmoke:
Instead of UUID can we use? http://www.boost.org/doc/libs/1_46_0/libs/filesystem/v3/doc/reference.html#unique_path

talaj:
@flippmoke Thanks for the info, added in https://github.com/mapnik/mapnik/pull/2805. Looks simpler than UUID.

PR #2804, opened 2015-05-12
remove useless log

talaj:


src/text/placement_finder.cpp
```diff
@@ -88,7 +88,6 @@ bool placement_finder::next_position()
         horizontal_alignment_ = layout->horizontal_alignment();
         return true;
     }
-    MAPNIK_LOG_WARN(placement_finder) << "next_position() called while last call already returned false!\n";
```
talaj: Please correct me if I'm wrong, I think this log is relict and doesn't make any sense now. It is triggered when no placement is found.

Issue #2706, opened 2015-02-15
Use libimagequant for 8-bit PNG generation


pornel:
There are several issues related to low quality of PNG8 palettes: #2292 #2497 #2110

I'd like to recommend you to have a look at [libimagequant](http://pngquant.org/lib/), which is an embeddable C library for generation of *high-quality* palettes. The API allows deep integration, so you generate palette in memory without having to create any temporary PNG files.

Here's an example: 585KB down to 264KB:

![map-png24](https://cloud.githubusercontent.com/assets/72159/6204201/2570766e-b53a-11e4-9453-dece37cbc73f.png)

![map-png8](https://cloud.githubusercontent.com/assets/72159/6204202/2571c23a-b53a-11e4-9d70-b9680d0a5e90.png)

The output PNG is less than a half size and doesn't [suffer from banding, unlike palettes generated with old algorithms](http://pngquant.org/vsphotoshop.html) that always throw away least significant bits.

springmeyer:
@pornel - thank you very much for the suggestion: `libimagequant` looks impressive. If the speed is as good as Mapnik's internal code (octree.hpp and hextree.hpp) then I think moving to `libimagequant` would be possible. The lack of banding would be a great improvement.

pornel:
On x86 with SSE the speed is comparable with other quantizers, and if it's not fast enough you can use `liq_set_speed()` to disable some expensive minor enhancements.

springmeyer:
@pornel - thanks. I'm concerned that it appears libimagequant does not have unit tests. Is this the case?

pornel:
Yes. I have a large functional test suite which tracks regressions in quality, but can't redistribute it due to images being copyrighted.

der-stefan:
Yes, I would really like to have this feature. Currently, our map server saves 8-bit tiles, since 32-bit consumes too much disk space. But we get ugly color rings. See #2497 and openstreetmap/mod_tile/issues/64

springmeyer:
Okay, next action here is for someone to create an experimental branch of mapnik that tries integrating `libimagequant` and assess performance. One easy way is to use the `image_encoding_speed_test.py` like https://github.com/mapnik/mapnik/pull/2671#issuecomment-72548546. Before for reference this lives over in the python bindings now: https://github.com/mapnik/python-mapnik/blob/master/test/python_tests/image_encoding_speed_test.py

danpat:
[...]

Issue #2806, opened 2015-05-12
MAPNIK_BIG_ENDIAN


flippmoke:
I am not sure we have proper support for this define through out the code base. Should we remove it or check that all code properly supports it?

/ cc @artemp @springmeyer

springmeyer:
@flippmoke - good catch - there was a time long ago I maintained this support as I had a big endian mac laptop. Now its untested and unsupported, so we should remove. Done in #2807

PR #2817, opened 2015-05-16
Add support for libimagequant

danpat:
This change integrates support for `libimagequant` from https://github.com/pornel/pngquant.  It's BSD licensed.  It's small and self-contained, so it's simply being compiled and linked in along with the rest of the source code tree using `scons`.  This addresses #2706.

It adds new png encoder options:

```
m=iq  - use libimagequant to reduce 32bpp images to 8bit (or lower)
iqs=[1-10] - libimagequant speed setting 10 = fast, 1 = better palette, default is 3
iqd=[0-1.0] - set Floyd-Steinberg dithering level - default depends on iqs, set to 0 to disable dithering
```

The gamma setting `g=N` is also supported for values between 0 and 1.0.

For best performance, these changes should be compiled with SSE instruction support.  A new flag `USE_SSE` has been added to `SConstruct`.  By default, it will attempt to detect SSE support on the machine doing the build.  If cross-compiling, SSE instruction support will be disabled, but this can be overridden.

Generally, without SSE support, `libimagequant` is significantly slower (10x or more) than `hextree`.  With SSE, it's in the order of 2x-5x slower, depending on the `iqs` speed setting.  In some rare circumstances it actually seems faster (blank and solid images are one example).

On big-endian architectures, some work is still needed to convert the pixel `uint32_t` byte order into the RGBA byte order that `libimagequant` needs.  However, I don't know of any big-endian architectures that also support SSE, so I don't see `m=iq` being used in many situations where endianness will be a problem.

The only other outstanding issue that needs someone more familiar with C++ than me is dealing with generics properly.  `libimagequant` only operates on 32bpp encoded images.   While it doesn't look like there's a current code path that will give it anything other than an `image_rgba8_t`, it looks like it might be possible to construct such a call.  Should the generics be removed for `save_as_png8_libimagequant`?


springmeyer:
Fantastic results with SSE :)

A couple more thoughts:

 - Mapnik does not support big-endian systems: we purged last support in #2806. So I think its okay to ignore big-endian implications
 - I wonder if SSE support could be detected via header checks rather than a scons configure check? I see some attempt at this at https://github.com/danpat/mapnik/commit/d9ba2d006e7bcb62b7005e0ce613634a30ea42d5#diff-576d6116000f556c4324cecfcb54041eR31. /cc @flippmoke who may know
 - WRT to bundling: can you trim the files to just the bare minimum of libimagequant needed for the functionality?

danpat:
A little bit of digging revealed that `__SSE__` is defined by both `clang` and `gcc` when targeting an architecture that supports the SSE instruction set.  I don't have access to MSVC to check, but the docs (https://msdn.microsoft.com/en-us/library/b0084kay.aspx) indicate that `_M_IX86_FP` should be > 1 if SSE is available.

- On OS X 10.9 with `clang`, it looks like it's on by default, and can be forced on for X86 with `-msse`
- MSVC claims to build with SSE2 support by default (`/arch:SSE2`)
- On Ubuntu 14.04 with GCC 4.9, it's necessary to pass `-msse` to `gcc` to enable, it's *off by default*.

I think it's probably reasonably safe to just blindly add `-msse` and depend on those built-in headers, it's ignored for non-x86 architectures.  That will cover `clang`, `gcc` and MSVC, do we need to try to support anything else?

springmeyer:
> I think it's probably reasonably safe to just blindly add -msse and depend on those built-in headers, it's ignored for non-x86 architectures.

Okay, this sounds good: We'll just want a way to disable with a scons option like `./configure USE_SSE=False` if someone hits a platform where it will not work.

> That will cover clang, gcc and MSVC, do we need to try to support anything else?

Exactly, Those are the only compilers we test with regularly/target.

danpat:
I've added `USE_SSE` to `SConstruct` with `yes`, `no` and `platform_default` options.  Using `platform_default` will pass no SSE-like arguments to the compiler.  Turns out `gcc` will complain if you say `-mno-sse` when targeting a non-x86 architecture, so for builds on non-x86, `./configure USE_SSE=platform_default` should work.

danpat:
I'm thinking that `mapnik/png_io.h` might not be the best spot to implement this.  It adds the requirement that `libimagequant.h` be available on the include path for anyone linking against `libmapnik.(so|dylib|dll)`.  I either need to update the build script to copy `libimagequant.h` into the headers path (like is done for `agg`) or move the `libimagequant` stuff into a `.cpp`.

Suggestions?

springmeyer:
> move the libimagequant stuff into a .cpp.

Good call. Installing like `agg` is doable but not as ideal as isolating in a cpp.

danpat:
Stale

Issue #2746, opened 2015-03-30
mapnik::geometry::is_valid_rings


springmeyer:
https://github.com/mapnik/mapnik/pull/2739 brings in new polygon storage that allows `exterior` rings to be stored separately from `interior` rings (aka holes).

The next priority is to add a validation method that works on `mapnik::geometry::multi_polygon` and `mapnik::geometry::polygon` to ensure that the rings stored in `mapnik::geometry::polygon` are truly exterior and interior.

So, to make this concrete, once implemented, we could write a unit test like:

```cpp
SECTION("polygon with incorrect storage of exterior and interior rings") {
    mapnik::geometry::linear_ring exterior;
    exterior.add_coord(0,0);
    exterior.add_coord(-10,0);
    exterior.add_coord(-10,10);
    exterior.add_coord(0,10);
    exterior.add_coord(0,0);
    mapnik::geometry::linear_ring hole;
    hole.add_coord(-7,7);
    hole.add_coord(-7,3);
    hole.add_coord(-3,3);
    hole.add_coord(-3,7);
    hole.add_coord(-7,7);

    mapnik::geometry::polygon correct;
    // add rings correctly and ensure we detect these as correct
    {
        mapnik::geometry::linear_ring exterior2(exterior);
        mapnik::geometry::linear_ring hole2(hole);
        correct.set_exterior_ring(std::move(exterior2));
        correct.add_hole(std::move(hole2));        
    }
    REQUIRE( mapnik::geometry::is_valid_rings(correct) );

    mapnik::geometry::polygon in_correct;
    // add rings incorrectly and ensure we detect these as incorrect
    {
        mapnik::geometry::linear_ring exterior2(exterior);
        mapnik::geometry::linear_ring hole2(exterior2);
        correct.set_exterior_ring(std::move(hole2));
        correct.add_hole(std::move(exterior2));        
    }
    REQUIRE( !mapnik::geometry::is_valid_rings(correct) );
}
```

@artemp - I'll leave it to you to think through how this algo will work internally. My sense is that it won't be fast (okay) and for self-intersecting polygons will likely not be robust. But the primary goal is to be 100% confident that the rings we later expose via `polygon_vertex_adapter.vertex` (for multipolygons) are consistently ordered (and oriented) such that each `move_to` unambiguously can be determined as either a new outer ring (ccw) or an interior ring (cw) truly inside the previous outer ring.

springmeyer:
[...]

PR #2818, opened 2015-05-17
Add validation for polygon rings

danpat:
This adds `mapnik::geometry::is_valid_rings()` which performs ring validation on `polygon` and `multi_polygon` objects.  This is to address #2746.  It performs the following checks:
- Exterior ring fully encloses interior rings
- Interior rings don't overlap
- Exterior ring wind CCW
- Interior rings wind CW
- No rings self-intersect (i.e. all rings are simple polygons)

Test coverage included in `test/unit/geometry/geometry_is_valid.cpp`

### DIFF
--- a/include/mapnik/geometry_is_valid.hpp
+++ b/include/mapnik/geometry_is_valid.hpp
@@ -30,9 +30,12 @@
 
 #include <mapnik/geometry.hpp>
 #include <mapnik/geometry_adapters.hpp>
-#include <boost/geometry/geometries/geometries.hpp>
+#include <boost/geometry/geometries/point_xy.hpp>
+#include <boost/geometry/geometries/segment.hpp>
 #include <boost/geometry/algorithms/is_valid.hpp>
 #include <boost/geometry/algorithms/intersection.hpp>
+#include <boost/geometry/algorithms/area.hpp>
+#include <boost/geometry/algorithms/crosses.hpp>
 
 namespace mapnik { namespace geometry {
 
@@ -108,26 +111,23 @@ inline bool is_valid(T const& geom)
     return detail::geometry_is_valid() (geom);
 }
 
-typedef boost::geometry::model::point<double, 2, boost::geometry::cs::cartesian> Point;
+typedef boost::geometry::model::d2::point_xy<double> Point;
 typedef boost::geometry::model::segment<Point> Segment;
 
 template <typename T>
 std::vector<Segment> make_segment_vector(linear_ring<T> ring)
 {
-
     std::vector<Segment> result;
 
-    for (int i = 0; i < ring.size()-1; ++i)
-    {
-        Point a( ring.at(i).x, ring.at(i).y );
-        Point b( ring.at(i+1).x, ring.at(i+1).y );
-        Segment s(a, b);
-        result.push_back(std::move(s));
+    if (ring.size() > 1) {
+        for (int i = 0; i < ring.size()-1; ++i)
+        {
+            Point a( ring.at(i).x, ring.at(i).y );
+            Point b( ring.at(i+1).x, ring.at(i+1).y );
+            Segment s(a, b);
+            result.push_back(std::move(s));
+        }
     }
-    Point a( ring.back().x, ring.back().y );
-    Point b( ring.front().x, ring.front().y );
-    Segment s(a, b);
-    result.push_back(std::move(s));
 
     return result;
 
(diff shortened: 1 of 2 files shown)


include/mapnik/geometry_is_valid.hpp
```diff
@@ -106,6 +110,140 @@ inline bool is_valid(T const& geom)
...
+typedef boost::geometry::model::d2::point_xy<double> Point;
+typedef boost::geometry::model::segment<Point> Segment;
+
+template <typename T>
+std::vector<Segment> make_segment_vector(linear_ring<T> ring)
+{
+    std::vector<Segment> result;
+
+    if (ring.size() > 1) {
+        for (int i = 0; i < ring.size()-1; ++i)
```
springmeyer: I recommend caching the value of ring.size() (just in case the compiler is not smart enough to do this) and calling `result.reserve` since the final size of the `result` is known. Should lead to small speedups.

```cpp
std::size_t num_rings = ring.size();
result.reserve(num_rings-1);
```

include/mapnik/geometry_is_valid.hpp
```diff
@@ -106,6 +110,140 @@ inline bool is_valid(T const& geom)
...
+        {
+            Point a( ring.at(i).x, ring.at(i).y );
+            Point b( ring.at(i+1).x, ring.at(i+1).y );
+            Segment s(a, b);
+            result.push_back(std::move(s));
+        }
```
springmeyer: Doing `result.emplace(a,b)` might be slightly faster.

include/mapnik/geometry_is_valid.hpp
```diff
@@ -106,6 +110,140 @@ inline bool is_valid(T const& geom)
...
+inline bool is_valid_rings(polygon<T> const& poly)
+{
+    // If there are no interior rings, then the polygon must be valid
+    if (poly.interior_rings.size() == 0) return true;
+
+    // If there is no exterior rings, and any interior rings have points
+    // then this is not a valid polygon
+    if (poly.exterior_ring.size() == 0) 
+    {
+        for (linear_ring<T> interior_ring : poly.interior_rings)
```
springmeyer: `for (linear_ring<T> interior_ring : poly.interior_rings) ` will lead to a copy of each ring. Best to ask for a `const&` to avoid unintentional copy since you don't need to modify the `interior_ring`. I recommend:

```
for (auto const& interior_ring : poly.interior_rings) 
````

File: include/mapnik/geometry_is_valid.hpp [hunk 4]
```diff
@@ -106,6 +110,140 @@ inline bool is_valid(T const& geom)
...
+
+    return true;
+}
+
+template <typename T>
+inline bool is_valid_rings(multi_polygon<T> const& multipoly) {
+    for(polygon<T> poly : multipoly) 
+    {
+        if (!is_valid_rings(poly)) return false;
+    }
```
springmeyer: I think this is missing a `return true;` after the loop over each poly.

springmeyer:
@danpat - this looks excellent.

I made some minor suggestions inline about avoiding copies. It would be great if you could ensure `const&` is used for passing arguments and within `for` loops across the board.

Before merging I'll plan run this against an OSM database I have based on osm2pgsql and used for rendering vector tiles. My plan is to slip the validation into https://github.com/mapbox/mapnik-vector-tile/blob/master/src/vector_tile_processor.ipp#L1252 to get a better sense of invalid polygons. The database is known to have self-intersecting polygons (which we don't have a plan to fix yet) but should not have any interior rings that extend outside their exteriors, so I'm assuming if that fails it is a bug in mapnik we'll want to fix.

Overall aim is to have Mapnik able to generate geometries viable for triangulation with https://github.com/mapbox/earcut. There is a balance we're still exploring: which is how much invalid geometries are fixable serverside (mapnik) and how much earcut can/should cope with invalid geometries.

include/mapnik/geometry_is_valid.hpp
```diff
@@ -106,6 +110,140 @@ inline bool is_valid(T const& geom)
...
+                }
+            }
+        }
+    }
+
+    // Then, make sure there are no line intersections between the exterior ring
+    // and any of the interior rings
+    // TODO: implement boost::geometry::crosses, or convert this to Shamos–Hoey
+
+    auto exterior_segments = make_segment_vector(poly.exterior_ring);
```
flippmoke: Rather the doing make segment and copying what could be a large amount of memory, I am curious if you could make simply use `boost::geometry::intersects` like you do with the poly rings rings below.

include/mapnik/geometry_is_valid.hpp
```diff
@@ -106,6 +110,140 @@ inline bool is_valid(T const& geom)
...
+        {
+            return false;
+        }
+    }
+
+    // Quick test to see if the interior rings have at least one point
+    // inside the exterior ring.  This test, combined with the line crossing
+    // test next will ensure that all rings are inside the exterior.
+    // Also, check that rings start/end on the same point
+    for (linear_ring<T> ring : poly.interior_rings)
```
flippmoke: Would `boost::geometry::intersects` not catch this?

include/mapnik/geometry_is_valid.hpp
```diff
@@ -106,6 +110,140 @@ inline bool is_valid(T const& geom)
...
+    {
+        if (boost::geometry::intersects(ring)) 
+        {
+            return false;
+        }
+    }
+
+    // Now, verify the winding directions of everything:  CCW for the exterior ring, and
+    // CW for any interior rings.
+    if (boost::geometry::area(poly.exterior_ring) < 0)
```
flippmoke: We should be careful with declaring CCW vs CW (in comments), because if the y axis is positive downward then this logic is then reversed.

include/mapnik/geometry_is_valid.hpp
```diff
@@ -106,6 +110,140 @@ inline bool is_valid(T const& geom)
     return detail::geometry_is_valid() (geom);
 }
 
+typedef boost::geometry::model::d2::point_xy<double> Point;
```
flippmoke: We already have defined Point in `include/mapnik/geometry.hpp` that works with boost algorithms, we should probably use that if we need Point declared.

test/unit/geometry/geometry_is_valid.cpp
```diff
@@ -26,6 +26,218 @@ SECTION("line_string") {
 
 }
 
+SECTION("polygon ring tests") {
```
flippmoke: Would suggest to add a test case for repeated points within the polygon exterior and interior rings. I would suggest that there be at least 3 points repeated in a row in this test as well. This might not be "invalid" but just a poorly constructed polygon.

include/mapnik/geometry_is_valid.hpp
danpat: `boost::geometry::intersects` when performed on polygons returns `true` if their surfaces overlap, which is always true for interior rings.  There is `boost::geometry::crosses` which does take the `linear_ring<T>` type, but it throws `not_implemented` as of Boost 1.58.

There is definitely a speedup to be had here by switching to the Shamos-Hoey algorithm (from O(n^2) -> O(nlogn)), ideally that would be in `boost::geometry::crosses`.

flippmoke: Since linear_ring can be downcast to line_string you should be able to do it and implement crosses. 
https://github.com/mapnik/mapnik/blob/master/include/mapnik/geometry.hpp#L77

danpat: Ahh, my C++-fu is weak, I think I see how that should be implemented.  I'll make that change.

danpat: I'm a doofus, `boost::geometry::within` exists and does what's required here.

include/mapnik/geometry_is_valid.hpp
danpat: The only way to do that reliably would be to know the CRS in use and use the appropriate logic.  I'm not clear on where this code is going to be called from, will it be practical to discover the CRS, or pass in a flag do indicate y-positiveness?

flippmoke: We probably should just call it positive orientation and negative orientation instead of winding order, then it is ambiguous of the axis. This code will actually be used in a situation where y positive axis is downward (mapnik-vector-tiles), so wanted to be clear on this.

test/unit/geometry/geometry_is_valid.cpp
danpat: Just to be clear, `is_valid_rings()` should fail on >2 repeated points?

danpat:
Stale

PR #2824, opened 2015-05-18
Port some python tests

zerebubuth:
Ported across a few tests from the Python bindings test suite, which has been removed from the main repository. Tests were ported pretty much as-is, and pass on my machine (Ubuntu 14.04).

I wasn't quite sure whether these belong to the "unit" or "standalone" set - I've put the "smaller" ones in with the unit tests for now, but it would be great to get some guidance as to which is more appropriate.

### DIFF
--- a/test/build.py
+++ b/test/build.py
@@ -11,9 +11,9 @@ if not env['CPP_TESTS']:
         os.unlink(cpp_test_bin)
 else:
     test_env['LIBS'] = [env['MAPNIK_NAME']]
-    test_env.AppendUnique(LIBS=copy(env['LIBMAPNIK_LIBS']))
     test_env.AppendUnique(LIBS='mapnik-wkt')
     test_env.AppendUnique(LIBS='mapnik-json')
+    test_env.AppendUnique(LIBS=copy(env['LIBMAPNIK_LIBS']))
     if env['RUNTIME_LINK'] == 'static' and env['PLATFORM'] == 'Linux':
         test_env.AppendUnique(LIBS='dl')
     test_env.AppendUnique(CXXFLAGS='-g')
--- a/test/standalone/agg_rasterizer_integer_overflow_test.cpp
+++ b/test/standalone/agg_rasterizer_integer_overflow_test.cpp
@@ -1,0 +1,70 @@
+#define CATCH_CONFIG_MAIN
+#include "catch.hpp"
+
+#include <mapnik/map.hpp>
+#include <mapnik/memory_datasource.hpp>
+#include <mapnik/json/feature_parser.hpp>
+#include <mapnik/feature_type_style.hpp>
+#include <mapnik/rule.hpp>
+#include <mapnik/symbolizer.hpp>
+#include <mapnik/layer.hpp>
+#include <mapnik/agg_renderer.hpp>
+#include <mapnik/debug.hpp>
+
+#include <iostream>
+
+// geojson box of the world
+const std::string geojson("{ \"type\": \"Feature\", \"properties\": { }, \"geometry\": { \"type\": \"Polygon\", \"coordinates\": [ [ [ -17963313.143242701888084, -6300857.11560364998877 ], [ -17963313.143242701888084, 13071343.332991421222687 ], [ 7396658.353099936619401, 13071343.332991421222687 ], [ 7396658.353099936619401, -6300857.11560364998877 ], [ -17963313.143242701888084, -6300857.11560364998877 ] ] ] } }");
+
+TEST_CASE("agg_rasterizer_integer_overflow") {
+
+SECTION("coordinates_do_not_overflow_and_polygon_is_rendered") {
+  try {
+    auto expected_color = mapnik::color("white");
+
+    mapnik::Map m(256,256);
+    m.set_background(mapnik::color("black"));
+
+    mapnik::feature_type_style s;
+    {
+      mapnik::rule r;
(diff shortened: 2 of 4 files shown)


zerebubuth:
I should note: this needs the extra CSV file added in mapnik/test-data#1 to pass.

test/standalone/agg_rasterizer_integer_overflow_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    mapnik::image_rgba8 im(256, 256);
+    {
+      mapnik::agg_renderer<mapnik::image_rgba8> ren(m, im);
+      ren.apply();
+    }
+
+    REQUIRE(im(128,128) == expected_color.rgba());
+
+  } catch (std::exception const &e) {
```
flippmoke: You don't need try catch around test sections, I believe that is all handled internally by catch.hpp

test/standalone/agg_rasterizer_integer_overflow_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+    std::clog << e.what() << std::endl;
```
flippmoke: If you ever want to print the output in the event of a test failure you can use INFO("str");

test/standalone/csv_test.cpp
```diff
@@ -0,0 +1,715 @@
...
+      std::vector<bfs::path> broken;
+      add_csv_files("test/data/csv/fails", broken);
+      add_csv_files("test/data/csv/warns", broken);
+      broken.emplace_back("test/data/csv/fails/does_not_exist.csv");
+
+      for (auto const &path : broken) {
+        bool threw = false;
+
+        try {
+          auto ds = get_csv_ds(path.native());
```
flippmoke: Wrap this in a 'REQUIRE_THROWS()'  you won't need the try catch.

test/standalone/csv_test.cpp
```diff
@@ -0,0 +1,715 @@
...
+    auto ds = get_csv_ds("test/data/csv/number_types.csv");
+    auto fields = ds->get_descriptor().get_descriptors();
+    require_field_names(fields, {"x", "y", "floats"});
+    require_field_types(fields, {mapnik::Integer, mapnik::Integer, mapnik::Double});
+
+    auto fs = all_features(ds);
+    for (double d : { .0, +.0, 1e-06, -1e-06, 0.000001, 1.234e+16, 1.234e+16 }) {
+      auto feature = fs->next();
+      REQUIRE(bool(feature));
+      CHECK(feature->get("floats") == d);
```
flippmoke: When comparing floats its typically a good idea to use 'Approx' which is also part of catch.hpp. An example of it being used is [here](https://github.com/mapnik/mapnik/blob/master/test/unit/geometry/geometry_equal.hpp#L121).

flippmoke:
I merged your test/data changes, go ahead and do a `git submodule upgrade --remote` and then add the need test/data tag as well to your merge request :+1:

zerebubuth:
@flippmoke - thanks for the comments! I've had a go at fixing them, if you'd like to take another look? :smile:

test/build.py
```diff
@@ -11,9 +11,9 @@
         os.unlink(cpp_test_bin)
 else:
     test_env['LIBS'] = [env['MAPNIK_NAME']]
-    test_env.AppendUnique(LIBS=copy(env['LIBMAPNIK_LIBS']))
     test_env.AppendUnique(LIBS='mapnik-wkt')
     test_env.AppendUnique(LIBS='mapnik-json')
+    test_env.AppendUnique(LIBS=copy(env['LIBMAPNIK_LIBS']))
```
flippmoke: Any reason why this order needed to change?

test/unit/core/box2d_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+  e1 = mapnik::box2d<double>(4772116.5490, -2744395.0631, 5765186.4203, -1609458.0673);
+  e2 = mapnik::box2d<double>(5124338.3753, -2240522.1727, 5207501.8621, -2130452.8520);
+  e1.clip(e2);
+  REQUIRE(e1 == e2);
+
+  // nz in lon/lat
+  e1 = mapnik::box2d<double>(163.8062, -47.1897, 179.3628, -33.9069);
+  e2 = mapnik::box2d<double>(173.7378, -39.6395, 174.4849, -38.9252);
+  e1.clip(e2);
+  REQUIRE(e1 == e2);
```
flippmoke: Approx here might be good

flippmoke: Nevermind, I take that back

test/build.py
zerebubuth: Yup, `libmapnik-json` refers to `icu_52::UnicodeString::~UnicodeString()`, which is defined in `libicuuc`. With the previous link order, `-licuuc` preceded `-lmapnik-json` and so I was getting:

    /usr/bin/ld: src/libmapnik.so: undefined reference to symbol '_ZN6icu_5213UnicodeStringD1Ev'
    /usr/lib/x86_64-linux-gnu/libicuuc.so: error adding symbols: DSO missing from command line
    collect2: error: ld returned 1 exit status

Swapping the link order means that the ICU symbol is available.

springmeyer: good fix @zerebubuth. Since I develop locally on OS X I always miss linux problems like this. Thanks for catching it.

PR #2827, opened 2015-05-19
visual tests: set number of parallel threads by JOBS environment variable

talaj:
Speed up visual tests.


springmeyer:
Do you think we should run tests in parallel by default?

talaj:
Why not? In most cases one wants to run them as fast as possible. It also might uncover some concurrency problems.

talaj:
Maybe use some another variable, `TEST_JOBS`...

springmeyer:
I think manually setting JOBS is fine for now.

PR #2830, opened 2015-05-19
Rename set_alpha and multiply_alpha

flippmoke:
Renamed both set_alpha and multiply_alpha to set_opacity and multiply_opacity. Enabled the possibility for multiply_alpha to be values greater then 1.0. Added a benchmark that tests numeric_cast vs clamping. 

@artemp would love if you could review


artemp:
@flippmoke - functionally looks ok but here are *important* points 

*   current set_opacity just sets opacity  ```row_to[x] = (a1 << 24)| (b << 16) |  (g << 8) | (r) ;``` This is not intuitive as it might seem to you. Operations on alpha are commonly (and for the good reason!) multiplicative e.g  new_alpha = a0 * a1.  If you feel real strong about `set_opacity` naming lets rename it to `apply_opacity`. 

* multiply_alpha - I'm not comfortable with  APIs like  `multiply_alpha(image, 10000)`.  Why just 'alpha` ?
Should we add `multiply_red` etc ?  

As I mentioned for arbitrary image transformations like `scaling` lets provide proper framework like ```transform(image, functor)``` 

/cc @springmeyer

PR #2832, opened 2015-05-20
Message about unused text nodes now also contains line numbers.

afflerbach:
Helped me a lot when I needed to find a superfluous `-->` in a large `map.xml`.


springmeyer:
Thank you, good fix.

Issue #2841, opened 2015-05-21
Release candidate tag(s) need test data packaged inside


springmeyer:
@flippmoke - realizing that to be able to run the tests we need to package the test data. I just tested building the rc2 from scratch like:

```sh
wget https://github.com/mapnik/mapnik/archive/v3.0.0-rc2.tar.gz -O mapnik-v3.0.0-rc2.tar.gz
tar xf mapnik-v3.0.0-rc2.tar.gz
cd mapnik-3.0.0-rc2/
source ./bootstrap.sh
./configure && make && make test
```

And I get:

```
===============================================================================
test cases:  1 |  0 passed | 1 failed
assertions: 41 | 40 passed | 1 failed

* Running visual tests...
libc++abi.dylib: terminating with uncaught exception of type boost::filesystem::filesystem_error: boost::filesystem::directory_iterator::construct: No such file or directory: "test/data-visual/styles"
./test/run: line 31: 91307 Abort trap: 6           ./test/visual/run -j $JOBS
make: *** [test] Error 177
```

So, this says to me we need a defined rc and release protocol that packages the test data, tars up the result, posts to s3, and then provides that link to users.

springmeyer:
> libc++abi.dylib: terminating with uncaught exception of type boost::filesystem::filesystem_error: boost::filesystem::directory_iterator::construct: No such file or directory: "test/data-visual/styles"

@flippmoke @talaj - what do you think about blocking the user from running the tests if the data is missing with a friendly message?

talaj:
> what do you think about blocking the user from running the tests if the data is missing with a friendly message?

Yes, I will take a look on that.

PR #2845, opened 2015-05-22
visual tests: skip visual tests when test data are not present

talaj:
Addresses https://github.com/mapnik/mapnik/issues/2841#issuecomment-104448028.

I was thinking about handling it somehow in the visual test runner itself, but I'm not convinced to do so as it looks better to handle it outside of the runner. On the other side I'm planning to improve visual tests reporting and I will probably handle missing styles directory more gracefully.

I hope this is sufficient for now.


springmeyer:
> I hope this is sufficient for now.

Yes, good step forward for now. Thanks!

Issue #2237, opened 2014-05-06
Validate expressions types at load_map?


springmeyer:
We have just one failing test in master (3.x) after the expr-v2 merge. The reason is that stroke-width now being an expression is not validated at load_map but rather lazily parsed at render time. So:

``` xml
        <LineSymbolizer stroke-width="foobar"/>
```

No longer throws even though eventually at render time it will not evaluate into anything meaningful. 

Going tweak this test for now to make it pass, but meditate on this more....

```
======================================================================
FAIL: python_tests.load_map_test.test_broken_files
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/python-libs/nose-1.3.0-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/dane/projects/mapnik/master/tests/python_tests/load_map_test.py", line 37, in test_broken_files
    eq_(len(failures),0,'\n'+'\n'.join(failures))
AssertionError: 
Loading broken map (../data/broken_maps/css_value_broken.xml) did not raise RuntimeError!
----------------------------------------------------------------------
```

PR #2847, opened 2015-05-23
No throw rendering

springmeyer:
There are very few to near zero situations where rendering should throw. Instead if something unexpected happens during rendering the problem item should be skipped or not rendered and rendering should continue.

One situation, that is newly a problem, is that after all properties moved to expressions they became much harder to validate at XML load (#2237).

This branch contains a reduced testcase for when an invalid color value is passed that may slip through XML loading and will throw during rendering (not ideal).


springmeyer:
This was a worthy effort but is now stale, closing.

PR #2853, opened 2015-05-23
Additional markers symbolizer shapes

afflerbach:


flippmoke:
Can you please provide some visual tests for these new markers? https://github.com/mapnik/mapnik/wiki/A-perfect-testcase

afflerbach:
Okay, will do. Shall I put them in https://github.com/mapnik/test-data-visual/blob/master/styles/marker-symbolizer-expressions-all.xml or rather separated in a new file?

afflerbach:
See [pull request in test-data-visual](https://github.com/mapnik/test-data-visual/pull/7).

springmeyer:
Thanks for the efforts here @afflerbach. However, I'm not convinced we should add any new built-in markers because:

 - its easy to provide markers via external svg
 - given the current design of the markers_cache, it takes a bit more memory for every process since these are always loaded and not lazily loaded. (not a problem with your pull, but based on current design)
 - so you might wonder why we provide the current defaults (ellipse and arrow) at all? Well, I added them to be back compatible with Mapnik 2.x. But I'm not convinced defaults are a great thing.

So, can you make a pitch for why you think defaults are useful and why these particular defaults are a positive addition?

afflerbach:
>  given the current design of the markers_cache, it takes a bit more memory for every process since these are always loaded and not lazily loaded. (not a problem with your pull, but based on current design)

Okay, agreed. Didn't think about that.

> so you might wonder why we provide the current defaults (ellipse and arrow) at all?  Well, I added them to be back compatible with Mapnik 2.x. But I'm not convinced defaults are a great thing.

Okay, I didn't know that :) I thought more symbols and especially more arrow-like shapes could be useful for  `<MarkersSymbolizer placement="line" …/>`. 

But in my opinon, if Mapnik does not offer more built-in symbols it would be a good idea to at least provide a small collection of suitable SVG marker files like arrows, asterisks, flags, etc. – no lady bugs ;). And an appropriate example in the docs (I know, difficult subject :)  

For me as a software guy and part-time map designer it was a nice little exercise in Mapnik coding and the chance to give something back to a (really cool!) open source project. 

Anyway, no problem for me if you don't accept my changes.

springmeyer:
Okay, thanks for the discussion @afflerbach - will close this and track at #3062. And thanks for your contribution - it was very well done. Let me know if there are other ways you are interested in contributing and we'll be happy to help out. Just create an issue to discuss.

PR #2857, opened 2015-05-24
Switch default compiler to cc/c++

pnorman:
On Linux this will normally be gcc/g++ and remain equivalent. On FreeBSD these point to clang and gcc is not installed by default.

It might be possible to remove the OS X special casing below, but I don't have a machine suitable for testing.


pnorman:
This would eliminate the need to special case the compiler for OS X, but then the checks for what compiler later on would need to check what compiler, not the name of the executable.

springmeyer:
The special casing for OS X should no longer be needed as g++ and gcc are symlinks to clang++ and clang on recent OS X versions. So applied the fix manually in 2513dcc.

PR #2878, opened 2015-05-29
Add Map XML I/O tests

zerebubuth:
This adds tests, ported from the original Python `load_map_test.py` and `save_map_test.py`, which test loading, saving and round-tripping the XML through Mapnik.

This also adds a wrapper around GDAL's error handling so that messages are passed through Mapnik's error system, which allows filtering or silencing errors. Useful in the tests, when we're expecting failure.

### DIFF
--- a/test/standalone/map_xml_test.cpp
+++ b/test/standalone/map_xml_test.cpp
@@ -1,0 +1,171 @@
+#define CATCH_CONFIG_MAIN
+#include "catch.hpp"
+
+#include <mapnik/map.hpp>
+#include <mapnik/load_map.hpp>
+#include <mapnik/save_map.hpp>
+#include <mapnik/datasource.hpp>
+#include <mapnik/datasource_cache.hpp>
+#include <mapnik/debug.hpp>
+
+#include <boost/filesystem.hpp>
+#include <boost/range/iterator_range_core.hpp>
+
+#include <iostream>
+
+namespace bfs = boost::filesystem;
+
+namespace {
+
+class tmp_dir {
+private:
+  bfs::path m_path;
+
+public:
+  tmp_dir()
+    : m_path(bfs::temp_directory_path() / bfs::unique_path("mapnik-test-%%%%-%%%%-%%%%-%%%%")) {
+    bfs::create_directories(m_path);
+  }
+
+  ~tmp_dir() {
+    // files might be deleted by other things while the test is
+    // running, which isn't necessarily an error as far as this
+    // code is concerned - it just wants to delete everything
+    // underneath the temporary directory.
+    boost::system::error_code err;
+
+    // catch all errors - we don't want to throw in the destructor
+    try {
+      // but loop while the path exists and the errors are
+      // ignorable.
+      while (bfs::exists(m_path)) {
+        bfs::remove_all(m_path, err);
+
+        // for any non-ignorable error, there's not much we can
+        // do from the destructor. it's in /tmp anyway, so it'll
+        // get reclaimed next boot.
+        if (err && (err != boost::system::errc::no_such_file_or_directory)) {
+          break;
+        }
+      }
+
+    } catch (const std::exception &e) {
+      std::cerr << "Exception caught while trying to remove "
+                << "temporary directory " << m_path
+                << ": " << e.what() << "\n";
+
+    } catch (...) {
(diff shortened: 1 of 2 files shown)


flippmoke:
Only comment here for @springmeyer as well, do you think that this should be a unit test or a standalone test?

test/standalone/map_xml_test.cpp
```diff
@@ -0,0 +1,206 @@
...
+    std::vector<bfs::path> broken_maps;
+    add_xml_files("test/data/broken_maps", broken_maps);
+    broken_maps.emplace_back("test/data/broken_maps/does_not_exist.xml");
+
+    for (auto const &path : broken_maps) {
+      CAPTURE(path.native());
+
+      mapnik::Map m(256, 256);
+      REQUIRE(m.register_fonts("fonts", true));
+      REQUIRE_THROWS(mapnik::load_map(m, path.native(), true));
```
talaj: I would maybe try `REQUIRE_THROWS_AS( ... , mapnik::config_error)`. Just saying...

zerebubuth: Is it intended to only throw `mapnik::config_error`? There are several places where it throws other things, for example: the exceptions defined in `xml_node.hpp` (`node_not_found`, `attribute_not_found` and `more_than_one_child`) all derive from `std::exception`. Should they derive from `config_error`, or be caught and wrapped as `config_error`s in `load_map`?

talaj: Yeah, you are correct, there are also other exception types. Never mind, that was just idea.

zerebubuth: It would be pretty easy to trap the exceptions in `load_map` - and I'm not sure whether the extra information about the type of error is useful to the caller of `load_map` beyond that it's a config error. @springmeyer, do you think it would be good to have a single, documented expected exception type thrown from `load_map`?

flippmoke: This is kind of a problem across the board, we need well document exception types for almost everything and we don't right now honestly. I don't think having just one is really something pressing, I think just knowing what they can be is important.

springmeyer: Yeah, I propose not caring about the type of exception now in this test.

springmeyer:
Looks like this is failing on travis because in test/data/good_maps/empty_parameter2.xml `<Parameter name="empty"><![CDATA[ ]]></Parameter>` is round tripping as `<Parameter name="empty" />`. I think that is fine and we can ignore the difference. That test was added to ensure that an empty parameter did not throw. Its not totally correct to strip that whitespace but does not seem like much harm either. Wonder why this was not happening with libxml2....

zerebubuth:
@springmeyer yup, the behaviour differed between the rapidxml and libxml2 implementations of the XML loader -- the libxml2 one was trimming leading and trailing whitespace for content values. That seemed fairly intentional (added in 533b95f0), so #2880 was a fix to make the behaviour the same. If you think the behaviour should be _not_ to trim whitespace, then I'm happy to submit another PR to take out the trimming for both XML loaders -- just say the word.

springmeyer:
@zerebubuth thanks for the details. I'm not totally clear on what trimming gets us. And I recall it is somewhat expensive. So, yes, I'm interested if we could consistently not trim. Want to check if there are any suboptimal test failures if 1) we remove trimming from libxml2 and 2) (of course) build to use libxml2?

zerebubuth:
Cool. I've rebased this on top of master to include #2880, so I'm hoping for a clean build this time. I'll start working on removing the trimming.

zerebubuth:
@springmeyer - I've put a PR for you to look at: #2886. It's pretty much orthogonal to this one, so I didn't think it belonged on the end of this PR.

lightmare:
Can this be closed without merging?
The main commit adding good/broken map tests was already cherry-picked into master.
The second one adding round-trip to string is not that important, I think; or actually, it might be better to *only* `save_map_to_string` and not write any files. I started playing with this and reporting only relevant portions of the dumps that differ, not the whole xml, so I can make a follow-up PR if that's desired.
The third commit suppressing GDAL error output I'm not sure about. Given that it's possible to use GDAL alongside Mapnik in an application, interfering with its global error handler doesn't seem right.

springmeyer:
> Can this be closed without merging?

👍

PR #2880, opened 2015-05-31
Fix rapidxml XML loader to trim whitespace

zerebubuth:
The libxml parser already trims text nodes, which leads to a difference in behaviour when reading map style XML. This patch normalises the behaviour between the two, and adds a test case to ensure the future behaviour stays the same.

Note that this appears to be the reason why #2878 fails the Travis build; it seems that Travis uses the rapidxml parser only, whereas my machine chooses the libxml2 parser. After merging this patch, both should pass #2878.

### DIFF
--- a/include/mapnik/xml_loader.hpp
+++ b/include/mapnik/xml_loader.hpp
@@ -28,9 +28,9 @@
 
 namespace mapnik
 {
-class xml_node;
-void read_xml(std::string const & filename, xml_node &node);
-void read_xml_string(std::string const & str, xml_node &node, std::string const & base_path="");
+class MAPNIK_DECL xml_node;
+MAPNIK_DECL void read_xml(std::string const & filename, xml_node &node);
+MAPNIK_DECL void read_xml_string(std::string const & str, xml_node &node, std::string const & base_path="");
 }
 
 #endif // MAPNIK_LIBXML2_LOADER_HPP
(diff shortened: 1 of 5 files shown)


zerebubuth:
Except this patch doesn't compile on Travis either... The error is:

    /home/travis/.ccache/tmp/xml_parser.tmp.testing-worker-linux-docker-344e9ddd-3213-linux-3.7660.ii:(.text+0x177): undefined reference to `mapnik::xml_tree::xml_tree(std::string const&)'

and similar undefined references. On my system, these symbols are defined in `src/xml_tree.os`, which is part of `libmapnik.so.3.0` and this also includes the symbols. I don't see anything in `xml_tree.cpp` or `build.py` which would indicate conditional compilation. I'm really not sure what's going on differently on Travis than on my own machine - @flippmoke: any ideas?

flippmoke:
Often when I get these sorts of undefined references its because something needs to `MAPNIK_DECL` wrapped on it. You might try compiling locally with `-fvisibility=hidden`. Then it should throw at compile time errors in linking.

flippmoke:
Continuing on this topic `bootstraph.sh` uses that as a `CUSTOM_CXX` flag. @springmeyer do we really want to do this, because then we will start requiring lots of things to use `MAPNIK_DECL` for unit tests?

zerebubuth:
I've rebased and added those visibility attributes. Fingers crossed for the test. :smile:

springmeyer:
@zerebubuth - this looks good. The last mac build issue is to due another include needed, which I'll add shortly as a followup.

springmeyer:
As far as the problem of needing to expose more classes just for unit testing - @flippmoke still mulling on this issue. For now let's just expose on a case by case basis when needed for new tests.

PR #2886, opened 2015-06-04
Don't trim whitespace in the XML loaders.

zerebubuth:
The libxml2 and ptree (a.k.a rapidxml) XML loaders were inconsistent in their handling of whitespace:

* Rapidxml trimmed some, but not did not process CDATA child
  nodes.
* Libxml2 trimmed all, due to manual application of the trim
  function.

As discussed in https://github.com/mapnik/mapnik/pull/2878#issuecomment-108728845
it would be preferable not to trim. However, simply disabling trimming breaks CSV reading in many places, as we have lots of test files like this:

```xml
    <Layer name="frame" srs="+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs">
        <StyleName>frame</StyleName>
        <Datasource>
            <Parameter name="inline">
wkt|name
Polygon((-75.0 -75.0, -75.0 75.0, 35.0 75.0, 35.0 -75.0, -75.0 -75.0))|bounds
            </Parameter>
        </Datasource>
    </Layer>
```

This is due to turning off trimming _entirely_ - if we leave trimming on in rapidxml via `rapidxml::parse_trim_whitespace` (added in 95d5b73d) then we are only left with the case that rapidxml does not trim whitespace inside of a `<![CDATA[ ]]>' node, which breaks round-tripping the "empty_parameter2" test case. Libxml has no equivalent setting; whitespace trimming is either done all or nothing, manually and without regard to whether it came in a CDATA node or not.

However, failures are confined to the text node formatting and CSV plugin's inline functionality.

Therefore, if a manual call to trim is made for CSV inline handling and all-whitespace child nodes are dropped in the text node formatting, then all the tests pass (for me, let's see what Travis has to say...). Note that we still don't have full test coverage, so it's possible that other behaviour may have changed.

### DIFF
--- a/include/mapnik/util/trim.hpp
+++ b/include/mapnik/util/trim.hpp
@@ -92,6 +92,13 @@ static inline void unquote(std::string & s)
     s.erase(std::find_if(s.rbegin(), s.rend(), not_quoted).base(), s.end());
 }
 
+// returns true if the string `s` consists entirely of whitespace
+// characters, as defined by the `not_whitespace` function. returns
+// false if any character is not whitespace.
+static inline bool is_whitespace(std::string const &s) {
+  return !std::any_of(s.begin(), s.end(), not_whitespace);
+}
+
 }} // end of namespace mapnik
 
 #endif // MAPNIK_TRIM_HPP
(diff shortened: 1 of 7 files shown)


springmeyer:
did not have the bandwidth to review and properly test before 3.0.0, but will plan to hit this soon after.

PR #2892, opened 2015-06-08
fix rendering raster without scaling

talaj:
This patch fixes bug which causes raster not being rendered under specific circumstances:
* Raster resolution matches map resolution.
* Raster is rendered from (0, 0) positon.
* Raster is single-band type.

Test case is here https://github.com/mapnik/test-data-visual/pull/10.
[Here](https://github.com/mapnik/mapnik/blob/51172c8fdfdf6432a204758fc916dac228c7a2b6/include/mapnik/renderer_common/process_raster_symbolizer.hpp#L239-L242) is the place causing the problem.

I also cannot fully understand following condition:

```c++
if ( (std::fabs(image_ratio_x - 1.0) <= eps) &&
     (std::fabs(image_ratio_y - 1.0) <= eps) &&
     (std::abs(start_x) <= eps) &&
     (std::abs(start_y) <= eps) )
```
Isn't `(std::abs(start_x) <= eps) && (std::abs(start_y) <= eps)` superfluous here? This condition limits the optimization only to the first tile. Or do I miss something?

### DIFF
--- a/include/mapnik/renderer_common/process_raster_symbolizer.hpp
+++ b/include/mapnik/renderer_common/process_raster_symbolizer.hpp
@@ -53,7 +53,8 @@ struct image_dispatcher
                           double scale_x, double scale_y,
                           scaling_method_e method, double filter_factor,
                           double opacity, composite_mode_e comp_op,
-                          raster_symbolizer const& sym, feature_impl const& feature, F & composite, boost::optional<double> const& nodata)
+                          raster_symbolizer const& sym, feature_impl const& feature,
+                          F & composite, boost::optional<double> const& nodata, bool scale)
         : start_x_(start_x),
           start_y_(start_y),
           width_(width),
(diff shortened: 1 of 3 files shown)


springmeyer:
> I also cannot fully understand following condition

@talaj - I added that in aaae8b1cbae9fab7903a912e8f2a2d48fe957b84#diff-e6b967728b887a3988f89dd623d07554. The idea was that scaling is expensive and when the source image is the same exact dimensions as the target we should skip scaling. This does not happen often when rendering directly from GDAL but should be the case everytime when rendering via a vector tile (with an embedded raster, but that raster will have been pre-scaled). So, that was the motivation, not sure if the condition is correct - there has been a lot of refactoring in that code since the original optimization landed. I'll add that I noticed this optimization when working on a separate performance issue: https://github.com/springmeyer/mapnik-raster-overzoom



https://github.com/springmeyer/mapnik-raster-overzoom

talaj:
I see, it really does make sense in the context of vector tiles. However, I'm concerned this optimization is broken from https://github.com/mapnik/mapnik/commit/bf9c99e3d30bebcca7be7bc43665e5a697ef6fda, where rendering is skipped for single-band raster:

```diff
                 if ( (std::fabs(image_ratio_x - 1.0) <= eps) &&
                      (std::fabs(image_ratio_y - 1.0) <= eps) &&
                      (std::abs(start_x) <= eps) &&
                      (std::abs(start_y) <= eps) )
                 {
-                    composite(source->data_, comp_op, opacity, start_x, start_y);
+                    if (source->data_.is<image_data_32>())
+                    {
+                        composite(util::get<image_data_32>(source->data_), comp_op, opacity, start_x, start_y);
+                    }
                 }
                 else
                 {
```

That just occurred to me with "single color" layers like hillshade or bathymetry.

springmeyer:
> However, I'm concerned this optimization is broken from bf9c99e,

Okay, assigning to 3.x. Although I may not have time to look at this before the weekend. As far as 3.x release we are aiming for next tuesday now.

talaj:
I have added test for this in https://github.com/mapnik/test-data-visual/commit/84732f56308171b94c79bf21b04be535c5951160.

PR #2898, opened 2015-06-09
Offset Fix

flippmoke:
Fixes offsets but does not change direction of offsets. /cc @springmeyer @artemp


springmeyer:
followups at https://github.com/mapnik/mapnik/issues/2899

springmeyer:
per https://github.com/mapnik/mapnik/issues/2899#issuecomment-110407769 this needs to be reverted until ready for master. I propose not a complete revert but rather:

  - [ ] creating `offset_converter2` and putting the work-in-progress improvements into that `struct`
  - [ ] restoring `offset_converter` to use the same code as pre this merge
  - [ ] restoring visual tests to expect `offset_converter` for now.
  - [ ] add tests that replicate problems with `offset_converter2`: #2900
  - [ ] new branch that tries to fix `offset_converter2` and only after fixed would it become the default

flippmoke:
I do not think an offset_converter2 is required, as this has now been fixed.

springmeyer:
We missed updating the visual after this so master has been failing on travis since. Just hit that in 62b0d2b. As per voice with @flippmoke - tracking next actions on offsets at #2904

PR #2906, opened 2015-06-11
Generate python binding docs using Sphinx

danpat:
Example output:

```
http://danpat.net/mapnik-python-3.0.0-pre/html/
```

Uses the doc theme from readthedocs.org.  Also imports and converts the `mapnik-reference` JSON file to flesh out the docs a bit.

Currently only covers auto-generated documentation (Python API, JSON reference).  Could be fleshed out with a lot more examples.

Python docstrings need to be reviewed.  We should also remove classes for symbolizers if they're not going to be supported, or at least add some docstrings to indicate that the property reference can be found in the "XML Reference" section.


danpat:
Stale

Issue #2893, opened 2015-06-08
Remove numeric_cast from `image_util.cpp`


springmeyer:
`image_util.cpp` using `numeric_cast` in performance critical places `fill` mostly but also `get_pixel`. This needs to be removed to avoid code bloat from exception handling in functions that should otherwise be simple and easy for the compiler to inline.

artemp:
@springmeyer @flippmoke 

`fill` is less problematic because `numeric_cast<>` is called only once before applying result to all pixels.

'get/set_pixel' are potentially performance issue even so boost docs suggest that implementation should not incur runtime overhead when input doesn't overflow. 

`boost::numeric_cast' is policy based but I don't see any simple ways to implement something like 

``` 
if (range_check == cNegOverflow)  return std::numeric_limits<T>::min() ;
else if (range_check == cPosOverflow) return std::numeric_limits<T>::max() ;
else return static_cast<T>(input);
```

Maybe we can borrow some range checking facilities from `numeric_cast` and implement our own `pixel_cast`. Alternatively we can just use `static_cast<T>`  ??

artemp:
@springmeyer @flippmoke 

I'm going to settle on 
``` c++
template <typename T, typename S>
inline T pixel_cast(S s)
{
    using namespace boost::numeric;
    if (s > bounds<T>::highest() )
    {
        return bounds<T>::highest();
    }
    else if (s < bounds<T>::lowest())
    {
        return bounds<T>::lowest();
    }
    else return static_cast<T>(s);
}
```

artemp:
96f7120ecc11c81ebe90c91d2db9757a844ef2eb - replaces all numeric_cast instances with pixel_cast

springmeyer:
@artemp above plan is great except that `s > bounds<T>::highest()` fails unpredictably when the sign of `S` and `T` differ.

springmeyer:
d29a0f1 should fix the problem - @artemp can you review and merge if it looks good? After that ping @BergWerkGIS to rebuild mapnik packages so we can get unblocked tomorrow on https://github.com/mapnik/node-mapnik/issues/461.

It looks like this was not failing in core because we did not have much test coverage of `mapnik::get_pixel` on anything except colors and not on potentially out of range `set_pixel` values. But @flippmoke added pretty comprehensive tests for this in node-mapnik, and that is what caught this: https://github.com/mapnik/node-mapnik/blob/master/test/image.test.js#L339-L449 :ok_hand:

[...]

PR #2908, opened 2015-06-12
Fix pixel cast

springmeyer:
My attempt to address https://github.com/mapnik/mapnik/issues/2893#issuecomment-111386739

/cc @artemp for review, fixups, merge?

BTW, I tried to use `constexpr boost::integer_traits<type>::const_max` but failed to get that working in a generic way because we also need to support doubles. But using something that can be `consexpr` should help performance.

Also, I did some benchmarking when testing this and could not pick up a difference between this implementation and `boost::numeric_cast` for the non-throwing case. But of course when things overflow and `numeric_cast` throws performance drops horrendously. So, my sense is we basically want all the numeric cast offers except the throw but I also can't see any better way to get that then this pixel_cast approach.

### DIFF
--- a/include/mapnik/pixel_cast.hpp
+++ b/include/mapnik/pixel_cast.hpp
@@ -24,122 +24,139 @@
 #define MAPNIK_PIXEL_CAST_HPP
 
 #include <type_traits>
-#include <boost/numeric/conversion/bounds.hpp>
 
 namespace mapnik {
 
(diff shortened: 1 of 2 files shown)


artemp:
some reading for later 

http://jwwalker.com/pages/safe-compare.html
http://craighenderson.co.uk/papers/numeric_compare/

Issue #2937, opened 2015-07-01
text-spacing causes 100% CPU usage and rendering never finishes


pnorman:
Adding `text-spacing: 400;` to [openstreetmap-carto admin text](https://github.com/gravitystorm/openstreetmap-carto/blob/master/admin.mss#L119-L127) causes Mapnik to take 100% CPU and never finish rendering.

The original MSS is
```mss
#admin-text[zoom >= 16] {
  text-name: "[name]";
  text-face-name: @book-fonts;
  text-fill: @admin-boundaries;
  text-halo-radius: 1.0;
  text-halo-fill: rgba(255,255,255,0.6);
  text-placement: line;
  text-dy: -10;
}
```

Adding `text-spacing: 400;` works fine in 2.2.0.

Is my best approach of attack to profile to see where it's stuck, or work on the style side?

pnorman:
Additional notes so I can reproduce later

area is http://www.openstreetmap.org/#map=17/49.28231/-123.21202
No time is being spent in postgres

springmeyer:
@pnorman - this is severe, thanks for catching it.

> Is my best approach of attack to profile to see where it's stuck, or work on the style side?

Ideally try to reduce it down to just a single geometry the dump that geom as WKT or WKB. Then we'll be able to create a ["perfect" testcase](https://github.com/mapnik/mapnik/wiki/A-perfect-testcase) and use that as a regression once fixed.

As far as fixing I bet @talaj or @artemp will be able to see the fix quickly in the morning if you can provide a testcase that does not depend on postgres.

pnorman:
Working on reducing it now, but tackling it from the code side, there's a bunch of threads in `/lib/libthr.so.3` and two that seem stuck in mapnik

```
Thread 6 (Thread 803007c00 (LWP 101440/node)):
#0  0x00000008049d9c76 in mapnik::vertex_cache::get_offseted () from /home/pnorman/mapnik/lib/libmapnik.so.3.0
#1  0x00000008049e8fb4 in mapnik::placement_finder::single_line_placement () from /home/pnorman/mapnik/lib/libmapnik.so.3.0
[...]

PR #2941, opened 2015-07-01
Fix situation where offset_converter might start off with an SEG_END …

flippmoke:
…from the vertex, therefore starting invalid processing. Tests added to confirm fix. Ref #2937


pnorman:
I'm getting a rendering bug with this. This is without text-spacing or similar, so it is rendering.

image

Currently tracking down if it's in this PR, or also present in master

pnorman:
Bug is not caused by this PR

springmeyer:
@flippmoke - my understanding is that this pull was created as a guess of what the problem was in #2937. Now that you've fixed the real issue behind #2937 should this be discarded or still has value in being applied?

springmeyer:
per chat @flippmoke says still worthy to apply.

Issue #2951, opened 2015-07-06
Revisit conditional enabling of clipping


springmeyer:
Clipping should be smart to the geometry type. Currently in the code:

 - it is easy to get this wrong like clipping a polygon with the agg line clipper (maybe partially related to #2950)
 - it is also easy to forget to handle clipping for both single geometries and multi types (fixed for markers in )

This ticket stands as a reminder to consider how to clean this up to be less error prone.

PR #2955, opened 2015-07-07
Offset converter fix

### LINKED ISSUE #2950: incorrect rendering with `line-clip` + `line-offset` on polygons
This is a followup to #2942. #2942 fixed a bug that caused lines to be draw between interior rings and another ring. But another more minor bug still exists when:

  - polygon data is used
  - line symbolizer is used
  - with offsets
  - `line-clip:true` (which is not the default)

The bug is evident in any of the `multipolygon-two-holes-*` tests. It manifests slightly differently in the agg and cairo renderer.

Expected (seen with `line-clip:false`):

multipolygon-two-holes-geojson-300-300-2 0-cairo-reference

Poor rendering (seen with `line-clip:true` which in this case should not change the rendering):

multipolygon-two-holes-geojson-300-300-2 0-cairo

One patch I tried that does not fix the problem:

```diff
diff --git a/src/agg/process_line_symbolizer.cpp b/src/agg/process_line_symbolizer.cpp
index e2004e9..e4b20ce 100644
--- a/src/agg/process_line_symbolizer.cpp
+++ b/src/agg/process_line_symbolizer.cpp
@@ -31,6 +31,8 @@
 #include <mapnik/vertex_processor.hpp>
 #include <mapnik/renderer_common/clipping_extent.hpp>
 #include <mapnik/renderer_common/apply_vertex_converter.hpp>
+#include <mapnik/geometry_type.hpp>
+
 // agg
 #include "agg_basics.h"
 #include "agg_rendering_buffer.h"
@@ -164,12 +166,19 @@ void agg_renderer<T0,T1>::process(line_symbolizer const& sym,
         rasterizer_type ras(ren);
         set_join_caps_aa(sym, ras, feature, common_.vars_);
 
-        using vertex_converter_type = vertex_converter<clip_line_tag, transform_tag,
+        using vertex_converter_type = vertex_converter<clip_line_tag, clip_poly_tag,transform_tag,
                                                        affine_transform_tag,
                                                        simplify_tag, smooth_tag,
                                                        offset_transform_tag>;
         vertex_converter_type converter(clip_box,sym,common_.t_,prj_trans,tr,feature,common_.vars_,common_.scale_factor_);
-        if (clip) converter.set<clip_line_tag>(); // optional clip (default: true)
+        if (clip)
+        {
+            geometry::geometry_types type = geometry::geometry_type(feature.get_geometry());
+            if (type == geometry::geometry_types::Polygon)
+                converter.template set<clip_poly_tag>();
+            else if (type == geometry::geometry_types::LineString)
+                converter.template set<clip_line_tag>();
+        }
         converter.set<transform_tag>(); // always transform
         if (std::fabs(offset) > 0.0) converter.set<offset_transform_tag>(); // parallel offset
         converter.set<affine_transform_tag>(); // optional affine transform
@@ -183,14 +192,21 @@ void agg_renderer<T0,T1>::process(line_symbolizer const& sym,
     }
     else
     {
-        using vertex_converter_type = vertex_converter<clip_line_tag, transform_tag,
+        using vertex_converter_type = vertex_converter<clip_line_tag, clip_poly_tag, transform_tag,
                                                        affine_transform_tag,
                                                        simplify_tag, smooth_tag,
                                                        offset_transform_tag,
                                                        dash_tag, stroke_tag>;
         vertex_converter_type converter(clip_box, sym,common_.t_,prj_trans,tr,feature,common_.vars_,common_.scale_factor_);
 
-        if (clip) converter.set<clip_line_tag>(); // optional clip (default: true)
+        if (clip)
+        {
+            geometry::geometry_types type = geometry::geometry_type(feature.get_geometry());
+            if (type == geometry::geometry_types::Polygon)
+                converter.template set<clip_poly_tag>();
+            else if (type == geometry::geometry_types::LineString)
+                converter.template set<clip_line_tag>();
+        }
         converter.set<transform_tag>(); // always transform
         if (std::fabs(offset) > 0.0) converter.set<offset_transform_tag>(); // parallel offset
         converter.set<affine_transform_tag>(); // optional affine transform
```

artemp:
Fixes #2950

### DIFF
--- a/src/agg/process_line_symbolizer.cpp
+++ b/src/agg/process_line_symbolizer.cpp
@@ -31,6 +31,7 @@
 #include <mapnik/vertex_processor.hpp>
 #include <mapnik/renderer_common/clipping_extent.hpp>
 #include <mapnik/renderer_common/apply_vertex_converter.hpp>
+#include <mapnik/geometry_type.hpp>
 // agg
 #include "agg_basics.h"
 #include "agg_rendering_buffer.h"
@@ -164,12 +165,19 @@ void agg_renderer<T0,T1>::process(line_symbolizer const& sym,
         rasterizer_type ras(ren);
         set_join_caps_aa(sym, ras, feature, common_.vars_);
 
-        using vertex_converter_type = vertex_converter<clip_line_tag, transform_tag,
+        using vertex_converter_type = vertex_converter<clip_line_tag, clip_poly_tag, transform_tag,
                                                        affine_transform_tag,
                                                        simplify_tag, smooth_tag,
                                                        offset_transform_tag>;
         vertex_converter_type converter(clip_box,sym,common_.t_,prj_trans,tr,feature,common_.vars_,common_.scale_factor_);
-        if (clip) converter.set<clip_line_tag>(); // optional clip (default: true)
+        if (clip)
+        {
+            geometry::geometry_types type = geometry::geometry_type(feature.get_geometry());
+            if (type == geometry::geometry_types::Polygon)
+                converter.template set<clip_poly_tag>();
+            else if (type == geometry::geometry_types::LineString)
+                converter.template set<clip_line_tag>();
+        }
         converter.set<transform_tag>(); // always transform
         if (std::fabs(offset) > 0.0) converter.set<offset_transform_tag>(); // parallel offset
         converter.set<affine_transform_tag>(); // optional affine transform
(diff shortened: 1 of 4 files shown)


src/agg/process_line_symbolizer.cpp
```diff
@@ -164,12 +165,19 @@ void agg_renderer<T0,T1>::process(line_symbolizer const& sym,
...
                                                        offset_transform_tag>;
         vertex_converter_type converter(clip_box,sym,common_.t_,prj_trans,tr,feature,common_.vars_,common_.scale_factor_);
-        if (clip) converter.set<clip_line_tag>(); // optional clip (default: true)
+        if (clip)
+        {
+            geometry::geometry_types type = geometry::geometry_type(feature.get_geometry());
+            if (type == geometry::geometry_types::Polygon)
+                converter.template set<clip_poly_tag>();
+            else if (type == geometry::geometry_types::LineString)
+                converter.template set<clip_line_tag>();
```
springmeyer: This needs to handle MultiPolygon and MultiLineString, right? refs #2951

artemp: yes!

springmeyer:
Looks great. Added more test coverage for this in https://github.com/mapnik/test-data-visual/commit/a4038f4b09901ebff4d58fa8593f5df1a78c1abe. If you look really close at the images you can see some minor differences in line endings between cairo and agg and a difference in the winding order of the hole-inside-the-hole for geojson vs shape/ogr. But these issues are minor enough I'm not planning on ticketing now.

PR #2966, opened 2015-07-12
Allow mapnik-json and mapnik-wkt to be built as shared libraries

tomhughes:
This is an initial attempt at allowing the `mapnik-json` and `mapnik-wkt` libraries to be built as shared libraries, mostly by copying the way the main library is built.

That does mean that this introduces quite a lot of duplication, which somebody more familiar with scons could likely avoid by factoring out the duplicated code into a common function somewhere that knows how to do all the magic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,13 +17,13 @@ python:
 	make
 	python bindings/python/test/visual.py -q
 
-src/json/libmapnik-json.a:
+src/json/libmapnik-json.so:
 	# we first build memory intensive files with -j1
 	$(PYTHON) scons/scons.py -j1 \
 		--config=cache --implicit-cache --max-drift=1 \
 		src/renderer_common/process_group_symbolizer.os \
-		src/json/libmapnik-json.a \
-		src/wkt/libmapnik-wkt.a \
+		src/json/libmapnik-json.so \
+		src/wkt/libmapnik-wkt.so \
 		src/css_color_grammar.os \
 		src/expression_grammar.os \
 		src/transform_expression_grammar.os \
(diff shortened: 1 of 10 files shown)


springmeyer:
@tomhughes - I'd prefer to keep this as only static libraries. I shared my reasoning over at https://github.com/mapnik/python-mapnik/issues/30#issuecomment-120270946. If you see a strong need for this, let me know. Otherwise can we just stick with static only?

tomhughes:
Sorry, I think I must have missed your reply on this for some reason...

My purpose of course if the Fedora packaging, specifically https://fedoraproject.org/wiki/Packaging:Guidelines#Packaging_Static_Libraries which gives the policy on shipping static libraries.

One solution might just be not to ship them at all I think? I mean they're linked in to some of the plugins during the build but normal building against mapnik doesn't require them? or does it?

springmeyer:
> but normal building against mapnik doesn't require them

Correct, libmapnik build and link does not depend on them. That is the reason they are separate (to avoid the extra binary size).

External applications that want json and wkt parsing and serialization support then have two options:

1) include the code as "header only" by doing:

```
#include <mapnik/json/feature_generator_grammar_impl.hpp>
```

Then linking `libmapnik-json` is uneeded

2) Linking `libmapnik-json`

And only including:

```
#include <mapnik/json/feature_generator_grammar.hpp>
```

The advantage of this is a speedup in compile and link times.

tomhughes:
In that case I think packaging them in a separate `mapnik-static` package will work, so I'll close this.

Issue #2980, opened 2015-07-16
Move occi, osm, and rasterlite plugins out of core


springmeyer:
We currently have 3 unmaintained plugins in core: `occi`, `rasterlite`, and `osm`.

Breaking these down:

  - `osm` is unmaintained because the origin author is gone and the plugin is of dubious value. Because OSM is best pre-processed before reading with Mapnik this plugin has never been a good solution and is more of a trap than anything. The only reason I've not removed it sooner is that I've pondered re-writing it from scratch using https://github.com/osmcode/libosmium. However I don't have time to do this, so the best solution to avoid confusion is to remove the plugin.

  - `occi` is an excellent plugin written by @kunitoki. However, it frequently breaks because:
   - it has no unit tests
   - none of the active core devs of Mapnik has knowledge of how to install oracle databases
   - it cannot be tested *currently* on travis.

This `occi` plugin deserves love and attention, testing, and more development. But it still does not make sense in Mapnik core because otherwise we'll just be releasing a potentially broken plugin. Outside of core it can be versioned and updated whenever development effort is available by @kunitoki or others that are interested

  - `rasterlite` should be removed because it depends on a third party library that is deprecated and no longer supported - refs https://github.com/mapnik/mapnik/issues/2977

springmeyer:
closed by #2981

am2222:
@springmeyer Hi,
Can you please tack a look at this [topic ](https://github.com/mapnik/mapnik/issues/3741) as I am planning to create osm input plugin using osmium but I am faced some problems with scons and library integrations, I'd be thankful if you please give me some advices about it.
Thanks

artemp:
@am2222 -` libosmium` is a header only library so there's nothing to build https://github.com/osmcode/libosmium. For the plug-in you can use `SCons` - you should be able to get some ideas from existing plug-ins (look for `build.py` files). If you have some more specific questions, please, provide more details. Cheers.

kunitoki:
it's a pity the occi plugin has been put in external stage area as it was a great piece of code actually outperforming postgres in some areas (if you can afford an oracle database of course). but i have no time into maintaining it so if nobody can take ownership...

am2222:
[...]

PR #2981, opened 2015-07-16
move occi, osm, rasterlite to non-core repo

springmeyer:
These plugins addded to non-core repo in https://github.com/mapnik/non-core-plugins/commit/84b0bfecb36bea2e7a61c5bc21609cef2685058a

refs #2980


kunitoki:
makes sense. now that they live in non mapnik master repo i would like to have my name on the occi plugin (since Artem has not created those, and putting my name on the copyright of the mapnik repo was an issue before).

Issue #2963, opened 2015-07-09
Performance issues after removing !detector_.extent().intersects(box)


talaj:
Unfortunately https://github.com/mapnik/mapnik/pull/2949 is not always performance improvement. I have encountered map with four times longer rendering time, most of it spent in collision detection:

```
  48.20%      run  libmapnik.so              [.] mapnik::placement_finder::collision(mapnik::box2d<double> const&, icu_48::UnicodeString const&, bool) const
  37.17%      run  libmapnik.so              [.] mapnik::box2d<double>::intersects(mapnik::box2d<double> const&) const
   4.64%      run  libharfbuzz.so.0.935.0    [.] 0xaa01
   1.02%      run  libcrypto.so.1.0.0        [.] 0xbd4d7
   0.50%      run  libc-2.13.so              [.] _int_malloc
   0.46%      run  libz.so.1.2.7             [.] 0xba0e
   0.28%      run  libmapnik.so              [.] agg::pod_bvector<mapnik::svg::path_attributes, 6u>::allocate_block(unsigned int)
   0.27%      run  libmapnik.so              [.] mapnik::gradient::gradient()
   0.23%      run  libm-2.13.so              [.] __ieee754_atan2
   0.20%      run  libfreetype.so.6.8.1      [.] 0x50285
   0.18%      run  libc-2.13.so              [.] _int_free
   0.14%      run  libc-2.13.so              [.] malloc_consolidate
   0.14%      run  libc-2.13.so              [.] free
   0.13%      run  libc-2.13.so              [.] __memcpy_ssse3
   0.12%      run  libc-2.13.so              [.] malloc
   0.12%      run  libpq.so.5.4              [.] 0x14091
   0.12%      run  libmapnik.so              [.] mapnik::placement_finder::single_line_placement(mapnik::vertex_cache&, mapnik::enumeration<mapnik::text_uprig
   0.09%      run  libmapnik.so              [.] mapnik::quad_tree<mapnik::label_collision_detector4::label>::query_node(mapnik::box2d<double> const&, std::ve
   0.08%      run  libm-2.13.so              [.] __sincos
   0.07%      run  libc-2.13.so              [.] __memset_sse2
``` 

With https://github.com/mapnik/mapnik/pull/2949 reverted:

```
  28.19%      run  libharfbuzz.so.0.935.0    [.] 0x1aa61
   7.69%      run  libcrypto.so.1.0.0        [.] 0xbcb16
[...]

PR #2988, opened 2015-07-20
do not store out of extent collision boxes

talaj:
Possible solution for https://github.com/mapnik/mapnik/issues/2963.

### DIFF
--- a/include/mapnik/label_collision_detector.hpp
+++ b/include/mapnik/label_collision_detector.hpp
@@ -212,12 +212,18 @@ public:
 
     void insert(box2d<double> const& box)
     {
-        tree_.insert(label(box), box);
+        if (tree_.extent().intersects(box))
+        {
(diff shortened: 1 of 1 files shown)


springmeyer:
confirmed the speedup locally:

before:

```
$ time ./test/visual/run  collision-extent
Visual rendering: 0 failed / 8 passed / 0 overwritten / 0 errors

real	0m7.225s
user	0m6.858s
sys	0m0.320s
```

after:

```
$ time ./test/visual/run  collision-extent
real	0m1.325s
user	0m1.302s
sys	0m0.016s

```

springmeyer:
Based on the discussion at https://github.com/mapnik/mapnik/issues/2963#issuecomment-122877101 I think this fix makes sense performance wise. And only one minor visual test failure results, which I will update after merging. Great work @talaj!

PR #3003, opened 2015-07-27
SVG parser refactoring and improvements

springmeyer:
- Drops requirement on libxml2 depedency
 - Switches to `ptree` (rapidxml inside boost::property_tree) as default xml parser
 - refactors SVG parser to collect errors
 - Adds more test coverage

Followups:

 - https://github.com/mapnik/mapnik-packaging/pull/193


src/svg/svg_parser.cpp
```diff
@@ -148,540 +164,497 @@ bool parse_style (const char* str, pairs_type & v)
...
+            if (!mapnik::svg::parse_path(value, parser.path_))
             {
-                xmlFree(value);
-                xmlChar *id_value;
-                id_value = xmlTextReaderGetAttribute(reader, BAD_CAST "xml:id");
-                if (!id_value) id_value = xmlTextReaderGetAttribute(reader, BAD_CAST "id");
-                if (id_value)
+                auto const* id_attr = node->first_attribute("xml:id");
+                if (id_attr != nullptr) id_attr = node->first_attribute("id");
+                if (id_attr)
```
jakepruitt: Just to clarify, the way this code works now is that the path must have an `xml:id` property **and** and `id` property in order to have the error message print with the id. Is this the expected behavior? According to the [SVG spec](http://www.w3.org/TR/SVGMobile12/struct.html#idAttrs) it seems like there will only ever be `id` or `xml:id` on an element, and never both.

In that case, `if (id_attr != nullptr)` on line 493 should be `if (id_attr == nullptr)`, unless I'm reading the spec wrong.

jakepruitt: Based on [line 808](https://github.com/mapnik/mapnik/pull/3003/files#diff-8b937fdb27d4bb8dd67a191b68511c25R808), I believe this is a typo. Going to fix and adjust fixtures.

src/marker_cache.cpp
```diff
@@ -175,7 +175,15 @@ std::shared_ptr<mapnik::marker const> marker_cache::find(std::string const& uri,
             svg_path_adapter svg_path(stl_storage);
             svg_converter_type svg(svg_path, marker_path->attributes());
             svg_parser p(svg);
-            p.parse_from_string(known_svg_string);
+
+            if (!p.parse_from_string(known_svg_string))
+            {
+                for (auto const& msg : p.error_messages())
+                {
+                    std::cerr << "SVG PARSING ERROR:\"" << msg << "\"" << std::endl;
```
jakepruitt: @artemp Should these be going to `std::cerr` or `MAPNIK_LOG_ERROR(svg_parser)`? I believe this is the cause of #3007.

test/unit/svg/svg_parser_test.cpp
```diff
@@ -291,5 +554,235 @@ TEST_CASE("SVG parser") {
...
+        auto const& attrs = storage->attributes();
+        REQUIRE(attrs.size() == 1 );
+        REQUIRE(attrs[0].fill_gradient.get_gradient_type() == mapnik::RADIAL);
+        REQUIRE(attrs[0].fill_gradient.has_stop());
+        attrs[0].fill_gradient.get_control_points(x1, y1, x2, y2, r);
+        REQUIRE(x1 == 0);
+        REQUIRE(y1 == 25);
+        REQUIRE(x2 == 10);
+        REQUIRE(y2 == 10);
+        REQUIRE(r == 75);
```
jakepruitt: Question, @artemp. The gradient fixture for this test has the following format
```
<radialGradient id="grad1b" gradientUnits="userSpaceOnUse" cx="10%" cy="10%" r="75%" fx="00.00%" fy="25%">
```
This translates to the above values when `get_control_points` is called. Is this the expected value? Should the values be `0, 0.25, 0.10, 0.10, 0.75` respectively due to the percentage conversion?

src/marker_cache.cpp
artemp: @jakepruitt - yes, fixed in 3a4d57492fe68d3afe4ce2b02344dd5c0a78e1ed. I also made `marker_cache` to return `marker_null()` if case of any parsing/values errors. I think this is desired logic :  only accept SVG we can interpret correctly and fail otherwise + logging errors.

test/unit/svg/svg_parser_test.cpp
artemp: @jakepruitt - great catch! this was a bug in grammar - fixed in 42bf2303e0dd4423c33a9b00208f0f33c281175d

artemp:
#### Summary of main changes

* added container to log SVG parsing errors  
* reimplemented to use rapidxml for parsing XML (DOM)
* support both `xml:id` and `id` attributes ( `xml:id` takes precedence )
* added `parse_id_from_url` using boost::spirit
* added error tracking when parsing doubles
* unit tests for svg_parser to improve coverage  
* fixed rx/ry validation for rounded_rect
* fixed dimensions parsing
* remove libxml2 dependency

PR #3013, opened 2015-08-02
Bugfix: dot symbolizer not acknowledging scale factor

brianreavis:
**Before (1x/2x)**
dots-500-100-2 0-agg-reference
dots-500-100-2 0-agg-reference
**After (1x/2x)**
dots-500-100-2 0-agg-reference
dots-500-100-2 0-agg


brianreavis:
@springmeyer @flippmoke Just making sure this is on your radar. No huge rush though, we're using SVGs as an alternative for now.

springmeyer:
@brianreavis - great, thanks for catching this. I'll plan to merge asap.

brianreavis:
@springmeyer @artemp Thanks!

PR #3023, opened 2015-08-10
Expose Image Filters with function on Image Object

flippmoke:
This is a pull request to allow for exposing of image filters as a method outside of rendering.


artemp:
@flippmoke  - looks good. We need to add non-mutating version though:

```c++
auto filtered = filter(image, filter_string)
```

Issue #3069, opened 2015-09-14
Zero allocation constructor for mapnik::image


springmeyer:
We previously, in 3.x-dev. supported creating a mapnik image via an already allocated chunk of memory. 

Bringing back this support will enable mapnik image encoding to operate on pixels previously allocated by other applications in the same process. This is nice for zero-copy perf but also for decoupling - those applications don't need to compile against mapnik headers to be able to pass their buffer and benefit from its encoding (this is targeting node.js).

Assumptions:

 - pixel packing of the calling application will match the way mapnik stores pixels (arbg).
 - we only care about rgba

artemp:
looks good, closing

jfirebaugh:
I take it there's nothing that could usefully made async in `Image.fromPixelsSync`?

@springmeyer / @artemp can you clarify "the way mapnik stores pixels (arbg)" vs "we only care about rgba"? I believe the mapbox-gl-native format is rgba.

flippmoke:
@jfirebaugh ABGR is the format most uint32 pixels are stored and that is the case for mapnik. This is very common for many libraries and so we leverage it commonly when using outside libraries for algorithms. If the data is in RGBA format actually (we often say stuff is RGBA but it actually stores in the order of ABGR internally), then it would break the format of all our internally optimized operations in many cases and we would either have to use something like boost GIL to make things operate correctly. This would not be an easy task to do currently as it would involve many changes to the way we use images (though we would love to support stuff like this eventually).

artemp:
@jfirebaugh - arbg (typo) --> abgr

jfirebaugh:
Thanks @flippmoke @artemp -- mapbox-gl-native is probably ABGR too.

PR #3071, opened 2015-09-15
re-enable zero allocation image interface - refs #3069

artemp:


artemp:
@springmeyer - looks good.

PR #3075, opened 2015-09-18
fix parsing colors in hexadecimal notation

talaj:
Hexadecimal colors needs to be considered a parsing primitives. 

If hexa colors are not primitives, `#000000 100%` is parsed as `#00000010` and `0%` in this example:

```
image-filters="colorize-alpha(#000000 0%, #000000 100%)"
```


artemp:
@talaj - where is `hexa` colour format described, could you post a link pls ?

PR #3089, opened 2015-09-29
Csv disk index

artemp:


plugins/input/csv/csv_datasource.cpp
```diff
@@ -596,6 +608,12 @@ mapnik::featureset_ptr csv_datasource::features(mapnik::query const& q) const
                 return std::make_shared<csv_inline_featureset>(inline_string_, locator_, separator_, headers_, ctx_, std::move(index_array));
             }
         }
+        else if (has_disk_index_)
+        {
+            std::cerr << "DISK_INDEX" << std::endl;
```
springmeyer: @artemp - remove this debug print?

plugins/input/csv/csv_index_featureset.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+mapnik::feature_ptr csv_index_featureset::next()
+{
+    /*
+    if (row_limit_ && count_ >= row_limit_)
+    {
+        return feature_ptr();
+    }
+    */
```
springmeyer: @artemp why is `row_limit` commented?

springmeyer:
@artemp - looking really good / promising. What do you think about a single `mapnik-index` command that works to create `.index` for both shapefiles and csv (and more, like geojson, in the future)? What would it take to provide that?

plugins/input/csv/csv_datasource.cpp
artemp: @springmeyer - yes.

plugins/input/csv/csv_index_featureset.cpp
artemp: @springmeyer - do we really need row_limit for disk-index access ? [this-is-work-in-progress]

PR #3099, opened 2015-10-01
remove Feature alias of feature_impl

talaj:
I was just passing by the `// TODO - remove at Mapnik 3.x` comment so I removed it as @springmeyer proposed in https://github.com/mapnik/mapnik/commit/5a6ea9ee6f297d64e0f699f362de25f7e3b2743f.

But it's still used in node-mapnik [here](https://github.com/mapnik/node-mapnik/blob/40fb7a7cc717bc45ea3dece7523e1743b12e14c4/src/mapnik_expression.cpp#L115) and probably also somewhere else, so I'm not sure whether it is a good idea.


artemp:
@talaj - good idea, we should fix node-mapnik too

PR #3100, opened 2015-10-01
use move semantics instead of shared_ptr

talaj:
Please correct me if I'm wrong, but this `shared_ptr` usage is superfluous if we have move semantics.


springmeyer:
:+1: looks good. Please merge when you consider it ready.

PR #3107, opened 2015-10-02
Add section for Python Bindings

nvkelso:
Since these aren't included by default, the other package needs to be referenced here to avoid a wild goose chase.


springmeyer:
Thank you @nvkelso!

Issue #3115, opened 2015-10-07
Postgis plugin: ability to discard key_field from attributes


springmeyer:
After adding support for `key_field` in the postgis.input plugin we gained the ability to ask the plugin to make the `feature.id()` represent some integer attribute. However the outcome is that the attribute value was then stored twice: once in the `feature.id()` value and still in feature context/attributes.

In order to create the smallest features in memory and - more importantly - the smallest serialization of features we should allow cartographers to request that the `key_field` is discarded from the attributes.

Note: users that wish to drop the `key_field` from the attributes will then be unable to later filter on the value using the attribute name. So, to compensate for this we should followup by moving on #3116.

springmeyer:
landed in #3117. Followup at #3116.

PR #3117, opened 2015-10-07
Postgis plugin: Add key_field_as_attribute 

springmeyer:
The new `key_field_as_attribute ` will default to true (to keep back compatible/existing behavior). Flipping to false will trigger the value to not be duplicated as feature attribute

- refs #3115


yhahn:
:eyes:

Issue #3113, opened 2015-10-07
scaling methods of RasterSymbolizer broken


talaj:
I have issue with one particular tiff when `scaling` is set to non-default value on `RasterSymbolizer`.

Here is simple test case:

``` xml
<Map>
    <Parameters>
        <Parameter name="sizes">512,512</Parameter>
    </Parameters>

    <Style name="raster">
        <Rule>
            <RasterSymbolizer scaling="bilinear">
                <RasterColorizer default-color="red">
                    <stop value="290" color="white"/>
                    <stop value="400" color="black"/>
                </RasterColorizer>
            </RasterSymbolizer>
        </Rule>
    </Style>

    <Layer name="dataraster">
        <StyleName>raster</StyleName>
        <Datasource>
            <Parameter name="file">../sample.tif</Parameter>
            <Parameter name="type">gdal</Parameter>
            <Parameter name="band">1</Parameter>
        </Datasource>
    </Layer>
</Map>
```

Complete test including tiff file is here: https://github.com/mapycz/test-data-visual/commit/a5416d735ee291d18d05de1d101ee5f8ed3a116a

This is how output looks **without** `scaling="bilinear"`:
![raster_symbolizer-512-512-1 0-agg](https://cloud.githubusercontent.com/assets/1950911/10338977/4df61d1e-6d09-11e5-8dda-88bc5c1586f2.png)

**With** `scaling="bilinear"`, only `default-color="red"` is visible.
![raster_symbolizer-512-512-1 0-agg-2](https://cloud.githubusercontent.com/assets/1950911/10338976/4df5c7ce-6d09-11e5-8df8-fae04330ad36.png)

The same problem occurs for any non-default value of `scaling`, so [this branch](https://github.com/mapnik/mapnik/blob/8c6bf0eef67cbf97f717bc388424d0dc5cf0c037/src/image_scaling.cpp#L160-L166) of code is affected. [These scaling traits](https://github.com/mapnik/mapnik/blob/8c6bf0eef67cbf97f717bc388424d0dc5cf0c037/include/mapnik/image_scaling_traits.hpp#L118-L127) are used.

I cannot reproduce it with other Float32 tiffs I have.

It was working fine with old versions of Mapnik before redesign of `image_data`. I was trying to bisect, but there is a lot of commits I cannot compile.

Here is some info about the tiff file:

```
Driver: GTiff/GeoTIFF
Files: sample.tiff
Size is 250, 219
Coordinate System is:
[...]

PR #3122, opened 2015-10-12
remove needless header file

talaj:
Found this abandoned header file while working on https://github.com/mapnik/mapnik/issues/3113. I can compile Mapnik without it. Does it have any purpose?


springmeyer:
@talaj - yes this can be removed. It was disabled by me in https://github.com/mapnik/mapnik/commit/4c66d6dda04da4e6d49e771a8ae626673a0423d5#diff-4ccab151c91982096bbbb37cdd512a16. I left it behind only because at that time I was still learning how to fix the premultiplied problems. I think it is good to remove now. The header originally resulted from @artemp copying it out of agg and modifying. If I recall correctly the modifications were in an attemp to fix alpha problems but the proper fix was to work consistently in premultiplied space.

Issue #3050, opened 2015-08-28
Some character combinations aren't rendered


MaxSem:
![map](https://cloud.githubusercontent.com/assets/1260606/9553174/28de0e30-4d72-11e5-9574-881291455f54.png)
Minimum repro: "א-" - that's Hebrew letter aleph and minus sign. This is reproduceable at least with Hebrew, Arabic and Georgian characters, if appropriate fonts are present in fallback chain.

In the sample attached, aleph is not properly rendered in combination with minus sign while it has no problems with being rendered alone or combined with Latin (and that's the same Unicode range as minus) or Chinese characters.

Full XML to reproduce this bug, along with fonts used and C++ and Node.js examples is at https://github.com/MaxSem/mapnik-bug

artemp:
Debugging reveals that call to ` ubidi_getDirection(bidi);` in  itemizer.cpp returns  `UBIDI_MIXED`  for  `aא+` input
This results in three text_items (correctly)
```
text_item:0,1 // LTR
text_item:1,2 // RTL
text_item:2,3 // LTR
```
and UBIDI_RTL -> only one item for `א+` 
```
text_item:0,2 // RTL
```

@MaxSem - I think the problem is 'ICU itemizer' sets RTL direction to the whole run even there is a neutral char '+' or '-' etc. Your feedback would be super useful as I'm not an expert on RTL languages.

artemp:
http://www.w3.org/International/articles/inline-bidi-markup/uba-basics

MaxSem:
I don't know anything about it either, but I have people around who do:) My colleague says that the second example is correct too.

artemp:
ICU has 
`
ubidi_setReorderingMode(..);
ubidi_setReorderingOptions(..);
` 
Passing various option combinations ^ result in "א-" rendered or not. But I don't know if we should be using non-default options or .. I feel we need an expert advice here :) 

/cc @springmeyer @behdad @MaxSem

behdad:
I have no clue what this bug is about, but playing with bidi settings sounds like the wrong approach.  It's definitely a bug in your (mapnik) code.

behdad:
> Debugging reveals that call to ubidi_getDirection(bidi); in itemizer.cpp returns UBIDI_MIXED for aא+ input
> This results in three text_items (correctly)
> 
> text_item:0,1 // LTR
> text_item:1,2 // RTL
> text_item:2,3 // LTR
> 
[...]

PR #3151, opened 2015-10-27
Issue 3050

jkroll20:


flippmoke:
I don't know this section of the code really well -- @talaj or @artemp could you review this?

flippmoke:
@jkroll20 -- small thing but could you change the code to follow the mapnik [conventions](https://github.com/mapnik/mapnik/blob/master/docs/contributing.md#coding-conventions)

jkroll20:
@flippmoke added spaces around operators. comment if you still see something that doesn't conform to mapnik conventions.

include/mapnik/text/harfbuzz_shaper.hpp
```diff
@@ -134,14 +141,16 @@ static void shape_text(text_line & line,
             double max_glyph_height = 0;
             for (unsigned i=0; i<num_glyphs; ++i)
             {
-                auto const& gpos = positions[i];
-                auto const& glyph = glyphs[i];
+                glyph_face_info const* gfi = gfaceinfos.find(i) != gfaceinfos.end()? &gfaceinfos[i]: nullptr;
```
flippmoke: Can you change this logic so that it doesn't use a raw pointer? Could just making an if statement here be more clear?

artemp:
@jkroll20 - thanks for digging this issue and the fix!  
Before merging I'd like to ensure performance is acceptable. In particular I see there's call to `std::map::find` for every glyph which for me is raising a warning. Have you done any profiling ?

include/mapnik/text/harfbuzz_shaper.hpp
jkroll20: could make the variables mutable and reassign...

jkroll20:
@artemp yes, it does the lookup for each glyph. haven't done any profiling. is it true that the number of glyphs returned by hb_buffer_get_glyph_infos is never larger than the number of characters? if so, i could use a array or vector instead of a map.

include/mapnik/text/harfbuzz_shaper.hpp
jkroll20: Removed the raw pointer and using an if statement instead of ternary operators.

jkroll20:
@artemp Changed it to use a std::vector instead of a map. If there are still performance concerns, we could profile it. Although as it is a simple table lookup now, any overhead should be negligible.

talaj:
There is basically only one thing failing in visual tests after this pull request. As you can see on following images, one glyph in front of `abcd` is missing after this change:

Here is reference image of current master:
lines-6-800-800-2 0-agg-reference

Here is image after this pull request with one glyph missing:
lines-6-800-800-2 0-agg

Is questionable what version is correct. Does anyone know?

talaj:
I have added testcase from https://github.com/mapnik/mapnik/issues/3050#issue-103767196 in https://github.com/mapnik/test-data-visual/pull/27. Thanks @MaxSem.

Before:
text-typographic-2-256-256-1 0-agg-reference
After:
text-typographic-2-256-256-1 0-agg

@jkroll20 Would be nice to have more tests for this.

jkroll20:
The problem with the missing Khmer glyph was that one glyph, which was supposed to be rendered with the Khmer face, was available in another face in the fontset, but not rendered correctly. This should be fixed now. 
I'm going to add some more tests.

jkroll20:
The cause of the Travis fail is apparently a compiler version problem (clang 3.5 doesn't support c++11).

artemp:
@jkroll20 - yep, don't worry about travis and thanks for your contribution !

springmeyer:
@jkroll20 would you mind providing another pull request adding code comments to these lines of code: The goal would be to write how things are intended to work in detail for your future self or others that might hack on that fallback code. If you have questions or confusions then add them in the comments and lets make sure to get clear on how the code works. Sound good?

springmeyer:
> providing another pull request

Whoops, said that when I though this had already merged. Fine to add to this pull @jkroll20

jkroll20:
added comments, and here's a [pull request](https://github.com/mapnik/test-data-visual/pull/30) for a test which should cover the missing khmer glyph.

Issue #3051, opened 2015-08-31
Boost.Geometry Centroid calculation exception


wlredeye:
Hi,

I'm rendering tiles using mapnik3.0.4 with OSM data. On a few tiles rendering fails with exception: 
Boost.Geometry Centroid calculation exception 
I don't have this problem with mapnik2.3 .
Tile coords example: Zoom: 16, X: 32760, Y: 21784 
We have our own style so I can't reproduce it on some small example.

Thanks

talaj:
I'm also getting this error when rendering around Greenwich :-).

Will take a look.

talaj:
So I find out that the problem is empty polygon geometry fetched by postgis datasource.

wlredeye:
So this may be postgis or postgis plugin bug?

talaj:
@wlredeye - Postgis plugin is not skipping empty geometries as other plugins does. I'm going to send a fix today.

artemp:
@talaj  :+1:

talaj:
But wait, Mapnik is always querying some bounding box, so empty geometry should never pass. But it does. Postgis bug?

Issue #3156, opened 2015-10-29
GeoJSON.input  - Avoid returning features with empty geometries


artemp:
This is precautionary measure to avoid mishandling empty geometries in user code.

PR #3164, opened 2015-11-05
check input data of centroid algorithm

talaj:
Postgis plugin should skip features with empty geometries as other plugins do ([sqlite](https://github.com/mapnik/mapnik/blob/8c6bf0eef67cbf97f717bc388424d0dc5cf0c037/plugins/input/sqlite/sqlite_featureset.cpp#L85-L88), [geojson](https://github.com/mapnik/mapnik/blob/133ca16d3a87c8a57d66b4652990593097d19a0e/plugins/input/geojson/geojson_memory_index_featureset.cpp#L78-L79)).

Fixes https://github.com/mapnik/mapnik/issues/3051.

https://github.com/mapnik/mapnik/issues/3156 is similar.

Please do not merge yet. I will make a testcase.


talaj:
Postgis plugin is always making query by bounding box thus empty geometry should never pass. However it does sometimes. I think it's a bug in postgis, but I cannot isolate it to simple testcase, it only manifests with huge osm database.

@artemp - Do you think it still makes sense to skip features with empty geometry as it's only workaround to Postgis issue? On the other side it's cheap test and the worst thing which can happen during rendering is uncaught exception.

I would rather fix wkb parsing that way empty geometry is really parsed as `mapnik::geometry::geometry_empty` because currently empty polygon is parsed as `mapnik::geometry::polygon`, see [`wkb_reader`](https://github.com/mapnik/mapnik/blob/8c6bf0eef67cbf97f717bc388424d0dc5cf0c037/src/wkb.cpp#L120-L198). Such solution would prevent centroid beeing computed: [geometry_centroid](https://github.com/mapnik/mapnik/blob/b2c85e0c65b9386fa449362888350d33ca54942c/include/mapnik/geometry_centroid.hpp#L48-L51).

artemp:
@talaj - I think fixing `wkb_reader` to output `geometry_empty` in cases like that is a good approach. One thing to bare in mind is that envelope of `geometry_empty` is invalid (0,0,-1,1) and needs to be checked. 
*Note to myself - I wonder if we should re-visit how we handle invalid bounding boxes

talaj:
> One thing to bare in mind is that envelope of geometry_empty is invalid (0,0,-1,1) and needs to be checked. 

@artemp - Parsing empty polygons, linestrings, etc. as `geometry_empty` should help here, because `geometry_empty` should lead to no-op in all the static visitors.

I have added commit for that to this pull request. I'm thinking about test case now.

artemp:
@talaj  :+1:

talaj:
@artemp - Testcase added, but I have doubts about solution I proposed. According to solution I proposed also wkt parser should parse all the empty geometry types as `geometry_empty`. But that would break universality of these parsers. I think the behavior that empty polygon is parsed as `geometry::polygon` is correct. It's even inevitable when parsing multi-polygon where one polygon of the set is empty. In such case all the geometries must be of the same type. When processing such multi-polygon, you usually iterate through the set and then bang! Exception on empty polygon.

artemp:
@talaj - Yes, I see your points. Lets not rush with this change but think over this weekend.  Thanks for your work!

artemp:
@talaj - after sleeping over this issue - I think we should output genuine `empty` geometries from input plug-ins but ensure client code handles them gracefully.

talaj:
@artemp - Exactly. I think all algorithms like centroid should have it's own check for input data. There are two basic reasons.  Firstly, various algorithms can have various requirements on input data and it's not possible to have one point in the code which will check all these requirements. Secondly, even if it would be possible, it's bad design to couple centroid algorithm and postgis plugin for instance.

I have added commit here which checks for empty geometries before call to `boost::geometry::centroid()`. But the unit test I also added here is currently crashing on multi-linestring where one linestring component is empty. I'm now thinking what to do about it.

talaj:
I had to introduce `has_empty()` test because boost centroid algorithm crashes on multi-geometries where one component is empty. `has_empty` tests for empty components of multi-geometries. I have also added unit tests for centroid algorithm, `is_empty` and `has_empty`.

This pull request tries to fix https://github.com/mapnik/mapnik/issues/3051 particularly. I think there should be separate issue to revisit how empty geometries are handled in the code.

artemp:
@talaj - Just wondering if returning `false` for Multi(Line|Polygon) when only some parts are `empty` is right thing to do? Would it make more logical to calculate centroid for all non-empty geometries and skip empty ones?

talaj:
@artemp - Yes, calculate centroid for all non-empty geometries makes more sense, but returning `false` is better than seg-fault now.

The centroid algorithm would ideally ignore empty sub-geometries, but the logic is in Boost library. We can only make a copy (or some kind of geometry view) omitting empty geometries and pass it to Boost. Or to have own centroid implementation.

artemp:
> returning false is better than seg-fault now. 
:+1:

PR #3180, opened 2015-11-20
1st iteration of PostGIS tests

BergWerkGIS:


springmeyer:
Passing on travis now (mostly). The coverage run is timing out. Solution there is to move the coverage run to linux, but that is a bigger issue. So, merging this now. Great work on this @BergWerkGIS!

Issue #3169, opened 2015-11-11
centroid algorithm: handling empty sub-geometries


talaj:
Centroid algorithm does not calculate on multi-geometries with empty sub-geometries after https://github.com/mapnik/mapnik/pull/3164 because it is seg-faulting in Boost.

I'm going to fix the crash and think about consequent steps.

talaj:
The crash in Boost is fixed now by https://github.com/boostorg/geometry/pull/336. We will probably need to introduce some preprocessor condition for Boost version. Would be sufficient to just keep skipping problematic geometries for versions of Boost which are crashing?

What do you think @artemp?

artemp:
@talaj - great stuff fixing boost.geometry! Skipping sounds good to me, but wouldn't this change current behaviour and would result in some 'invalid' geometries not being rendered ? Still for consistency feels like we should just skip them :)

talaj:
@artemp - You are right, skipping will be best. I will do that.

talaj:
Closed by https://github.com/mapnik/mapnik/pull/3185.

PR #3185, opened 2015-11-24
centroid: enable algorithm on multi-geometries with empty sub-geometries

talaj:
Addresses https://github.com/mapnik/mapnik/issues/3169.

* Removes empty sub-geometries from multi-geometry if Boost version is <= 1.59.
* Uses try-catch for `boost::geometry::centroid_exception` instead of testing emptiness. The try-catch should be there anyway as exceptions coming from Boost should not break rendering whatsoever.


artemp:
@talaj - would be cool to get this into https://github.com/mapnik/mapnik/milestones/Mapnik%203.0.9
Would you like to merge?

talaj:
I have added note to changelog. Can be merged from my point of view.

PR #3197, opened 2015-12-06
Fix name reported by mapnik-render

tomhughes:


springmeyer:
Thanks for catching this @tomhughes

PR #3199, opened 2015-12-09
2.3.x: fix feature_style_process_style_impl

wsw0108:


wsw0108:
the problem that this PR fixed also existed on branch master.

wsw0108:
the problem that this PR fixed also existed on branch master.

springmeyer:
@wsw0108 - thank for for this pull request/fix. Before merging we need a testcase that demonstrates the problem you are fixing. Is that something you could provide? See https://github.com/mapnik/mapnik/blob/master/docs/contributing.md#testing for more details.

Issue #2101, opened 2013-12-03
SQLite plugin does not support !scale_denominator! token


SNiLD:
PostGIS plugin has this feature and I need it for Spatialite as well. I have made a patch to the SQLite plugin to support this.

I don't have rights to commit to the project directly, so here are the patches:

**plugins/input/sqlite/sqlite_datasource.cpp**:

``` cpp
@@ -37,10 +37,12 @@

 // boost
 #include <boost/algorithm/string.hpp>
 #include <boost/tokenizer.hpp>

+const double sqlite_datasource::FMAX = std::numeric_limits<double>::max();
+
 using mapnik::box2d;
 using mapnik::coord2d;
 using mapnik::query;
 using mapnik::featureset_ptr;
 using mapnik::layer_descriptor;
@@ -63,10 +65,11 @@ sqlite_datasource::sqlite_datasource(parameters const& params)
       geometry_field_(*params.get<std::string>("geometry_field", "")),
       index_table_(*params.get<std::string>("index_table", "")),
       key_field_(*params.get<std::string>("key_field", "")),
       row_offset_(*params.get<mapnik::value_integer>("row_offset", 0)),
       row_limit_(*params.get<mapnik::value_integer>("row_limit", 0)),
+      scale_denom_token_("!scale_denominator!"),
       intersects_token_("!intersects!"),
       desc_(*params.get<std::string>("type"), *params.get<std::string>("encoding", "utf-8")),
       format_(mapnik::wkbAuto)
 {
     /* TODO
@@ -354,15 +357,24 @@ sqlite_datasource::sqlite_datasource(parameters const& params)
 }

 std::string sqlite_datasource::populate_tokens(std::string const& sql) const
 {
     std::string populated_sql = sql;
+
+    if (boost::algorithm::icontains(sql, scale_denom_token_))
+    {
+        std::ostringstream ss;
+        ss << FMAX;
+        boost::algorithm::replace_all(populated_sql, scale_denom_token_, ss.str());
+    }
+
     if (boost::algorithm::ifind_first(populated_sql, intersects_token_))
     {
         // replace with dummy comparison that is true
         boost::algorithm::ireplace_first(populated_sql, intersects_token_, "1=1");
     }
+
     return populated_sql;
 }

 sqlite_datasource::~sqlite_datasource()
 {
@@ -503,10 +515,17 @@ featureset_ptr sqlite_datasource::features(query const& q) const
         }
         s << " FROM ";

         std::string query(table_);

[...]

PR #3230, opened 2016-01-13
add support for scale_denominator token in sqlite sources

JesseCrocker:
This code was originally presented as issue #2101 2 years ago but was never submitted as a pull request.


springmeyer:
Thanks for this contribution. This will need tests added. Since the tests for sqlite have not yet been ported from python into c++ core, that would be the first step towards getting this pull viable.

JesseCrocker:
Closing, because im probably never going to make time to port the sqlite tests all over.

PR #3232, opened 2016-01-13
Throttle parallelization of memory-intensive compilation tasks

lightmare:
There's this trick in [Makefile](https://github.com/mapnik/mapnik/blob/6e93b05a063387e3ec8a3493661ed8cd91d9c52d/Makefile#L39):

``` make
src/json/libmapnik-json.a:
    # we first build memory intensive files with -j1
    $(PYTHON) scons/scons.py -j1 --config=cache --implicit-cache --max-drift=1 \
        src/...

mapnik: src/json/libmapnik-json.a
    # then install the rest with -j$(JOBS)
    $(PYTHON) scons/scons.py -j$(JOBS) --config=cache --implicit-cache --max-drift=1
```

But that only works when `libmapnik-json.a` doesn't exist. If you don't delete the library from the previous build and do `make JOBS=2`, everything will be built with `-j2`. Which may be rough. For me `gcc src/*/process_markers_symbolizer.cpp` takes ~140s and ~4GB of memory, so whenever I change a header included by two of those, it's very likely they will meet, eat all 8GB of my RAM and end up in a swap-thrashing limbo.

If you do delete the library, or change the `-j1` target to a phony one, you'll be waiting for scons to start twice every time.

This PR adds a new option to scons: --hogs=N (defaults to 1)
It controls how many jobs (out of the total number given by --jobs) may run memory-intensive tasks.

Which tasks are considered memory-intensive is specified by tagging sources or headers with new scons environment method `MemoryHog`. Here I'm not entirely sure that was the correct way to do this, my knowledge of the scons system is quite shallow. Note: this modifies `SCons.Node` class, so you may need to `rm -f config.cache .sconsign.dblite` to get rid of old pickled data.

In the second commit I marked all grammar headers and heavy symbolizer sources as memory hogs, and replaced the above Makefile trick with an additional make variable `HOGS`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,10 +7,14 @@ ifeq ($(JOBS),)
 	JOBS:=1
 endif
 
+ifeq ($(HOGS),)
+	HOGS:=1
+endif
+
 all: mapnik
 
 install:
-	$(PYTHON) scons/scons.py -j$(JOBS) --config=cache --implicit-cache --max-drift=1 install
+	$(PYTHON) scons/scons.py -j$(JOBS) --hogs=$(HOGS) --config=cache --implicit-cache --max-drift=1 install
 
 release:
 	export MAPNIK_VERSION=$(shell ./utils/mapnik-config/mapnik-config --version) && \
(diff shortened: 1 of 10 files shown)


springmeyer:
Thank you for looking deeper into this. Totally agree the existing trick has pitfalls. I'm hesitant to change scons since my knowledge of it is also limited and it would make updating the bundled version more tedious. I'm been playing around with using `ninja` which has the ability to throttle builds based on system load `ninja -l 2` which seems to work decently (https://github.com/mapnik/mapnik-gyp/blob/master/build.sh#L95-L105).

The real problem here is of course the time and memory it takes to compile these files. My hope is that the spirit-x3/c++14 branch may help here but I've not looked closely yet at whether it does.

springmeyer:
> However, it has Pools, which look like a much better, finer-grained variant of what I did.

Ah, nice find! I'm interested/open to porting to ninja as the primary build tool. The questions in my mind are: cmake or gyp to create ninja files on the fly or generate them ourselves? what kind of lift would it be?

lightmare:
I can't answer that, as I have zero experience with ninja or gyp, and very little with cmake. Ninja manual also hints at [premake](http://premake.github.io/). I also don't know how difficult the transition from scons.py to something less "scriptable" would be.

lightmare:
Despite having come up with a better, more flexible, way of doing this some time after posting the PR, I've never implemented it. Closing this because it's stale and not good enough to bother with resolving conflicts.

PR #3235, opened 2016-01-14
Raster Type Memory Datasource

flippmoke:
Made it possible to change the type of a memory datasource by inspecting the features that are added to the memory datasource /cc @artemp


flippmoke:
@artemp we didn't put any tests around this and probably should

PR #3239, opened 2016-01-18
condense expressions tests

lightmare:
There are some inconveniences in TEST_CASE("expressions")

```c++
expr = mapnik::parse_expression("[foo]='bazaar'");
REQUIRE(evaluate(*feature, *expr) == true);
```
```
test/unit/core/expressions_test.cpp:80: FAILED:
  REQUIRE( evaluate(*feature, *expr) == true )
with expansion:
  false == true
```

- REQUIRE makes it stop on the first failure => use CHECK instead
- if `parse_expression` throws, you're out, no more checks will be done => CHECK_NOTHROW
- if the check fails, the output doesn't contain the failing expression string

I couldn't find a way to combine CHECK_NOTHROW and CHECK, so I added an extra header with a modified Catch macro that does a CHECK but also swallows exceptions (which still fail the test, of course).

The two lines from above thus become a single-line check. If parsing or evaluation throws, or gives wrong result, the check will fail, with the original expression string included in the output.
```c++
NOTHROW_CHECK(eval(" [foo]='bazaar' ") == true);
```
```
test/unit/core/expressions_test.cpp:77: FAILED:
  NOTHROW_CHECK( eval(" [foo]='bazaar' ") == true )
with expansion:
  false == true
```


springmeyer:
:+1: looks great. I've used the `INFO` macro to log things out when they fail (Catch.hpp automatically only prints the INFO if the following assert fails - https://github.com/philsquared/Catch/blob/master/docs/logging.md) but this looks even nicer for our usecase.

lightmare:
I'm tempted to propose the added macro directly to Catch, as I think it's very useful, only in need of a better name perhaps.

lightmare:
Posted https://github.com/philsquared/Catch/issues/582 last week, so far no response, but since then I found a way to do this with even less code (and more importantly, less copy-past from Catch internals):

```c++
#define NOTHROW_CHECK( expr )                               \
    try {                                                   \
        CHECK( expr );                                      \
    }                                                       \
    catch ( Catch::TestFailureException& ) {                \
        /* thrown by CHECK after it catches and reports */  \
        /* an exception from expr => swallow this       */  \
    }
```

So I'm going to use this, keeping it in a header so that other tests can include it, too.

PR #3271, opened 2016-01-28
Compile image_filter_grammar separately

springmeyer:
I did not do this originally (when porting many grammars to be split between `.hpp` and `_impl.hpp`) because `image_filter_grammar_impl.hpp` is only used in one place. The logic goes: if the grammar is only used in one cpp file then compiling it separately will only add to overall compile times without saving much in terms of peak memory (because including `image_filter_grammar_impl.hpp` inside `src/image_filter_types.cpp` does not lead to a compile any more intensive that compiling `image_filter_grammar_impl.hpp` alone).

However on appveyor we are consistently hitting:

```
C1060: compiler is out of heap space (compiling source file ..\..\src\image_filter_types.cpp)
```

I have a small hope that this change will require slightly less memory and might help appveyor squeak by. We'll see.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ src/json/libmapnik-json.a:
 		src/css_color_grammar.os \
 		src/expression_grammar.os \
 		src/transform_expression_grammar.os \
-		src/image_filter_types.os \
+		src/image_filter_grammar.os \
 		src/agg/process_markers_symbolizer.os \
 		src/agg/process_group_symbolizer.os \
 		src/grid/process_markers_symbolizer.os \
--- a/src/build.py
+++ b/src/build.py
@@ -154,6 +154,7 @@ source = Split(
     well_known_srs.cpp
     params.cpp
     image_filter_types.cpp
+    image_filter_grammar.cpp
     miniz_png.cpp
     color.cpp
     conversions.cpp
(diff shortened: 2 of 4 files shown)


lightmare:
Here's an except from two compiles I did about a month ago; don't know commits/branches precisely, I just have these files still lying around (process_markers_symbolizer excluded since those are the usual 4GB suspects; these are the biggest followers):
```
  22.85s  mem:2197MB [0 page faults]      test/unit/imaging/image_filter.cpp
  23.49s  mem:2093MB [0 page faults]      src/json/mapnik_topojson_grammar.cpp
  18.13s  mem:1912MB [0 page faults]      utils/mapnik-index/process_geojson_file.cpp
  20.65s  mem:1885MB [2 page faults]      src/expression-new.cpp
  20.66s  mem:1883MB [0 page faults]      src/expression_grammar.cpp

  25.81s  mem:2198MB [0 page faults]      test/unit/imaging/image_filter.cpp
  24.15s  mem:2094MB [0 page faults]      src/json/mapnik_topojson_grammar.cpp
  17.62s  mem:1911MB [0 page faults]      utils/mapnik-index/process_geojson_file.cpp
  21.82s  mem:1885MB [0 page faults]      src/expression-new.cpp
  20.66s  mem:1882MB [0 page faults]      src/expression_grammar.cpp

```

springmeyer:
great news: one (at least) of the windows builds passed: https://ci.appveyor.com/project/Mapbox/mapnik/build/1.0.532. First successful builds on appveyor is some time (maybe months).

springmeyer:
> these are the biggest followers

@lightmare thanks for catching those - they should not be so expensive. I've ticketed the obvious problem at #3272

Issue #3274, opened 2016-01-29
cleanup `noexcept` specifiers


lightmare:
Pulling latest changes from mapbox/variant uncoverered an overly optimistic `noexcept`:
```
include/mapnik/value.hpp:762:5: error: exception specification of explicitly defaulted move constructor does not match the calculated one

    value( value && other) noexcept = default;

    ^
```

> http://en.cppreference.com/w/cpp/language/noexcept_spec
> Inheriting constructors and the implicitly-declared default constructors, copy constructors, move constructors, destructors, copy-assignment operators, move-assignment operators are all noexcept(true) by default, unless they are required to call a function that is noexcept(false), in which case these functions are noexcept(false).

I was still unsure whether *defaulted* constructor qualifies as *implicitly-declared*. It sounds like it shouldn't -- you've just explicitly, aka *user-declared* the constructor, and the *default* supplies only the definition. But it just wouldn't make sense if the *default* hadn't also provided the correct `noexcept`. Whether a function throws an exception or not depends on the definition, after all.

So I consulted the C++11 draft.
##### 8.4.2 Explicitly-defaulted function [dcl.fct.def.default]
1 ... A function that is explicitly defaulted shall
...
— not have an exception-specification.

2 An explicitly-defaulted function may be declared constexpr only if it would have been implicitly declared
as constexpr. If it is explicitly defaulted on its first declaration,
— it shall be public,
— it shall not be explicit,
— it shall not be virtual,
— it is implicitly considered to have the same exception-specification as if it had been implicitly de-
clared (15.4), and
...

lightmare:
If I read that correctly, this will have the correct (inherited from variant) `noexcept`:
```c++
value( value && other) = default;
```

artemp:
@lightmare - incidentally I pushed an attempt to have a conditional `noexcept` https://github.com/mapnik/mapnik/commit/a50036d421435b9ea23c4ce57fdc72cacef24d46

But this crashed clang on OS X (https://travis-ci.org/mapnik/mapnik)  - works locally for me using stock Xcode 7.2 clang .

lightmare:
Just let the compiler generate move/copy for mapnik::value:
```c++
diff --git a/include/mapnik/value.hpp b/include/mapnik/value.hpp
index 4b7a08e..45e2e31 100644
--- a/include/mapnik/value.hpp
+++ b/include/mapnik/value.hpp
@@ -757,13 +757,6 @@ public:
[...]

PR #3275, opened 2016-01-29
Cleanup defaulted constructors and assignment operators, noexcept specifiers

lightmare:
Refs #3274 

- remove explicitly-defaulted functions that don't need to be -- they'll be defined by the compiler
- rectify `noexcept` specifiers on conversion-from-forwarding-reference constructors, and replace some `std::move` that should've been `std::forward`
- attempt to give `mapnik::value` perfectly forwarding conversion constructor and assignment :hammer:

### DIFF
--- a/include/mapnik/feature.hpp
+++ b/include/mapnik/feature.hpp
@@ -90,7 +90,7 @@ private:
 using context_type = context<std::map<std::string,std::size_t> >;
 using context_ptr = std::shared_ptr<context_type>;
 
-static const value default_feature_value;
+static const value default_feature_value{};
 
 class MAPNIK_DECL feature_impl : private util::noncopyable
 {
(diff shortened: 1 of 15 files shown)


springmeyer:
Compiler is segfaulting on osx. I think this is for the same reason we just started seeing compiler segfaults on osx over at mapbox/variant: https://travis-ci.org/mapbox/variant/jobs/105705328#L84.

While I don't know the cause I think a decent workaround is to use latest Xcode on travis. Which I've done in https://github.com/mapnik/mapnik/commit/b3170c397c132056b2e6f9cdb6161d7bf5b10d04. So, can you rebase and see if that fixes your branch builds on OS X?

lightmare:
And I thought it was something we did with `mapnik::value`. Too many moving parts :space_invader: I'll just push my remaining changes to `mapnik::value`, see how it goes, and then rebase.

springmeyer:
Messed up a few things in master in terms of the travis config, so you should wait to rebase until master is green on travis (hopefully in a few minutes).

springmeyer:
Okay, all clear: master is working now again.

Issue #3214, opened 2015-12-28
barewords in expression grammar


lightmare:
I'm trying to get in sync after being offnik for quite some time.

Blame on [expression_grammar_impl.hpp#L195](https://github.com/mapnik/mapnik/blame/master/include/mapnik/expression_grammar_impl.hpp#L195) gives only this huge commit e9ca9c8 with no detailed history. After some searching I finally found that the rule for parsing unquoted strings aka barewords was added in d01da6f. But I still can't figure out why.

I think barewords are a bad idea, unless restricted to only be allowed in unambiguous contexts.

```
  eval(e = 'e') => true
  eval(pi = 'pi') => false
```

That's pretty weird, and constants you add in the future will break filters that use their names as unquoted strings. In my opinion it's better to force users to properly quote their strings, than leaving them prone to such surprises.

I noticed that the spirit-x3 branch currently doesn't parse barewords. Should I focus on spirit-x3, and just ignore what's broken in expression grammar in master?

flippmoke:
Welcome back @lightmare ! Everyone is on Christmas break, but I would expect @artemp and @springmeyer back after the new year sometime. 

If you need help navigating anything since plenty has changed let me know!

PR #3287, opened 2016-02-06
restrict barewords (unquoted strings) in expression grammar

lightmare:
It's not barewords per se that break the logical `not` operator, it's [this line](https://github.com/mapnik/mapnik/commit/6a51c6e7f6fe784d93332fd48657b2dde77ce020#diff-1e2ad6c59a4d7152bb76fc8547a9a796L191) actually -- `not expr` is not a `primary_expr`, this line makes `not x op y` be parsed as `not (x op y)`, in other words, giving `not` the lowest possible precedence.

But as I wrote in #3214, barewords are not future-proof, their use should be restricted and it's easier to just remove them.

Btw, in current master, this check passes:
```c++
TRY_CHECK(eval(" not not = not not ") == true);
```
The expression is parsed as `not ('not' = not 'not')`. The first `not` is grabbed by the evil line I linked above, the second is tried by that line, but `= not not` fails to match `expr` so it backtracks and matches `ustring`. The third `not` is grabbed by the evil line again, and the last by `ustring`. The second alternative in the `not_expr` rule higher in the grammar is never used.

### DIFF
--- a/include/mapnik/expression_grammar_impl.hpp
+++ b/include/mapnik/expression_grammar_impl.hpp
@@ -104,7 +104,9 @@ expression_grammar<Iterator>::expression_grammar(std::string const& encoding)
     standard_wide::no_case_type no_case;
     using boost::phoenix::construct;
 
-    expr = logical_expr.alias();
+    expr = logical_expr [_val = _1]
+        | ustring [_val = unicode_(_1)]
+        ;
 
     logical_expr = not_expr [_val = _1]
         >>
@@ -167,10 +169,11 @@ expression_grammar<Iterator>::expression_grammar(std::string const& encoding)
             )
         ;
 
-    unary_function_expr = unary_func_type > lit('(') > expr > lit(')')
+    unary_function_expr = unary_func_type >> '(' > logical_expr > ')'
         ;
 
-    binary_function_expr = binary_func_type > lit('(') > expr > lit(',') > expr > lit(')')
+    binary_function_expr = binary_func_type >> '(' > logical_expr > ','
+                                                   > logical_expr > ')'
         ;
 
     unary_expr = primary_expr [_val = _1]
@@ -188,11 +191,9 @@ expression_grammar<Iterator>::expression_grammar(std::string const& encoding)
         | lit("[mapnik::geometry_type]")[_val = construct<mapnik::geometry_type_attribute>()]
         | attr [_val = construct<mapnik::attribute>( _1 ) ]
         | global_attr [_val = construct<mapnik::global_attribute>( _1 )]
-        | lit("not") >> expr [_val =  !_1]
         | unary_function_expr [_val = _1]
         | binary_function_expr [_val = _1]
-        | '(' >> expr [_val = _1 ] >> ')'
-        | ustring[_val = unicode_(_1)] // if we get here then try parsing as unquoted string
+        | '(' > logical_expr [_val = _1 ] > ')'
         ;
 
     unesc_char.add("\\a", '\a')("\\b", '\b')("\\f", '\f')("\\n", '\n')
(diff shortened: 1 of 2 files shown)


lightmare:
There's one failing unit test which I think might be a typo in the test: @flippmoke can you check the `dy="t"`here? https://github.com/mapnik/test-data/commit/3b1f481ac2ceeb780255cd4c76a42ae06197b9bc#diff-2f6d1bb7f324e0d6a35ca20c7023f5beR18

Then there are 5 visual tests that actually rely on parsing bareword `bar` as a string:
`bar` in [test/data-visual/styles/group-symbolizer-line-2.xml](https://github.com/mapnik/test-data-visual/blob/f5623b5a312c58cd9c1926008fcf114944c769a3/styles/group-symbolizer-line-2.xml)
`bar` in [test/data-visual/styles/group-symbolizer-line-avoid-edges.xml](https://github.com/mapnik/test-data-visual/blob/f5623b5a312c58cd9c1926008fcf114944c769a3/styles/group-symbolizer-line-avoid-edges.xml)
`bar` in [test/data-visual/styles/group-symbolizer-line-minimum-padding.xml](https://github.com/mapnik/test-data-visual/blob/f5623b5a312c58cd9c1926008fcf114944c769a3/styles/group-symbolizer-line-minimum-padding.xml)
`label` in [test/data-visual/styles/multipolygon-centroid-postgis.xml](https://github.com/mapnik/test-data-visual/blob/f5623b5a312c58cd9c1926008fcf114944c769a3/styles/multipolygon-centroid-postgis.xml)
`XX` in [test/data-visual/styles/formatting-expr-alignment.xml](https://github.com/mapnik/test-data-visual/blob/f5623b5a312c58cd9c1926008fcf114944c769a3/styles/formatting-expr-alignment.xml)

And 1 where I think `[label]` was intended instead:
`label` in [test/data-visual/styles/multipolygon-centroid-wkt.xml](https://github.com/mapnik/test-data-visual/blob/f5623b5a312c58cd9c1926008fcf114944c769a3/styles/multipolygon-centroid-wkt.xml)

lightmare:
So, if you want to really want to keep barewords in the grammar, I suggest restricting them to single-term expressions only, never inside complex expressions with operators, i.e.:
```c++
expr = logical_expr | ustring;
```
What do you think?

artemp:
>```c++
expr = logical_expr | ustring;
```
^  I like this variant

**Comment by lightmare** [comment]
@artemp ok I'll redo the pull. One more question, as my previous suggestion:
> single-atom expressions only, never inside complex expressions with operators

Isn't equivalent to this grammar:
```
expr = logical_expr | ustring;
```
Since `expr` is used recursively in:
 - `unary_function_expr`
 - `binary_function_expr`
 - `'(' >> expr >> ')'`

This grammar will actually accept:
`(foo) + (bar) == "foobar"` -- both `foo` and `bar` will match `ustring` in parenthesized `expr`
`max(foo, bar) == max("foo", "bar")`

That means parentheses will be alternative string-quotes. Is that okay, or should I go with *single-atom expressions only*?

**Comment by artemp** [comment]
@lightmare - Thanks, I think we should go for `single-atom` expressions as it's conceptually cleaner and will allow easily drop unquoted string support in the future.

**Comment by lightmare** [comment]
Just realized one more change might be needed.
```c++
unary_function_expr = unary_func_type > lit('(') > logical_expr > lit(')');
```
The expectation operator `>` before the opening parentheses will prevent using a function name as a bareword. Currently `log` matches unquoted string, but if the `log` function is added in the future, it will be an error. However, as there is no ambiguity, changing the first expectation to `>>` shouldn't hurt, and won't break single-word expressions when functions are added.

artemp:
@lightmare - looking good, going to port to https://github.com/mapnik/mapnik/tree/spirit-x3

PR #3289, opened 2016-02-09
bootstrap: relieve leaning toothpick syndrome

lightmare:
It's a minor change, but one can't be 100% sure it's correct with osx builds never finishing on travis.


flippmoke:
@lightmare I am checking right now on OSX locally will merge if it is correct

lightmare:
@flippmoke thanks

PR #3293, opened 2016-02-15
Apply rectangle clip before painting background. Fixes #2157

### LINKED ISSUE #2157: background-color "bleeds" across whole context when rendering PDF via python-cairo
If specifying a background-color in the XML for a map, mapnik.render(map, context) in 2.3.0-dev bleeds the background colour across the whole PDFSurface, not just the area of the map. In 2.2.0-pre the background colour remained contained within the extent that the map takes up.

e.g. <Map srs="&srs900913;" background-color="&blue50;">

surface = cairo.PDFSurface(file.name, 220, 220)
ctx = cairo.Context(surface)
ctx.translate(10, 10)
map = mapnik.Map(200, 200)
mapnik.load_map(map, styleFile)
mapnik.render(map, ctx)

iqqmuT:
Perhaps this is not the way to fix the bug, but it shows where the problem is.
Apparently `start_map_processing()` was not called, so background `context._paint()` was applied to whole context instead of map rectangle. The same bug still exists for other renderers.

Here is PDF output of modified python-mapnik test:

``` diff
diff --git a/test/python_tests/cairo_test.py b/test/python_tests/cairo_test.py
index c6c25a3..0751c48 100644
--- a/test/python_tests/cairo_test.py
+++ b/test/python_tests/cairo_test.py
@@ -118,7 +118,7 @@ if mapnik.has_pycairo():
         m = make_tmp_map()
         m.zoom_to_box(mapnik.Box2d(-180, -90, 180, 90))
         test_cairo_file = '/tmp/mapnik-cairo-context-test.pdf'
-        surface = cairo.PDFSurface(test_cairo_file, m.width, m.height)
+        surface = cairo.PDFSurface(test_cairo_file, m.width * 2, m.height * 2)
         expected_cairo_file = './images/pycairo/cairo-cairo-expected.pdf'
         context = cairo.Context(surface)
         mapnik.render(m, context)
```

fixed_bg

[#2157]

### DIFF
--- a/configure
+++ b/configure
@@ -22,11 +22,11 @@ fi
 # These do not normally need to be set except when
 # building against binary versions of dependencies like
 # done via bootstrap.sh
-VARS=()
+
 if [ -f mapnik-settings.env ]; then
     echo "Inheriting from mapnik-settings.env"
     . ./mapnik-settings.env
     VARS=( $(cat mapnik-settings.env) )
 fi
 
-$PYTHON scons/scons.py --implicit-deps-changed configure ${VARS[*]} "$@"
+$PYTHON scons/scons.py --implicit-deps-changed configure ${VARS[*]:-} "$@"


flippmoke:
@iqqmuT I don't know if this is the right solution either (haven't had time to look and I am sure @artemp will), however, I noticed that your pull request didn't include any tests. Could you please add a test fixture to the visual tests perhaps along with comments on the intent of the test. You might need to put in a pull request on https://github.com/mapnik/test-data-visual/ as well and then link it here so we can merge it at the same.

PR #3298, opened 2016-02-17
various benchmark framework tweaks

lightmare:
This changes a few minor things in benchmarks:
 - fixes (read: completely changes) output formatting
   - with gcc it looked like this:
```
// before
std::getline:                                           t:0 i:1000000083 milliseconds
csv_utils::getline_csv:                                           t:0 i:1000000073 milliseconds
std::getline:                                           t:30 i:33333343 milliseconds
csv_utils::getline_csv:                                           t:30 i:33333342 milliseconds

// after
std::getline                                  0 threads   10M iters     86 milliseconds 8553ps/iter
csv_utils::getline_csv                        0 threads   10M iters     84 milliseconds 8396ps/iter
std::getline                                 30 threads  333k iters     47 milliseconds
csv_utils::getline_csv                       30 threads  333k iters     43 milliseconds
```
 - adds option `--log-severity=debug|warn|error|none`
   - when mapnik was compiled with spammy default log severity, benchmarks were basically measuring console throughput :dash: and no way to turn it off
 - adds option `--min-duration=SECONDS`
   - the test loop will run repeatedly until that amount of time elapses
   - may help getting smoother averages when the specified number of iterations finishes too fast
 - adds a chain & ball class for running tests in sequence
   - I modified some existing benchmarks that run multiple tests to show it off :tada:

### DIFF
--- a/benchmark/bench_framework.hpp
+++ b/benchmark/bench_framework.hpp
@@ -2,6 +2,7 @@
 #define __MAPNIK_BENCH_FRAMEWORK_HPP__
 
 // mapnik
+#include <mapnik/debug.hpp>
 #include <mapnik/params.hpp>
 #include <mapnik/value_types.hpp>
 #include <mapnik/safe_cast.hpp>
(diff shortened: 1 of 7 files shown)


benchmark/bench_framework.hpp
```diff
@@ -38,26 +39,79 @@ class test_case
...
+        }
+        else {
+            // missing second argument
+            std::clog << argv[0] << ": missing option '" << opt << "' value\n";
+            return -i; // negative means ill-formed option #i
         }
     }
+    return argc; // there were no non-option arguments
+}
+
```
springmeyer: For context I originally hand rolled this option parsing (instead of using boost::program_options) because blue sky I figured `bench_framework.hpp` could be pulled out of Mapnik and used elsewhere as a utility where boost was not around - like for variant.hpp. But this never happened and we can consider using `boost` here if this gets too painful.

springmeyer:
All the proposed changes sound excellent. One note is that @flippmoke has been interested in https://github.com/google/benchmark. I'm not clear yet on what value it adds that we are missing, but I sure has some good concepts.

benchmark/bench_framework.hpp
lightmare: boost::program_options is definitely better than this, you get long and short options, type constraints and --help right away. I never liked it, though, and once wrote a variation that used derived classes for option storage, validation and customization; and --help output more closely resembling that of GNU tools.

lightmare:
> https://github.com/google/benchmark

 - Catchy; also, there's a similar chain-call facitlity :8ball: 
 - it seems to have more reliable timing https://github.com/google/benchmark/blob/master/src/walltime.cc
 - more output formats
 - showed me the obvious that I should've added Xiters/s rather than Xs/iter :laughing: 
 - our little header is really lacking in the multi-threaded department; it starts measuring time after all threads have been created and are running (some might have finished already) -- should create them hanging on a condition, then start timing and notify_all

Issue #2923, opened 2015-06-24
remove pixel demultiply from agg_renderer


springmeyer:
Now that we track the premultiplied status on images I think we should consider:

 - [ ] stop calling `demultiply_alpha` at https://github.com/mapnik/mapnik/blob/master/src/agg/agg_renderer.cpp#L202
 - [ ] only demultiply during encoding when needed (like tiff supports premultipled pixes, but PNG does not, right?) 

The benefit of this would be:

 - If you want to take an image that goes through a renderer and then composite it to another image this would avoid needing to premultiply again.
 - Formats that support premultiplied pixels would avoid the speed cost of demultiply
 - Formats that do not support alpha would avoid the need for the speed cost (jpeg)
 - when needed the demultiplication might be able to be done during an existing loop over the pixels

/cc @artemp for comments

flippmoke:
:+1:

flippmoke:
I have been spending some time looking into this because frankly we have a big maze with the state of premultiplied in our code. In my opinion the following should always be followed:

* Only premultiply / demultiply right before the operations that need it. Upon exiting this algorithm the image should leave the image in the same state. 
* Changes in premultiplication / demultiplication state should be invisible and trivial for the user.

The state of premultiplication is already stored with the image so demultiply and premultiply operations already will not change the state! https://github.com/mapnik/mapnik/blob/master/src/image_util.cpp#L497

The reason for this line of thinking is that swapping state is something that is confusing and could be a waste of processing. Take for example `image_filter.hpp`:

* Different filter algorithms require different states of premultiplication
* Repeated algorithms of the same state could be called back to back. 
* An example https://github.com/mapnik/mapnik/blob/master/include/mapnik/image_filter.hpp#L395-L403 

So for example if I was to load a `jpg` render it with the image filter `blur` then with the filter `edge-detect` the state of the would be as follows:

 * Load Image  
 * Prepare to render (premultiply) 
 * image scaling (premultiplied required) and a new image (starting premultiplied is created)
 * filter blur (demultiply) 
 * after filter blur (premultiply) 
 * filter edge-detect (demultiply) 
 * after filter blur (premultiply) 
 * after rendering (demultiply). 
 * Save Image

So in order for this operations to be completed we did several extra steps of pre and multiplication which is just wasted CPU time and a reduction in algorithmic precision.

[...]

PR #3312, opened 2016-02-23
Lazily demultiply part one -- saving

lightmare:
This is the first bunch of commits towards #2923. These implement on-the-fly alpha demultiplication during image saving, so that when it's eventually removed from the renderer, savers will seamlessly consume premultiplied images and demultiply if needed.

Notes
 - TIFF supports premultiplied, so there's no conversion needed. The only change I made to tiff was that I enabled saving directly from image view

 - PNG with palette -- I didn't feel like demultiplying in several places (color reduction and encoding), so I just create a copy, demultiply and save that. This is not optimal, and might need a real on-the-fly implementation, or a warning "you should demultiply before saving indexed png for better performance"

 - WebP code was quite confusing, I tried to simplify it and also make sure that `RGB->YUV` conversion is consistent -- because when `pic.use_argb == 0`, `WebPPictureImportRGBA` does the conversion right away, but ignores `config.preprocessing`. Therefore I always set `pic.use_argb = 1`, for both lossy and lossless -- this is less efficient in the non-premultiplied lossy case, but it's the only way to ensure `config.preprocessing` setting is honored.

### DIFF
--- a/deps/agg/include/agg_pixfmt_rgba.h
+++ b/deps/agg/include/agg_pixfmt_rgba.h
@@ -56,7 +56,6 @@ template<class ColorT, class Order> struct multiplier_rgba
         }
     }
 
-
     //--------------------------------------------------------------------
     static AGG_INLINE void demultiply(value_type* p)
     {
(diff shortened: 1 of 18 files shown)


deps/agg/include/agg_pixfmt_rgba.h
```diff
@@ -78,6 +77,64 @@ template<class ColorT, class Order> struct multiplier_rgba
     }
 };
 
+//========================================================multiplier2_rgba
+template<class ColorT, class Order, class Order2 = Order> struct multiplier2_rgba
+{
```
springmeyer: I wonder if this addition is needed? Does it offer something that the code @artemp added in https://github.com/mapnik/mapnik/commit/2c931a238f8ff8fefeebfe92890b6e8f9bbb6b05 does not provide?

include/mapnik/image_view_any.hpp
```diff
@@ -44,12 +44,26 @@ using image_view_base = util::variant<image_view_null,
...
+    //      https://ci.appveyor.com/project/Mapbox/mapnik/build/1.0.670#L288
     template <typename T>
     image_view_any(T && data)
         noexcept(std::is_nothrow_constructible<image_view_base, T && >::value)
         : image_view_base(std::forward<T>(data)) {}
+#else
+    // inherit constructors from variant, in particular conversions
+    // from alternatives
+    using image_view_base::image_view_base;
+#endif
```
springmeyer: Is this #ifdef absolutely needed? I worry about this living in the code unnoticed.

test/unit/run.cpp
```diff
@@ -63,6 +96,7 @@ int main (int argc, char* const argv[])
 
     if (result == 0)
     {
+        auto tmp_base_dir = catch_temporary_path::create_dir("");
         result = session.run();
```
springmeyer: ^^ what is this temp directory used for @lightmare ?

deps/agg/include/agg_pixfmt_rgba.h
lightmare: The purpose was to avoid the copy needed with in-place `agg::multiplier_rgba`. Constructing color incurs a copy as well, and additionally color::demultiply is in .cpp so:
  1. can't be inlined
  2. will needlessly test premultiplied_ on every pixel

include/mapnik/image_view_any.hpp
lightmare: Well, we can keep the ugly forwarding constructor template. MSVC won't compile the #else branch.

test/unit/run.cpp
lightmare: image_io_test. I don't want a hundred files directly in /tmp, so they go in this subdirectory (/tmp/mapnik-test-unit)

PR #3315, opened 2016-02-25
Speed up CSV plugin / mapnik-index compile

springmeyer:
Moves all spirit usage to a cpp file.


lightmare:
`autodetect_csv_flavour` can probably be non-template taking just `std::istream &`

springmeyer:
@lightmare - nice spot  - you are correct. Committed that in 08d7bf2. Thank you.

lightmare:
It doesn't need to be included:

```diff
diff --git a/utils/mapnik-index/build.py b/utils/mapnik-index/build.py
index 5bb1770..651075a 100644
--- a/utils/mapnik-index/build.py
+++ b/utils/mapnik-index/build.py
@@ -32,6 +32,7 @@ source = Split(
     mapnik-index.cpp
     process_csv_file.cpp
     process_geojson_file.cpp
+    ../../plugins/input/csv/csv_utils.cpp
     """
     )
 
diff --git a/utils/mapnik-index/process_csv_file.cpp b/utils/mapnik-index/process_csv_file.cpp
index 8f4345b..5e3f825 100644
--- a/utils/mapnik-index/process_csv_file.cpp
+++ b/utils/mapnik-index/process_csv_file.cpp
@@ -23,7 +23,6 @@
 #include "process_csv_file.hpp"
 #include "../../plugins/input/csv/csv_getline.hpp"
 #include "../../plugins/input/csv/csv_utils.hpp"
-#include "../../plugins/input/csv/csv_utils.cpp"
 #include <mapnik/datasource.hpp>
 #include <mapnik/geometry_envelope.hpp>
 #include <mapnik/util/utf_conv_win.hpp>
```

lightmare:
... but that won't work with
`mapnik.gyp:      "sources": [ '<!@(python glob-files.py "../utils/mapnik-index/*.cpp")' ],`

how about
```sh
ln -srt utils/mapnik-index/ plugins/input/csv/csv_utils.cpp
```

BergWerkGIS:
there is no `ln` on Windows.

lightmare:
> there is no ln on Windows.

Actually there is, but I should've been more specific. Git can store symlinks, and I think it should be able to handle them even on filesystems that don't support them.

lightmare:
`utils/mapnik-index/process_csv_file.cpp` looks largely copy-pasted from `plugins/input/csv/csv_datasource.cpp`
Here's something scary:
```diff
-    if (!manual_headers_.empty())
+
+    ::detail::geometry_column_locator locator;
+    std::vector<std::string> headers;
+    std::clog << "Parsing CSV using SEPARATOR=" << separator << " QUOTE=" << quote << std::endl;
+    if (!manual_headers.empty())
     {
         std::size_t index = 0;
-        auto headers = csv_utils::parse_line(manual_headers_, separator_, quote_);
+        headers = csv_utils::parse_line(manual_headers, separator, quote);
         for (auto const& header : headers)
         {
-            detail::locate_geometry_column(header, index++, locator_);
-            headers_.push_back(header);
+            ::detail::locate_geometry_column(header, index++, locator);
+            headers.push_back(header);
         }
     }
     else // parse first line as headers
     {
```
csv-plugin has `headers_` and `headers`, mapnik-index modifies `headers` during traversal :bomb: (vector modifiers invalidate all iterators).

I'd try merging the two `parse_csv` variants into a template method (the design pattern, not C++ member function template), with hook points where csv-plugin and mapnik-index diverge.

springmeyer:
> mapnik.gyp: "sources": [ '<!@(python glob-files.py "../utils/mapnik-index/*.cpp")' ],

@lightmare how about just (untested)?

```
mapnik.gyp: "sources": [
  '<!@(python glob-files.py "../utils/mapnik-index/*.cpp")',
  '../../plugins/input/csv/csv_utils.cpp'
],
```

springmeyer:
I'll finish this work in v3.0.11

lightmare:
I compared the functions in csv-plugin and mapnik-index side by side and it turns out the only real difference is that mapnik-index doesn't have [this loop](https://github.com/mapnik/mapnik/blob/ca0c0e58886abf79c486ed92ed395a1f5d2c7984/plugins/input/csv/csv_datasource.cpp#L363). Other differences are suble, basically flags that plugin branches on, and mapnik-index assumes false. I'll make a separate PR once I clean it up; hope the tests are solid ;)

lightmare:
Oops sorry, pushed to wrong remote :-1: Will rewind.

springmeyer:
This work is done and I'm merging now. Not seeing a notable improvement (still) to overall compile times like I hoped, but definitely helps with CSV plugin and with re-compile times when developing the CSV plugin, mapnik-index, and the unit tests that include json parsing. So, since there is no negative impact and some positive benefit I think this is worthwhile.

PR #3318, opened 2016-02-26
Value cpp [Experimental] 

artemp:
I'm undecided about this one - will reassess  again post 3.0.10


lightmare:
Once the `null != ''` hack is eradicated, `detail::equals` can be replaced with std equivalents (as template template parameters to detail::comparison).

artemp:
@lightmare - we should definitely deprecate `null != ''` as for using `std::` predicates - I tried but resorted to adding  custom ones in order to deduct return types and simplify/reduce casting blues e.g

c++ 
``` int == int ``` result is `bool`

ICU
```icu::UnicodeString == icu::UnicodeString``` result is ```integer```

Having said that there might be a more elegant solution for sure. 

```c++
template <typename T>
    static auto apply(T const& lhs, T const& rhs)
        -> decltype(lhs == rhs) // re
    { ... }
```

lightmare:
> icu::UnicodeString == icu::UnicodeString result is integer

Yea I noticed that, and I think the proper fix is (can't find the stash I had this (tested), so improvising now):
```diff
diff --git a/include/mapnik/value.hpp b/include/mapnik/value.hpp
index c6a95d3..5ffe8c9 100644
--- a/include/mapnik/value.hpp
+++ b/include/mapnik/value.hpp
@@ -164,7 +164,7 @@ struct comparison
     bool operator() (value_unicode_string const& lhs,
                      value_unicode_string const& rhs) const
     {
-        return Op::apply(lhs, rhs) ? true : false;
+        return Op::apply(lhs.compare(rhs), 0);
     }
 
     //////////////////////////////////////////////////////////////////////////
```

Issue #3317, opened 2016-02-26
Use geometry instead of box3d


pnorman:
Mapnik uses `SetSRID('BOX3D(-39135.75848201022 6222585.598639628,665307.8941941741 6927029.251315812)'::box3d,900913)` or similar for the bounding box substitution. After talking with people working on box3d at the OSGeo code sprint, this usage is surprising.

It would be better to use a geometry, as all usage will end up casting to a geometry anyways. ST_MakeEnvelope was suggested.

Additionally, only single precision is needed.

pnorman:
cc @dbaston

pnorman:
see also postgis/postgis#92

artemp:
@pnorman - good call, see https://github.com/mapnik/mapnik/pull/3319. 
As for single precision, sounds reasonable and we should resolve this - it might require updating tests. 

/cc @springmeyer

springmeyer:
Let's be very careful before changing this. I want to see full test coverage and a benchmark to validate the change is warranted. Also, this has been used historically since it was faster than constructing a geometry: https://trac.osgeo.org/mapserver/ticket/4023

pnorman:
> Let's be very careful before changing this. I want to see full test coverage and a benchmark to validate the change is warranted. Also, this has been used historically since it was faster than constructing a geometry: https://trac.osgeo.org/mapserver/ticket/4023

I'm not able to detect a statistically significant difference in speed with PostGIS 2.2.1, but this is likely to change when box3d turns into a polyhedral surface in the future. Also, a box3d with no third coordinates may have unexpected results.

pnorman:
see also #3551

PR #3319, opened 2016-02-26
use `ST_MakeEnvelope` (ref #3317)

artemp:
@springmeyer @pnorman - this can go into ~~3.0.10~~ 3.0.11  I reckon


pnorman:
As is it'll cause problems with PostGIS 1.5 and unknown SRIDs, as the SRID-less version was added in 2.0.0. It might be better to unconditionally emit the SRID, and if it is 0 or negative, the resulting geometry will have an unknown SRID.

```
select ST_asewkt(st_makeenvelope(1,2,3,4,0));
           st_asewkt            
--------------------------------
 POLYGON((1 2,1 4,3 4,3 2,1 2))
select ST_asewkt(st_makeenvelope(1,2,3,4,-1));
NOTICE:  SRID value -1 converted to the officially unknown SRID value 0
           st_asewkt            
--------------------------------
 POLYGON((1 2,1 4,3 4,3 2,1 2))
ST_asewkt(st_makeenvelope(1,2,3,4,1));
               st_asewkt               
---------------------------------------
 SRID=1;POLYGON((1 2,1 4,3 4,3 2,1 2))
```

artemp:
> unconditionally emit the SRID

updated in 63df9ee

Do we still need : https://github.com/mapnik/mapnik/blob/master/plugins/input/postgis/postgis_datasource.cpp#L340-L345 
?
/cc @pnorman @springmeyer ^

pnorman:
A srid of -1 and 0 should be the same to PostGIS (2.2 at least), but it's possible somewhere else in Mapnik there's a check for SRID which is expecting -1.

springmeyer:
taking off milestone as this needs more testing time.

pnorman:
conflicts with #3551

pnorman:
What's the status of this?

artemp:
Unless I'm missing something, this has been resolved by following implementation in both `postgis.input` and `pgraster.input`: 

```c++
std::string postgis_datasource::sql_bbox(box2d<double> const& env) const
{
    std::ostringstream b;
    b << "ST_MakeEnvelope(";
    b << std::setprecision(16);
    b << env.minx() << "," << env.miny() << ",";
    b << env.maxx() << "," << env.maxy() << ",";
    b << srid_ << ")";
    return b.str();
}
```

@pnorman - please, reopen and provide more context if you think this needs more tweaking.

lightmare:
@artemp That's the code in this PR/branch, but it hasn't been merged in master or v3.0.x.

I've wanted to apply some DRY cleaning to postgis/pgraster plugins for quite some time, and this function is a good candidate for moving to a shared header. But the plugins diverged, pgraster uses box2d because there was an issue with box3d turning into PolyhedralSurface.

One has to wonder why we're using box3d. I couldn't find any explanation via git log. So I went looking into PostGIS sources, and found that its [struct BOX3D](https://github.com/postgis/postgis/blob/2.5.2/liblwgeom/liblwgeom.h.in#L278-L284) is quite neat, in contrast to [struct GBOX](https://github.com/postgis/postgis/blob/2.5.2/liblwgeom/liblwgeom.h.in#L286-L303) which, while being used to store box2d, actually holds 4D coordinates and no `srid`. But that's essentially irrelevant, because there's no way to set that `box3d.srid` member from SQL -- `ST_SetSRID` accepts and returns a `geometry`, it never sees a box3d.

Even when we don't emit `ST_SetSRID`, PostGIS [OPERATOR(&&)](https://postgis.net/docs/geometry_overlaps.html) tests 2D bounding box intersection, it never uses more dimensions. And finally, this operator has overloads for `geometry`, `geography`, `box2df` and `gidx`. The last two are internal types used in indexes, they cannot be constructed in SQL. When `box2d` is used with operator `&&`, it is not used directly as is; it is cast to `geometry(POLYGON)`, and the bounding box used for the test is calculated from the polygon's vertices.

`ST_MakeEnvelope` looks like a clear winner to me. It allows specifying SRID without an extra `ST_SetSRID`, and avoids an intermediate box2d/box3d value that would get cast to geometry anyway.

Besides, the cost of construction of whatever we use for the bbox intersection test is largely irrelevant. `ST_MakeEnvelope` as well as `ST_SetSRID` and box2d/box3d casts are `IMMUTABLE` (SQL-speak for `constexpr`). What matters is whether the query plan utilizes the right index. While searching for clues on this issue I found [this unsettling report](https://gis.stackexchange.com/questions/253974/why-is-st-intersects-faster-than/253987) of `ST_Intersects(geom, ST_MakeEnvelope(...))` running a better/faster plan than `geom && ST_MakeEnvelope(...)`, which looks like it shouldn't ever be possible since `ST_Intersects` actually checks `&&` first and then does some more work. I don't understand why, but apparently the query planner may choose to ignore spatial index. However, when that happens, I don't think it can be blamed on some `IMMUTABLE` expression on the right-hand side of `&&`. Plus, the mentioned stackoverflow answer suggests a very nice solution: `WITH ... SELECT`, which is completely out of the scope of this PR, but good to keep in mind for when somebody encounters a bad query plan. Maybe we could switch to [WITH ... SELECT](https://www.postgresql.org/docs/current/queries-with.html) syntax eventually. In addition to helping the query planner, that would also make `!bbox!` token obsolete... hmm, now that I think of it, we could spill all the token values there, including `@vars`, making all our token substitutions obsolete.

Algunenano:
A quick note (I'll try to have a deeper look in the near future): the WITH ... syntax won't necessarily work in PG12 as it might be inlined but I'd need to find a modern example where intersects uses index and && doesn't to see what can be done in Postgis to address that.

PR #3321, opened 2016-02-26
rsync ccache for travis osx builds

lightmare:
First a few notes about `.travis.yml` changes
 - in `before_install` only `deps/mapbox/variant` submodule is updated
 - in `script` after `make`, test data is downloaded (remaining submodules updated) and `make test` run only if previous `make` succeeded *and less than 40 minutes elapsed*
 - coveralls and benchmarks are only run if everything succeeded and we're below 45 minutes

This effectively means that a build that would've successfully finished in 47 minutes will now probably skip the last two steps and thus fail. On the other hand, a build that would've been killed while in the middle of running tests or benchmarks will skip those and thus have a chance to upload cache -- which should help subsequent builds finish sooner (including re-running this one).

Required repository settings on travis:
 - secret environment variable `RCACHE_PASSWORD`
 - public environment variable `RCACHE_MODULE` or `RCACHE_REMOTE`
   - `RCACHE_MODULE` should be a double-colon path to rsync daemon
   - if you don't specify `RCACHE_MODULE`, it will be composed as `RCACHE_REMOTE:user`, with user obtained from `TRAVIS_REPO_SLUG`

I created a module for user "mapnik" on my cache store, will send details to @springmeyer. If you want to bake your own, it needs to be set up such that:
 - the given module is readable by user "travis-ci" using password "ic+sivart"
 - cache can be uploaded as the user obtained from `TRAVIS_REPO_SLUG` using `RCACHE_PASSWORD`

Example `/etc/rsyncd.conf` (rsync 3.1 or newer required for colon-options in `auth users`)
```conf
secrets file = /www/travis-cache/rsyncd.secrets
#hosts allow = 208.78.110.199
#hosts deny =
max connections = 10

use chroot = yes
read only = yes
list = yes
uid = nobody
gid = nogroup

refuse options = checksum partial partial-dir

[travis-cache:lightmare]
        comment = cache for builds on travis
        path = /www/travis-cache/lightmare
        log file = /www/travis-cache/lightmare.log
        auth users = lightmare:rw *:ro

[travis-cache:mapnik]
        comment = cache for builds on travis
        path = /www/travis-cache/mapnik
        log file = /www/travis-cache/mapnik.log
        auth users = mapnik:rw *:ro
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,14 +45,15 @@ matrix:
       compiler: clang
       # https://docs.travis-ci.com/user/languages/objective-c/#Supported-OS-X-iOS-SDK-versions
       osx_image: xcode7.2 # upgrades clang from 6 -> 7
-      env: JOBS=8 MASON_PUBLISH=true HEAVY_JOBS=3
+      env: MASON_PUBLISH=true CXX="ccache clang++" JOBS=8 HEAVY_JOBS=3
     - os: osx
       compiler: clang
       osx_image: xcode7.2 # upgrades clang from 6 -> 7
-      env: JOBS=8 COVERAGE=true HEAVY_JOBS=3
+      env: COVERAGE=true CXX="ccache clang++" JOBS=8 HEAVY_JOBS=3
 
 before_install:
  - source scripts/travis-common.sh
+ - source scripts/travis-rcache.sh
  - export PYTHONUSERBASE=$(pwd)/mason_packages/.link
  - export PATH=${PYTHONUSERBASE}/bin:${PATH}
  - export COVERAGE=${COVERAGE:-false}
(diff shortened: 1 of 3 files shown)


lightmare:
For some reason it doesn't work with COVERAGE=true -- ccache generates same-named files each time, but doesn't use them for subsequent runs.

springmeyer:
> For some reason it doesn't work with COVERAGE=true -- ccache generates same-named files each time, but doesn't use them for subsequent runs.

Unfortunately that is a blocker since 1) coverage runs on os x only (since it OOMs on linux due to requiring too much memory - at least it did ~2 weeks ago), and 2) a major motivation to trying to speed up OS X builds it to get coverage working.

springmeyer:
> Unfortunately that is a blocker

@lightmare - perhaps not a blocker if we can move the `--coverage` builds to linux --> #3344

lightmare:
@springmeyer - whether --coverage works with this or not is irrelevant; if it works for *some* build jobs, it's an improvement. I noted that just to make it clear I couldn't get it to work with *all* builds ;)

springmeyer:
>  couldn't get it to work with all builds

Ah, okay.

springmeyer:
Travis finally enabled caching for OS X machines (https://blog.travis-ci.com/2016-05-03-caches-are-coming-to-everyone), so I'm going to close this as it is working now after 609e5ba7afccbf53fe21e8a4dbc818e7468474ca and 83768dc3dddab039ac90efc42e32d85ad3055710 (build now took 9 minutes rather than 39). Next action is to restore the running of the C++ unit tests and hope it still is under the 48 min cut-off, which is done in e68de54

springmeyer:
Thanks for all the work on this @lightmare - hopefully depending on the travis solution is decently robust. If not we can revisit `rsync` - sound good?

Issue #2375, opened 2014-08-21
2.3.x PGRaster Branch showing tile boundries


adityanarsapurkar:
When Using ModTile and Renderd with OSM and Mapnik
The Plugin 2.3.x for PGRaster shows small boundries with tile size.
I am using layer as follows,

<Layer name="MS-OKL" status="on" srs="+proj=utm +zone=14 +datum=WGS84 +units=m +no_defs">
    <StyleName>raster</StyleName>
    <Datasource>
      <Parameter name="type">pgraster</Parameter>
      <Parameter name="table">okl_test</Parameter>
      <Parameter name="raster_field">rast</Parameter>
      <Parameter name="dbname">gis</Parameter>
      
      <Parameter name="estimate_extent">false</Parameter>
      
      
      <Parameter name="extent">-20037508,-19929239,20037508,19929239</Parameter>
    </Datasource>
</Layer>

And I have dumped the data into Postgres with raster2pgsql as follows,

raster2pgsql -s 900913 -I -C -M *.tif -t 5000x5000 public.ms_test > ms.sql

The default size of file is 5000x5000 only.

The detailed explanation of the issue with image can be found here (http://gis.stackexchange.com/questions/110832/mapnik-pgraster-plugin-creating-8x8-tiles-with-visible-boundaries).

pramsey:
I saw this effect doing a reprojection in MapNik (source UTM, destination web mercator). Each pgraster tile had a little seam. Fully zoomed out they disappeared, but zoomed close enough, they were visible.

springmeyer:
@pramsey - can you share a screenshot? There are a few things - as you can imagine - that could cause  this. Would be great to see some details.

pramsey:
Will do as soon as I get my mapnik running again... getting 3.x to build (actually getting the python bindings working) is turning into a quest

strk:
I'm just experiencing the tile boundaries myself, for the record.
I'll be looking at this soon. Do you still need a screenshot @springmeyer ?

NOTE: I see tiles only when the postgis raster table is itself tiled, but the postgis side tiles are ok because there's no more visible tiles by ST_Union()ing them back into a single image.

strk:
Here's a screenshot of a 128x128 tiled raster fetch (no overviews):

![screenshot from 2016-02-26 15 20 08](https://cloud.githubusercontent.com/assets/289678/13354629/7fdd4e3a-dc9c-11e5-9164-d97229deab1c.png)

[...]

PR #3325, opened 2016-02-26
Add "preunion_rasters" parameter for pgraster plugin

strk:
The parameter ("pre-union rasters") changes the PostGIS query to always return a single raster.
It is effective as a workaround for bug #2375 (visible source tile boundaries).

It would make much more sense to fix the bug than to provide a workaround, to be honest, anyway the parameter doesn't do any harm (defaults to false).

### DIFF
--- a/plugins/input/pgraster/README
+++ b/plugins/input/pgraster/README
@@ -14,6 +14,8 @@ with the following differences:
 
  - "clip_rasters" boolean introduced, defaults to false
 
+ - "preunion_rasters" boolean introduced, defaults to false
+
  - "band" introduced, with same semantic of the GDAL driver
 
 Credits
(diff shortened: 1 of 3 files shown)


strk:
I think this PR lacks:

 [ ] documentation (is there any place otehr than the REAMDE in plugin dir?)
 [ ] tests (don't remember if there were tests for pgraster input)

strk:
It looks like using ST_Resample gives better results by allowing specification of a grid to snap rasters to: https://github.com/mapnik/mapnik/pull/3348

strk:
As ST_Resample doesn't always help, I'm back to pre-union as a valid alternative (will be checking it more carefully next)

strk:
NOTE: applying this with current ST_Resize often results in an exception being thrown due to alignement being lost, so it's better to only apply this after ST_Resample is (even if neither of these two really fix the visible tile boundary, which is caused by something else).

See #3348

PR #3326, opened 2016-02-26
merge mapnik-index::process_csv_file and csv_datasource::parse_csv

lightmare:
Based upon #3315, melds two much-too-similar functions into one.

- the function in plugin was already configurable via flags, and only
  contained two un-conditioned blocks that process_csv_file didn't have

- so I extracted the common parts into a separate function (in a class
  holding the flags and state), process_csv_file calls it with default
  flags, plugin sets them from params

- removed namespace ::detail, moving stuff that was used outside
  csv_utils to ::csv_utils, and the rest to ::csv_utils::detail

For reference, diffs between each of the original functions and the new one:
[process_csv_file => csv_file_parser::parse_csv](https://gist.github.com/lightmare/f880caae8df0b466d16e#file-process_csv_file-diff)
[csv_datasource::parse_csv => csv_file_parser::parse_csv](https://gist.github.com/lightmare/948ac63dd2cda62a3114#file-csv_datasource-diff)

For 3-column vimdiff:
```bash
#! /bin/bash

commit=c21778f

parse_csv_from_mapdex () {
    git cat-file -p $commit^:utils/mapnik-index/process_csv_file.cpp |
    sed -ne '/process_csv_file(T/,/^}/p'
}

parse_csv_from_plugin () {
    git cat-file -p $commit^:plugins/input/csv/csv_datasource.cpp |
    sed -ne '/::parse_csv/,/^}/p'
}

parse_csv_from_worktree () {
    git cat-file -p $commit:plugins/input/csv/csv_utils.cpp |
    sed -ne '/::parse_csv/,/^}/p' 
}

vimdiff <(parse_csv_from_plugin) \
        <(parse_csv_from_worktree) \
        <(parse_csv_from_mapdex)
```

### DIFF
--- a/plugins/input/csv/csv_datasource.cpp
+++ b/plugins/input/csv/csv_datasource.cpp
@@ -69,21 +69,13 @@ DATASOURCE_PLUGIN(csv_datasource)
 csv_datasource::csv_datasource(parameters const& params)
 : datasource(params),
     desc_(csv_datasource::name(), *params.get<std::string>("encoding", "utf-8")),
-    extent_(),
-    filename_(),
-    row_limit_(*params.get<mapnik::value_integer>("row_limit", 0)),
-    inline_string_(),
-    separator_(0),
-    quote_(0),
-    headers_(),
-    manual_headers_(mapnik::util::trim_copy(*params.get<std::string>("headers", ""))),
-    strict_(*params.get<mapnik::boolean_type>("strict", false)),
     ctx_(std::make_shared<mapnik::context_type>()),
-    extent_initialized_(false),
-    tree_(nullptr),
-    locator_(),
-    has_disk_index_(false)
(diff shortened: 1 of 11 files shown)


springmeyer:
great refactoring work @lightmare - thank you. On monday @artemp is going to look into our CSV tests to make sure they are working right. So let's wait till he's confirmed that before merging this.

lightmare:
Absolutely. The function in mapnik-index was doing weird things with the headers vector; I'm pretty sure it was buggy, but not so sure whether it was supposed to work the same as the one in plugin. And I kinda hoped tests would say something if I got it wrong ;)

springmeyer:
I'm going to pick back up #3315 to try to help compile times. So, merging this into #3315 and will work from there. Thanks @lightmare!

PR #3337, opened 2016-03-01
speed up util::normalize_angle for stupidly-large values

lightmare:
It's pretty unlikely that real code calls this with stupidly large angles (like 100 radians), but I think these should be handled gracefully, too.

In my benchmarks the new function is
- slightly faster on positive angles (can explain this...)
- slightly slower on small negative angles > -15 (... but not this :confused: )
- starts using std::remainder above 32 pi (~100 radians), where it's slower
- no contest on angles of magnitude > 130

The point at which `std::remainder` is faster than subtracting `2 * pi` in a loop is of course highly architecture dependent. I haven't done any real investigation -- just ran a simple benchmark, `std::remainder` started winning around 110 radians on my laptop. So I picked 100 as the point where the extra precision might be worth it -- despite the fact that my function is still slower below ~130, because it does a few subtraction loops that are thrown away if the angle is too large.


include/mapnik/util/math.hpp
```diff
@@ -25,8 +25,12 @@
 
 #include <mapnik/config.hpp>
 
+#include <cmath>
+
 namespace mapnik { namespace util {
 
+constexpr double pi = 3.14159265358979323846;
+
```
springmeyer: M_PI can be used from `mapnik/global.hpp`

src/math.cpp
```diff
@@ -33,15 +29,42 @@ namespace util {
 
     double normalize_angle(double angle)
     {
-        while (angle >= M_PI)
+        double a = angle;
+
+        if (angle > pi)
```
springmeyer: Oh, @lightmare is there are reason you don't want to use `M_PI` from `global.hpp`?

lightmare: It's a bit of a story. When you include `<mapnik/util/math.hpp>`, which only contains one function that operates on angles, you're probably going to need `pi`, too. You don't want to include a header with a bunch of endian-conversion functions and macros. I noticed that some weeks ago and intended to move it to math.

Later I was digging in agg and saw `const double pi` there. And in C++11 we have `constexpr`. I've been using `M_PI` for years like everyone else, and was actually surprised when I learned it wasn't standardized. But `constexpr` is new, and new is always better :cocktail:

include/mapnik/util/math.hpp
artemp: `M_PI` is not part of c++ so yes, defining it in `include/mapnik/util/math.hpp` and removing MSVC related #if/def in `global.hpp` makes sense to me. One thing to be aware third party libs are using  `M_PI` from math.h, agg has its own #define ... We just have to be consistent

src/math.cpp
artemp: I like 

```c++ constexpr double pi = 3.14159265358979323846;
```

Issue #3327, opened 2016-02-27
Compiling render_markers_symbolizer.cpp needs unreasonable amounts of memory


tomhughes:
I was hopeful that #3265 would fix this, but now that 3.0.10 has been released it seems like things actually worse than in 3.0.9 :cry: 

With 3.0.9 I was able to build on all three of the Fedora primary architectures, but I needed to disable debug generation for 32 bit x86 and ARM or it would run out of memory.

With 3.0.10 not only do I need debug disabled still, but the ARM won't compile even with debug disabled without running out of memory compiling `src/renderer_common/render_markers_symbolizer.cpp`.

An example log of a failing ARM build is https://kojipkgs.fedoraproject.org//work/tasks/9176/13149176/build.log.

lightmare:
You're right, it has grown. Before the refactoring, there were 4 files that required 3.9+GB to compile with gcc 4.8. Afterwards only `src/renderer_common/render_markers_symbolizer.cpp` needed 4.2+GB -- that was good enough for me, as I could finally use JOBS=2; and I switch to building with clang by default, it only needs ~2GB to compile that.

I've just checked with gcc again and it went above 4.8GB, that's bad. It might be changes in variant, or boost/other libs I upgraded since then.

tomhughes:
The problem is that a 32 bit machine has an inherent limit of 4Gb for a single process regardless of how much memory there is in the machine, so a compile that needs that much memory is going to be impossible on any 32 bit machine...

artemp:
I just checked compile times for 3.0.9 and 3.0.10

##### v3.0.9

Total build time: 628.225099 seconds
Total SConscript file execution time: 0.697090 seconds
Total SCons execution time: 5.218378 seconds
Total command execution time: 622.309631 seconds

real	10m28.517s
user	73m18.336s
sys	1m36.722s

##### 3.0.10

Total build time: 559.878520 seconds
Total SConscript file execution time: 0.707139 seconds
Total SCons execution time: 4.951279 seconds
Total command execution time: 554.220102 seconds

real	9m20.166s
user	64m44.853s
sys	1m36.995s

They look good (as in better) in 3.0.10 but  `src/renderer_common/render_markers_symbolizer.cpp` memory requirement is breaking 32-bit builds now which is bad.

[...]

PR #3338, opened 2016-03-02
refactor markers_placement_finder

lightmare:
- refs #3327

Replace member variant of placement-type implementations with plain
union. The active implementation is chosen at construction time like
before.

Make placement-type implementation classes virtual to allow invoking
the active union member through a base class pointer.

clang 3.6
```
this
  43.61s  mem:  774MB [0 page faults]     src/renderer_common/render_markers_symbolizer.cpp
master
  64.61s  mem: 1325MB [0 page faults]     src/renderer_common/render_markers_symbolizer.cpp
```

gcc 4.8
```
this
  74.46s  mem: 1740MB [0 page faults]     src/renderer_common/render_markers_symbolizer.cpp
master
  85.71s  mem: 3459MB [6 page faults]     src/renderer_common/render_markers_symbolizer.cpp
```

### DIFF
--- a/include/mapnik/markers_placement.hpp
+++ b/include/mapnik/markers_placement.hpp
@@ -29,7 +29,6 @@
 #include <mapnik/markers_placements/vertext_first.hpp>
 #include <mapnik/markers_placements/vertext_last.hpp>
 #include <mapnik/symbolizer_enumerations.hpp>
-#include <mapnik/util/variant.hpp>
 
 namespace mapnik
 {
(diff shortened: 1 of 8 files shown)


tomhughes:
I've kicked off a scratch build on Fedora of 3.0.10 with this patch applied: http://koji.fedoraproject.org/koji/taskinfo?taskID=13194331

tomhughes:
Looks good - all three platforms built successfully on Fedora.

artemp:
Hmm.. but travis is still failing https://travis-ci.org/mapnik/mapnik/jobs/113005661.

@tomhughes - have you tried latest master on Fedora? I have a feeling it might pass as well, curious to find out. 
/cc @lightmare @springmeyer

artemp:
#### Just gathering some stats so we can make an intelligent decisions 
 
If travis is to believe (big ?) current master got much further with the compilation

/cc @lightmare @tomhughes @springmeyer 

#### refactor markers_placement_finder PR

```
Command execution time: test/unit/datasource/shapeindex.o: 45.599106 seconds
clang++ -o test/unit/geometry/geometry_envelope_test.o -c -std=c++11 -stdlib=libc++ -fvisibility=hidden -fvisibility-inlines-hidden -Wall -ftemplate-depth-300 -Wsign-compare -Wshadow -Wno-unsequenced -O3 -DMAPNIK_MEMORY_MAPPED_FILE -DMAPNIK_HAS_DLCFN -DBIGINT -DBOOST_REGEX_HAS_ICU -DHAVE_JPEG -DMAPNIK_USE_PROJ4 -DHAVE_PNG -DHAVE_WEBP -DHAVE_TIFF -DDARWIN -DMAPNIK_THREADSAFE -DBOOST_SPIRIT_NO_PREDEFINED_TERMINALS=1 -DBOOST_PHOENIX_NO_PREDEFINED_TERMINALS=1 -DBOOST_SPIRIT_USE_PHOENIX_V3=1 -DNDEBUG -DMAPNIK_MEMORY_MAPPED_FILE -DMAPNIK_HAS_DLCFN -DBIGINT -DBOOST_REGEX_HAS_ICU -DHAVE_JPEG -DMAPNIK_USE_PROJ4 -DHAVE_PNG -DHAVE_WEBP -DHAVE_TIFF -DDARWIN -DMAPNIK_THREADSAFE -DBOOST_SPIRIT_NO_PREDEFINED_TERMINALS=1 -DBOOST_PHOENIX_NO_PREDEFINED_TERMINALS=1 -DBOOST_SPIRIT_USE_PHOENIX_V3=1 -DNDEBUG -DHAVE_CAIRO -DGRID_RENDERER -DSVG_RENDERER -DHAVE_CAIRO -Itest -Imason_packages/.link/include/cairo -Imason_packages/.link/include/pixman-1 -Ideps -Ideps/agg/include -Iinclude -Imason_packages/.link/include -Imason_packages/.link/include/freetype2 -Imason_packages/.link/include/libpng16 test/unit/geometry/geometry_envelope_test.cpp
The job exceeded the maxmimum time limit for jobs, and has been terminated.

```

#### current master
```
* Running visual tests...
..............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
The job exceeded the maxmimum time limit for jobs, and has been terminated.
```

tomhughes:
Travis is only failing because the compile took too long right? I don't know what the timeout is in travis but  i386 and x86_64 each took about 1h20 on the Fedora build farm, and arm took just under 4 hours.

When you say "try master" do you mean with this patch? or without it?

artemp:
@tomhughes - just realised that `refactor markers_placement_finder` is already rebased on latest `master` so no need to try, sry for the noise:). 

If memory requirements are dropped this a very positive sign, thanks @lightmare. But I believe we can also squash compile times to fit travis imposed 50 min limit.

artemp:
Giving a travis another chance ;)

tomhughes:
One thing about the Fedora builds is that I deliberately don't pass -j when building so the build is not parallel at all - that was done to fix previous problems running out of memory

I think I'm right that mapnik now avoids building problem files in parallel though, so I can likely try adding that back in?

artemp:
### libmapnik.dylib size

#### master before https://github.com/mapnik/mapnik/pull/3338

```
ls -l /opt/mapnik/lib/libmapnik.dylib 
-rwxr-xr-x  1 root  wheel  15778820  2 Mar 13:32 /opt/mapnik/lib/libmapnik.dylib
```

#### after https://github.com/mapnik/mapnik/pull/3338

```
ls -l /opt/mapnik/lib/libmapnik.dylib 
-rwxr-xr-x  1 root  wheel  17258300  2 Mar 12:44 /opt/mapnik/lib/libmapnik.dylib
```

lightmare:
+1.5MB in lib size? That's another surprise :confused:  The difference the is even bigger on Linux: 17.2 => 18.9

lightmare:
@artemp - regarding travis build times, I think they vary wildly, perhaps depending on overall load. My favourite build is [5006](https://travis-ci.org/mapnik/mapnik/builds/105577713). There were a few green builds before it, but none after. And it coincides with the time we started fixing `noexcept` specifiers in variant... :bulb: 

As for this commit increasing the library size by 10%, I don't get it. I removed the whole variant move-construction and apply_visitor machinery, and replaced it with 6 virtual function tables, that's all. Where does it come from?

artemp:
@lightmare - it's hard to be 100% sure, but v-tables means compiler can't do certain optimisations/inlining. The binary size increase is probably a trickle-down side-effect of this change.   
This one of the reasons we don't usually use virtual inheritance in mapnik :) 
Lets keep this change in master as it helps with linux packaging which is important improvement but I'm keeping my eyes open as old approach results in tighter binaries - important too!

Issue #3339, opened 2016-03-02
visual test -j randomly running out of memory (linux)


lightmare:
#It happened to me a few times already, the test kept allocating memory until killed. So today I decided to stress it.

First a little function I use to kill it when it exceeds 2GB -- because if I let it run wild, it starts swapping and freezes the whole system.

```sh
watchrun () {
    while sleep 1; do
        ps --no-headers -o pid,rss "$@"
    done |
    while read pid rss; do
        test $rss -lt 200100 && continue
        echo "killing pid=$pid because rss=$rss"
        kill -QUIT $pid
    done
}

watchrun -C visual-test-run
```

In another terminal
```sh
# don't want to accidentally kill another process named 'run'
ln -s test/visual/run visual-test-run
# want core dump when killed
ulimit -c unlimited
while ! test -f core; do
    ./visual-test-run --jobs=20 --agg $(for x in {1..20}; do echo tiff-alpha-gdal; done)
done
```

edit: changed command to run 20 jobs with the same style; fails more reliably

lightmare:
Roughly half of the threads in core dump were in
```
Thread 6 (Thread 0x7f9ea57fa700 (LWP 24097)):
#0  __memcpy_sse2 () at ../sysdeps/x86_64/multiarch/../memcpy.S:443
#1  0x00007f9ebe9328ac in _int_realloc (av=0x7f9e7c000020, oldp=0x7f9e7c898640, 
    oldsize=609760, nb=609776) at malloc.c:4316
#2  0x00007f9ebe933fc9 in __GI___libc_realloc (oldmem=0x7f9e7c898650, bytes=609768)
    at malloc.c:3029
#3  0x00007f9eb9eb4ddc in ?? () from /usr/lib/libgdal.so.1
#4  0x00007f9eb9ebbbb4 in ?? () from /usr/lib/libgdal.so.1
#5  0x00007f9eb9f1f945 in ?? () from /usr/lib/libgdal.so.1
#6  0x00007f9eb9ebacb9 in ?? () from /usr/lib/libgdal.so.1
#7  0x00007f9eb9ec1803 in ?? () from /usr/lib/libgdal.so.1
#8  0x00007f9eb9ee3c78 in ?? () from /usr/lib/libgdal.so.1
#9  0x00007f9eb9ef0964 in ?? () from /usr/lib/libgdal.so.1
#10 0x00007f9eb9d34ead in ?? () from /usr/lib/libgdal.so.1
[...]

Issue #3093, opened 2015-09-30
Visual tests, Windows: Assertion in gdaldrivermanager.cpp


BergWerkGIS:
trying to run

```
test_visual_run.exe --jobs=%NUMBER_OF_PROCESSORS%
```

outputs:

```
ERROR 7: Assertion `FALSE' failed
in file `gdaldrivermanager.cpp', line 415
```

Does gdal require any special build flags to run the visual tests in parallel?

springmeyer:
On unix gdal needs to be built with `./configure --with-threads` (which is also default). Not sure how this works on windows... but here is the unix configure logic: https://github.com/OSGeo/gdal/blob/3cd6fe0e3adcf7b687e92dc3b823053eda9d9c43/gdal/configure.in#L504-L595

BergWerkGIS:
Does the `--jobs=` flag work on Linux?

First glance at gdal source suggests, that multithreading might be switched on by default: [`#define CPL_MULTIPROC_WIN32`](https://github.com/OSGeo/gdal/blob/e5f7f305dacf2d1a518c28bb2a61ebe47d2d51ea/gdal/port/cpl_multiproc.h#L44).

I've found a (rather old) reference that not all drivers support multithreading.
As [`'gdaldrivermanager.cpp', line 415`](https://github.com/OSGeo/gdal/blob/2.0/gdal/gcore/gdaldrivermanager.cpp#L403-L415) is at such place, an reason might be that one of the drivers (or just the Windows version thereof) is not multithread enabled?

@rouault any ideas about this?
We build gdal with [this script](https://github.com/mapbox/windows-builds/blob/master/scripts/build_gdal.bat).
The build target `test_visual_run.exe` is defined [here](https://github.com/mapnik/mapnik-gyp/blob/2c3b6a80631a40849ba0ab2761efad1c8c00f5a6/mapnik.gyp#L691-L718) and depends on the [mapnik target](https://github.com/mapnik/mapnik-gyp/blob/2c3b6a80631a40849ba0ab2761efad1c8c00f5a6/mapnik.gyp#L126-L232).

rouault:
Multi-threading is enabled by default in GDAL on Windows (and Unix as well since quite a few versions). Not sure why you get this error. Looking at the code, if one exclude multithreading problems, this could happen if there were 2 driver objects sharing the same name. Perhaps a plugin lying somewhere and the same driver also compiled in GDAL main lib ?

[...]

PR #3340, opened 2016-03-02
call GDALAllRegister once, from one thread only

lightmare:
- refs #3339, #3093

### DIFF
--- a/plugins/input/gdal/gdal_datasource.cpp
+++ b/plugins/input/gdal/gdal_datasource.cpp
@@ -45,6 +45,12 @@ using mapnik::layer_descriptor;
 using mapnik::datasource_exception;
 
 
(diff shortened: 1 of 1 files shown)


springmeyer:
@lightmare did you consider using http://en.cppreference.com/w/cpp/thread/call_once?

lightmare:
Yes, `call_once` is better, because it's new :cocktail:

> Initialization of function-local statics is guaranteed to occur only once even when called from multiple threads, and may be more efficient than the equivalent code using std::call_once. 

I dismissed it when I had the `static` guard on one line inside the constructor; `call_once` requires a separate global variable. And when I found that my one-liner generates an "unused variable" warning, I failed to assess it correctly and silenced it with a function. `call_once` would've been cleaner.

PR #3345, opened 2016-03-03
work around initialization data race in gdal/gtiff

### LINKED ISSUE #3093: Visual tests, Windows: Assertion in gdaldrivermanager.cpp

### LINKED ISSUE #3339: visual test -j randomly running out of memory (linux)

lightmare:
Hopefully fixes #3339 #3093 

It's not pretty, but seems to work. I've had

``` sh
./visual-test-run --jobs=20 --agg $(for x in {1..20}; do echo tiff-alpha-gdal; done)
```

running in a loop for an hour (4 processors); previously this quite reliably blew up in a few iterations.

@BergWerkGIS - can you please test on windows whether this actually fixes #3093 ?

### DIFF
--- a/plugins/input/gdal/gdal_datasource.cpp
+++ b/plugins/input/gdal/gdal_datasource.cpp
@@ -82,25 +82,7 @@ gdal_datasource::gdal_datasource(parameters const& params)
     shared_dataset_ = *params.get<mapnik::boolean_type>("shared", false);
     band_ = *params.get<mapnik::value_integer>("band", -1);
 
-#if GDAL_VERSION_NUM >= 1600
-    if (shared_dataset_)
-    {
-        auto ds = GDALOpenShared(dataset_name_.c_str(), GA_ReadOnly);
-        dataset_.reset(static_cast<GDALDataset*>(ds));
-    }
-    else
-#endif
-    {
-        auto ds = GDALOpen(dataset_name_.c_str(), GA_ReadOnly);
-        dataset_.reset(static_cast<GDALDataset*>(ds));
-    }
-
-    if (! dataset_)
-    {
-        throw datasource_exception(CPLGetLastErrorMsg());
-    }
-
-    MAPNIK_LOG_DEBUG(gdal) << "gdal_featureset: opened Dataset=" << dataset_.get();
+    open_dataset(dataset_name_.c_str(), shared_dataset_);
 
     nbands_ = dataset_->GetRasterCount();
     width_ = dataset_->GetRasterXSize();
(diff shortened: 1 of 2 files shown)


springmeyer:
#3093 is fixed by #3395. Even though windows is using still using GDAL 2.0.1 (https://github.com/mapbox/windows-builds/issues/84). I think the reason that #3395 fixes #3093 is because it fixes the ogr datasource as well as the gdal datasource.

artemp:
I'm getting random hangs (latest master) with 
```bash
./test/visual/run --jobs=10 --agg $(for x in {1..10}; do echo tiff-alpha-gdal; done)
```

/cc @lightmare @springmeyer

springmeyer:
@artemp I'm not seeing hangs in proj4. Even running 1000 times works for me:

```
./test/visual/run --jobs=10 --agg $(for x in {1..1000}; do echo tiff-alpha-gdal; done)
```

I tested with two builds of mapnik I have locally, both master mapnik, one against proj 4.8.0 and one against proj 4.9.2.

artemp:
@springmeyer - ok, I guess I have some stale binaries or something. I can replicate reliably, it hangs every other time.

springmeyer:
@artemp just checked and realized I'm building proj4 with `--without-mutex`. Been doing it this way for years. That might explain our difference?

artemp:
@springmeyer - yep, sounds like it! testing ...

artemp:
@springmeyer - confirming that rebuilding proj4 with `USE_THREAD = OFF` (CMake) solves this issue.

lightmare:
Closing without merging, reasons include:
 - the patch is ugly
 - serializes creation of all GDAL datasources until one is TIFF
 - underlying issue fixed in GDAL 2.0.2 released more than 2 years ago

It's better to recommend newer GDAL version. We already have a note in INSTALL.md, I added another to [wiki](https://github.com/mapnik/mapnik/wiki/GDAL#usage).

PR #3349, opened 2016-03-04
Raster resample division by zero

talaj:
Addresses https://github.com/mapnik/node-mapnik/issues/600.

It turned out to be a quite serious issue. The problem was not the `Floating point exception` itself, but accessing memory out of the source image bounds caused by a very stupid mistake. See the first commit.

I also added commit with explicit check for zero in the place where `Floating point exception` was thrown. It's not necessarily needed, but I rather added it as I feel that Mapnik's policy is not to throw whatsoever.

### DIFF
--- a/include/mapnik/span_image_filter.hpp
+++ b/include/mapnik/span_image_filter.hpp
@@ -79,16 +79,19 @@ public:
         {
             base_type::interpolator().coordinates(&x, &y);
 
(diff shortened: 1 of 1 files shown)


springmeyer:
@talaj - awesome work. As a followup could you summarize and add this issue to the CHANGELOG?

PR #3348, opened 2016-03-04
Use ST_Resample for prescaling

strk:
Ensures all tiles are properly aligned.
Fixes #2375 (visible tile boundaries).

PR #3358, opened 2016-03-11
Stub C++ version of pgraster test

strk:
NOTE: this is a much less interesting than the python test

This is a spin-off of #3348, only including the testcase stub.


strk:
BTW, https://github.com/mapnik/python-mapnik/blob/master/test/python_tests/pgraster_test.py should really be ported to C++ for it to be a good test.

lightmare:
I think the `run` function should be local, too. Unit tests are linked into a single executable. Either make it static, or move to anonymous namespace like you did `dbname`.

strk:
Also test name needed to be changed. Should work now.

PR #3375, opened 2016-03-21
Don't auto-open by default

pnorman:
Instead rely on the --open flag to indicate that the user wants to have mapnik-render open the image in a viewer

Also, opening the image is no longer OS X only.


pnorman:
Anything I can do to further this PR?

springmeyer:
thanks, done in 9833bfb and fca1075

PR #3379, opened 2016-03-23
Fix some comments and error messages referencing raster parameters

strk:
User exposed messages are affected, which were giving wrong information about available parameters.
I've also created a wiki page with the actual params: https://github.com/mapnik/mapnik/wiki/PgRaster


lightmare:
Applied in https://github.com/mapnik/mapnik/commit/69c317cd1fa5f531eed9de9d41f78970dbd5233a

PR #3382, opened 2016-03-23
Fix reading nodata value for data rasters

strk:
Without this commit the NODATA value for rasters is set to whatever value had the last pixel read (usually bottom-right one). Particularly funny with tiled rasters...

### DIFF
--- a/plugins/input/pgraster/pgraster_wkb_reader.cpp
+++ b/plugins/input/pgraster/pgraster_wkb_reader.cpp
@@ -269,15 +269,13 @@ mapnik::raster_ptr read_grayscale_band(mapnik::box2d<double> const& bbox,
   image.set(0xffffffff);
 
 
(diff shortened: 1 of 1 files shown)


lightmare:
Some minor notes
 - can you move `nodata` initialization to the variable definition, instead of separate line?
 - consequently, move `val` inside the loop, it's not used outside
 - same treatment [here](https://github.com/strk/mapnik/blob/nodata_for_datarasters/plugins/input/pgraster/pgraster_wkb_reader.cpp#L274) -- note that at the end of this function there's the same :beetle:  `raster->set_nodata(val)`

strk:
Thanks for the review, both fixes pushed.
The grayscale raster was indeed faulty too.
It's a pity there's no automated testfor pgraster in core yet
(#3358 is a step in that direction)

Issue #3387, opened 2016-03-31
Breaking change: no unquoted strings


jchamberlain:
3.0.10 has removed support for unquoted strings in filter expressions (d01da6f86acb0f14de44ac42ed21a073fe5e84cf). We've somewhat accidentally been using them, and were surprised that our maps broke on upgrading, especially since there's no mention of it in the changelog.

If you could update the changelog to reflect that I'd appreciate it, and it might save someone else my surprise. Thanks!

mapsam:
Hey @jchamberlain thanks for bringing this up! The removed support was intentional but we never recorded it in the changelog, and are now seeing this come up elsewhere so we brought the functionality back in for the time being via #3389 and #3390. We are planning on tagging this at `3.0.11`, but the unquoted strings in filter expressions will likely be removed soon.

jchamberlain:
That makes sense. We've already updated our code to quote strings properly--a pretty quick fix--I just wanted to make sure this was publicly recorded somewhere.

mapsam:
Excellent, glad to hear you got a fix in place. We'll make sure the changelog is updated appropriately the next time around!

PR #3389, opened 2016-03-31
concatenating strings & values

mapsam:
A recent core change in Mapnik no longer allows unquoted strings in filter expressions and leads to the following error:

```
Error: Failed to parse expression: <expression>
```

Discovered in the Node Mapnik `3.4.18` bump which included a new version of Mapnik core. This has the potential to break styles in Mapbox Studio Classic, and we are seeing it raise issues for folks over at #3387 

This PR currently is a Work In Progress. It adds test that we will want to have passing before merging. Below are how an expression needs to exist for a test to properly pass, requiring fully quoted strings. Notice the `'Hello'` vs `Hello` and `'m'` vs `m`.

```C++
TRY_CHECK(eval("'Hello' + '!'") == eval("'Hello!'"));
TRY_CHECK(eval("[int]+'m'") == eval("'123m'"));
```

Sliding this over to @artemp for some :eyes: and potential fix.

cc/ @springmeyer @flippmoke @jakepruitt

### DIFF
--- a/test/unit/core/expressions_test.cpp
+++ b/test/unit/core/expressions_test.cpp
@@ -182,4 +182,10 @@ TEST_CASE("expressions")
     // 'Québec' =~ m:^Q\S*$:
     TRY_CHECK(eval(" [name].match('^Q\\S*$') ") == true);
     TRY_CHECK(parse_and_dump(" [name].match('^Q\\S*$') ") == "[name].match('^Q\\S*$')");
+
+    // string & value concatenation
+    // this should evaluate as two strings concatenating, but currently fails
+    TRY_CHECK(eval("Hello + '!'") == eval("'Hello!'"));
+    // this should evaulate as a combination of an int value and string, but fails
+    TRY_CHECK(eval("[int]+m") == eval("'123m'"));
 }


lightmare:
It's not a regression, the change was intentional. See https://github.com/mapnik/mapnik/pull/3287 for reference.

mapsam:
@lightmare you are right, "regression" was not the proper word. Thanks for pointing those changes out. It *is* "breaking" for some users though, as seen in the reference above. Maybe all we need is an update to the changelog so users are aware, but I'll leave that decision to others.

lightmare:
Yes, it should've been mentioned in changelog; the pull fixed an operator precedence regression -- and that part is there -- but the retraction of unquoted strings isn't.

springmeyer:
> the pull fixed an operator precedence regression -- and that part is there -- but the retraction of unquoted strings isn't.

@lightmare - can you think of a clean way to keep the operator precedence fix but restore the support for unquoted strings? It looks like there are existing styles that were depending on this as a feature (which I was not previously aware of). I'm concerned that to maintain back compatibility we'll need unquoted strings working again in v3.0.11.

lightmare:
> can you think of a clean way to keep the operator precedence fix but restore the support for unquoted strings?

Yes, it's easy. Simply add `((alpha >> *alnum) - lit("not"))` to primary expressions. But I explained why having unquoted strings in arbitrary expressions is evil. Even Perl acknowledges that, and that's the language my friends call write-only.

artemp:
@lightmare - totally with you re:`bare strings are evil`. But on a balance looks like we might need to support them for a short while. 

How [this PR](https://github.com/mapnik/mapnik/pull/3390)  looking for everyone ?

/cc @springmeyer @mapsam @jakepruitt ^

lightmare:
Didn't check the old `ustring` rule before, it has `no_skip` which should be in the primary-expr alternative, too:

```c++
unquoted_ustring %= no_skip[alpha >> *alnum] - lit("not");
```

PR #3391, opened 2016-04-01
udpate changelog for bare string re-introduction

mapsam:
Just a quick update to the changelog before tagging `3.0.11` - @artemp all we need is the gitsha that we are tagging from and this should be :+1:


mapsam:
Closed in favor of 8c4892a

PR #3424, opened 2016-05-06
If there're overviews, just use them for rendering

rafatower:
The current code falls back to the base table for small scales.

That approach has some drawbacks cause it forces the original table to
share some conditions with its overviews (same SRID, alginment, scale
constraints) for the rendering to work properly.

What we propose is to always fall back to the highest resolution
overview (lowest scale), rather than the original table, in order to
avoid coupling the original table with the constraints imposed by
rendering and still have them linked (in postgis raster metadata).

Please note this approach is not 100% compatible as the base table
won't be used. This should be no big deal because overviews can have
an arbitrary resolution/scale.

### DIFF
--- a/plugins/input/pgraster/pgraster_datasource.cpp
+++ b/plugins/input/pgraster/pgraster_datasource.cpp
@@ -878,9 +878,10 @@ featureset_ptr pgraster_datasource::features_with_context(query const& q,process
         std::string table_with_bbox;
         std::string col = geometryColumn_;
 
-        if ( use_overviews_ ) {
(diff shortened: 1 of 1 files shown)


rafatower:
I noticed there might be some undesirable side effects of this patch and I'll give it some thought before proceeding.

rafatower:
Reopening this: there's no side effect. I was stumbling again into one particular problem of ours https://github.com/CartoDB/Windshaft/pull/241

artemp:
@rafatower - thanks for contributing!

PR #3465, opened 2016-07-12
skip tests in save_map_test.py that use unavailable input plugins

estadtherr:
This change modifies the "save_map_test" python test to bypass maps that depend on input plugin types that are not included in the build configuration.

This of course could bypass valuable tests for specific rendering capabilities that aren't tested elsewhere. This is an inherent drawback in the current test structure, however, that couples feature tests to specific input plugins (e.g. map test "M" tests a combination of feature "A" and plugin "X", and feature "A" is not tested elsewhere). Solving that coupling was beyond the scope of ensuring tests of different build configurations completed successfully.

### DIFF
--- a/run_tests
+++ b/run_tests
@@ -8,7 +8,7 @@ failures=$((failures+$?))
 
 echo "*** Running C++ tests..."
 for FILE in tests/cpp_tests/*-bin; do 
-	${FILE} -q -d .;
+  ${FILE} -q -d .;
   failures=$((failures+$?))
 done
 
--- a/tests/python_tests/datasource_test.py
+++ b/tests/python_tests/datasource_test.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 from nose.tools import *
-from utilities import execution_path, run_all
+from utilities import execution_path, run_all, datasources_available
 import os, mapnik
 
 def setup():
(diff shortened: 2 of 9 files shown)


lightmare:
This has been addressed in 2.3.x branch (https://github.com/mapnik/mapnik/commit/80d9171bd1dbf733a560a976d21796f8c0785a33). Although your method looks cleaner, wouldn't it be better to just backport the change, so that there's less friction?

estadtherr:
I was a little uncomfortable with the implementation in 80d9171 for a couple reasons:

1. Parsing the exception text is an unreliable approach, which could be easily and unintentionally subverted by either a change to or localization of the text. This would cause the test to fail unexpectedly.
1. It's entirely possible that an exception that happens to contain that same text could be encountered that doesn't necessarily indicate an unconfigured plugin. In that case, the exception should cause the test to fail instead of being suppressed. In fact, this is already a risk due to the fact that a call to `load_map()` without a "type" parameter will throw an exception that contains the same text.
1. The solution assumes that the absence of an input plugin means that the plugin wasn't built. What if the plugin should have been built, but wasn't able to be loaded? That should theoretically cause the test to fail.
1. The solution involves a "coding by exception" anti-pattern, where an exception is expected and handled as if it were a normal condition. One could argue that in the context of a unit test as opposed to production code, this practice could be allowed, but in this case a cleaner solution was readily available.

What if I forward-port the change to the 2.3.x branch (and possibly others) and submit an additional pull request?

lightmare:
I agree with every point you've made. It'd be good to replicate this fix in `python-mapnik` repository as well.

estadtherr:
I will plan to submit an additional pull request to this repository for the 2.3.x branch, and submit a third pull request to mapnik/python-mapnik for that version. Does that sounds like a workable plan?

artemp:
👍

estadtherr:
I updated the pull request to extend the logic I was using in the save_map_test to other test cases that had similar exception-driven logic, so it is ready for review again. Look for two additional pull requests that handle the 2.3.x and 3.0.x branches.

lightmare:
@artemp How important is travis on these old branches? I've been trying to get it working, managed to get it to compile with both clang and gcc. But then clang-compiled tests segfault, no idea why, locally I'm able to run them (clang 3.6, boost 1.55; on travis I tried default clang 3.4 and clang 3.6, boost 1.54).

I think this PR can be merged in, there's nothing it would break. But if 2.x branches are to have more maintenance, it'd be good to have travis working again.

artemp:
@lightmare - looks good to me to merge ^. 
@springmeyer - what are your thinking re:2.2.x travis ? Should we retire it ?

Issue #1409, opened 2012-08-17
Style level and feature level compositing in AGG/Cairo renderers


springmeyer:
We can support advanced compositing modes when rendering. So, this means per-feature pixel compositing can accept custom modes. And we can support intermediate buffers per style so that effects can be applied to groups of features as they are transferred to the main canvas.

So, porter-duff modes + other common blend modes (http://www.w3.org/TR/SVGCompositing/) + any custom gimp modes we want.

This will allow for things like:
- hole cutting:#713 / erasing features with other features:#335
- flexible variety of vector blend modes:#945
- per layer opacity: #314/#609

Issue #1934, opened 2013-07-01
Support style level compositing in Cairo backend


springmeyer:
We should consider the feasibility of supporting style level compositing in the cairo rendering backend.

dch0ph:
@artemp Closed by #3477? [Trying to make sense of the Cairo-related open issues.]

artemp:
@dch0ph yes 👍

PR #3477, opened 2016-08-03
Style level comp-op and opacity for Cairo renderer

talaj:
Refs https://github.com/mapnik/mapnik/issues/1409.


lightmare:
Refs #1934

PR #3481, opened 2016-08-05
revive travis on 2.2.x branch

lightmare:
It compiles with both clang and gcc.

In clang builds, python tests [segfault inside libboost-thread](https://travis-ci.org/lightmare/mapnik/jobs/150031307#L4964) when importing mapnik. It has something to do with gcc-built boost-thread + non-gcc-built module. Here are some threads about similar errors:
http://stackoverflow.com/questions/29855460/how-to-use-boost-threads-with-clang
http://stackoverflow.com/questions/18927939/segfault-during-static-initialization-when-linking-gcc-built-boost-into-an-intel
https://trac.openmodelica.org/OpenModelica/ticket/3322
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56393#c6 

In gcc builds, tests mostly pass.

There's at least one serious bug I found when one of the python tests had to be killed after allocating all memory. It is due to [wkb_reader](https://github.com/mapnik/mapnik/blob/2.2.x/src/wkb.cpp#L172) never checking input size and [this bogus input](https://github.com/mapnik/mapnik/blob/2.2.x/tests/python_tests/geometry_io_test.py#L59).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ before_script:
  - if [[ "${DEBUG}" == True ]]; then export JOBS=$((JOBS/2)); fi;
 
 script:
- - ./configure CXX=${CXX} CC=${CC} XML_PARSER=$XML_PARSER DEBUG=${DEBUG} DEMO=True BENCHMARK=False BINDINGS='python' CPP_TESTS=True CAIRO=True FAST=True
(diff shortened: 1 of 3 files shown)


lightmare:
2.2.x is obsolete

PR #3505, opened 2016-08-25
Fix closing segment of polygon in building symbolizer

talaj:
Building symbolizer is not rendering face for closing segment of a polygon.

Before:
building-symbolizer-1-256-256-1 0-agg
building-symbolizer-2-300-300-1 0-agg

After:
building-symbolizer-1-256-256-1 0-agg-2
building-symbolizer-2-300-300-1 0-agg-2


artemp:
@talaj - great patch

PR #3513, opened 2016-09-07
Fix image swap

lightmare:
mapnik::image::swap didn't swap `pData_` pointers. Instead of fixing that by adding one line, I removed the member, it was just `reinterpret_cast` pointer from `buffer_`

### DIFF
--- a/include/mapnik/image.hpp
+++ b/include/mapnik/image.hpp
@@ -76,7 +76,6 @@ public:
 private:
     detail::image_dimensions<65535> dimensions_;
     detail::buffer buffer_;
-    pixel_type *pData_;
     double offset_;
     double scaling_;
     bool premultiplied_alpha_;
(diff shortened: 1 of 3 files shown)


springmeyer:
Curious: what is the impact of the bug this fixes? Crash or unexpected behavior?

lightmare:
It's both. If you swap two images of the same size, it's unexpected that for example `im.set_row` writes to the other image's buffer. When you swap images of different sizes, you're going to crash soon.

```c++
    mapnik::image_rgba8 im;
    mapnik::image_rgba8 im2(16, 16, true);
    im.swap(im2);
    im(0, 0) = 1; // crash
```

springmeyer:
Great, thanks for the details. Sounds like a good and worthy fix for v3.0.12, which @artemp is planning to tag tomorrow. If you agree please go ahead and assign to the milestone and let me know if you need any help testing.

lightmare:
I think the test I added is currently the only place in mapnik where image swap is actually used. Otherwise this bug would've shown its ugly head long ago. I got hit by it while writing tests for another issue.

Issue #3524, opened 2016-09-27
Features not being rendered properly in Stereographic Projection


flippmoke:
While trying to support @rheh with #3521 -- ran into bug associated with data not being rendered properly in stereographic projection. 

## Code to Reproduce
Download all files in [gist](https://gist.github.com/flippmoke/45575c41b5fe32bfa075f611c9f6f905) and run:

```
npm install
npm test
```

## Test Results

Note: this dataset *does not* have Antarctica
 
Results with custom extent provided:

![custom_extent](https://cloud.githubusercontent.com/assets/1794907/18879826/5667e7d4-849b-11e6-8c94-b28103d2c071.jpeg)

Note that the blue line exists in the above result. Now, note the missing blue line in this result:

![south_pole_extent_datasource](https://cloud.githubusercontent.com/assets/1794907/18879881/80ab2394-849b-11e6-8f83-ac1d154e6e1e.jpeg)

Finally, note the missing countries as the extent expands:

![south_pole_lat_n5](https://cloud.githubusercontent.com/assets/1794907/18879893/8d6f8e80-849b-11e6-9a51-2f8dfa2ebf35.jpeg)

/cc @artemp @talaj

talaj:
The missing line should be solved by https://github.com/mapnik/mapnik/pull/3525.

I think the missing counties are due to change of meaning what is top or down and left or right of an extent after reprojection. Or something like that. If I set negative `buffer-size="-200"` to the layer with countries, it has opposite effect, countries from the middle of the map appears:

![southpolarstylesheet-840-840-1 0-agg2](https://cloud.githubusercontent.com/assets/1950911/18891938/c92841c6-8507-11e6-965c-f367100e10e4.png)

PR #3525, opened 2016-09-27
fix intersecting extents in different projections

talaj:
Partly solves https://github.com/mapnik/mapnik/issues/3524, the missing blue line in particular.

southpolarstylesheet-840-840-1 0-agg

### DIFF
--- a/include/mapnik/feature_style_processor_impl.hpp
+++ b/include/mapnik/feature_style_processor_impl.hpp
@@ -266,6 +266,7 @@ void feature_style_processor<Processor>::prepare_layer(layer_rendering_material 
     }
 
     box2d<double> layer_ext = lay.envelope();
+    const box2d<double> buffered_query_ext_map_srs = buffered_query_ext;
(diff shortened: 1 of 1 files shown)


include/mapnik/feature_style_processor_impl.hpp
```diff
@@ -281,9 +282,9 @@ void feature_style_processor<Processor>::prepare_layer(layer_rendering_material
         early_return = true;
     }
     // next try intersection of layer extent back projected into map srs
-    else if (prj_trans.backward(layer_ext, PROJ_ENVELOPE_POINTS) && buffered_query_ext.intersects(layer_ext))
+    else if (prj_trans.backward(layer_ext, PROJ_ENVELOPE_POINTS) && buffered_query_ext_map_srs.intersects(layer_ext))
```
talaj: The problem here was, that `buffered_query_ext` was reprojected into layer's projection in the `if` a few lines above.

PR #3529, opened 2016-10-05
fix #3526 GroupSymbolizer PairLayout with 3+ items

### LINKED ISSUE #3526: GroupSymbolizer: PairLayout fails with >2 labels?
I've been experimenting with actually using the GroupSymbolizer to put highway shields on coincident routes. I'm somewhat stumped with how to make it work for three or more coincident routes.

For various reasons, I've found it most expedient to create an auxiliary table to hold the data for route relations in a denormalized form, with columns that include 'way', 'highway', 'ref_1', 'network_1', 'ref_2', 'network_2' ... 'ref_8', 'network_8'. (There's actually a fair amount of manipulation that goes into this, to extract the relation information from the slim tables, enumerate coincidences, and coalesce consecutive ways that belong to the same routes.) In the attached tarball, the 'bugrep' directory contains a tiny extract of this table for routes near Augusta, Maine.
[https://kbk.is-a-geek.net/catskills/test3.html?la=42.5988&lo=-74.3336&z=15](http://www.openstreetmap.org/#map=15/44.3146/-69.7765&layers=N) shows how the data render in the default OSM map.

I've created a minimalist stylesheet to demonstrate the problem that I'm having, in the tarball as 'bugrep.xml'. The file 'bugrep.py' is a tiny harness to render a single PNG using that stylesheet. The style depends on having the two SVG's from the tarball as well, because the ShieldSymbolizer has 'file="./bugrep-shield-[network_%].svg"' as one of its options. 

_Side note:_ This is, of course, not the final product! But you really don't want to try to debug this by getting set up to reproduce [https://kbk.is-a-geek.net/catskills/test3.html?la=42.5988&lo=-74.3336&z=15](https://kbk.is-a-geek.net/catskills/test3.html?la=44.3135&lo=-69.7776&z=15) - the map is (I hope) not ugly, but the stylesheet surely is! In any case, it's surely good enough that views like [https://kbk.is-a-geek.net/catskills/test3.html?la=42.5988&lo=-74.3336&z=15](https://kbk.is-a-geek.net/catskills/test3.html?la=42.5988&lo=-74.3336&z=15) (just choosing at random something that I've been working on mapping recently) are useful to me in the field, printed out or downloaded to a smartphone.

What I'm seeing, using version 3.0.9+ds-1 on Ubuntu, is that single routes, like ME 105 at the east side of the rendered map, show up just fine, as do pairs of routes, like the ME 9/US 202 combination at the northeast. If three or more routes are coincident, nothing is rendered. Thus ME 11/ME 17/ME 100/US 202, exiting to the west, or ME 8/ME 11/ME 27 to the north, have no shields at all. 

This behaviour is not what I expect to see from the documentation, which states, "If more than two labels are present, the pairs stack vertically above and below the placement point.". Have I misunderstood what this sentence actually means, or is there something else amiss with my data that keeps larger clusters from rendering?

[bugrep.tar.gz](https://github.com/mapnik/mapnik/files/505470/bugrep.tar.gz)

lightmare:


talaj:
I don't know whether it was intended, but it also fixes a problem when `<GroupRule></GroupRule>` is empty and `<DebugSymbolizer/>` is used. In such situation, the code was generating huge bboxes, placing them into collision detector and causing debug symbolizer to draw effectively to eternity.

@lightmare , can you please rebase?

lightmare:
> it also fixes a problem when <GroupRule></GroupRule> is empty and <DebugSymbolizer/> is used

That's probably just a side-effect of getting rid of all the evil casts (e.g. `safe_cast<int>((member_boxes_.size() - 1) >> 1);` which may be safe but wrong :D). Now it's obvious how many times the loop will run.

PR #3537, opened 2016-10-17
Fix blur with scale_factor

talaj:
Currently, the blur radius is not multiplied by `scale_factor`, which causes different visual output for different resolutions.


artemp:
@talaj - looking good ^ I'm going merge

PR #3566, opened 2016-12-02
Better largest bbox

talaj:
Removes redundant code and uses `std::max_element` instead of `std::sort`, which is always less complex.

### DIFF
--- a/include/mapnik/box2d.hpp
+++ b/include/mapnik/box2d.hpp
@@ -116,6 +116,7 @@ public:
     bool valid() const;
     void move(T x, T y);
     std::string to_string() const;
+    T area() const;
 
     // define some operators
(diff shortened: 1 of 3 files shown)


lightmare:
You should also rename the comparator to `smallest_bbox_first` or use `min_element`. It's confusing now that it's reversed. ;)

The TODO comment is also worth considering. `envelope` is called *2n* times, whereas with a custom loop only *n* would be needed.

talaj:
> You should also rename the comparator to smallest_bbox_first or use min_element. It's confusing now that it's reversed. ;)

True, thanks.

> The TODO comment is also worth considering. envelope is called 2n times, whereas with a custom loop only n would be needed.

Good point.

talaj:
Updated.

Issue #3573, opened 2016-12-14
x3 usage leads to linking errors on Windows


springmeyer:
The windows builds are currently failing with:

```
..\..\src\svg\svg_parser.cpp(115): error C2995: 'bool mapnik::svg::grammar::parse_rule(unknown-type,Iterator &,const Iterator &,const Context &,Attribute &)': function template has already been defined [C:\projects\mapnik\mapnik-gyp\build\mapnik.vcxproj]
  ..\..\src\svg\svg_parser.cpp(114): note: see declaration of 'mapnik::svg::grammar::parse_rule'
```

https://ci.appveyor.com/project/Mapbox/mapnik/build/1.0.1463#L267

@artemp - this looks similar to the linking problems you hit previously on unix systems. Do you have ideas to fix?

artemp:
Doh, `c++14` support in VC++ is patchy to put it mildly.  Here is potential workaround : 
https://github.com/boostorg/spirit/commit/c51a013c412f247731e238617c57590992579d55

artemp:
@springmeyer @BergWerkGIS - tracking windows gotchas in https://github.com/mapnik/mapnik/tree/windows 

@BergWerkGIS - what is the timeline for c++14 support on windows ? Are there any updates in the pipeline?  I know I'm asking too much but I thought you might have some extra info re: Visual Studio upcoming releases.

springmeyer:
Capturing sync with @artemp for next actions on maintaining windows support:

  - Let's test master with latest visual studio - if that works we'll plan to wait for visual studio versions to support c++14 rather than trying to patch
  - Let's confirm that Mapnik pre-x3 (when only c++11 was needed) works with visual studio 2015 as-is. I've pushed a branch called `v3.0.x` to ensure this is the case.
   - Then we'll replay some patches from master onto that branch and ensure they still work on windows

springmeyer:
Looks like branching from before spirit-x3 merge passes (per the gitsha used for https://github.com/mapnik/mapnik/issues/3576#issue-195691907): https://ci.appveyor.com/project/Mapbox/mapnik/build/1.0.1472


Also testing replaying the non-spirit x3 related to commits onto that: https://ci.appveyor.com/project/Mapbox/mapnik/build/1.0.1473

BergWerkGIS:
Tried to build locally with _VS20**17**_ today and encountered these errors (with `boost@1.62`):

## `@master`

```
[...]

PR #3585, opened 2017-01-03
minor render_thunk_list change

lightmare:
The first commit is just a minor optimization, it removes unnecessary wrapping of `render_thunk` in `unique_ptr`.

The second replaces `std::vector` with `std::list` in `render_group_symbolizer`, because vector insertion requires its elements to be either nothrow_move_constructible, or copy_constructible. `render_thunk_list` isn't copy_constructible, and whether it's nothrow_move_constructible depends on STL implementation of `std::list`. This should fix one of the issues with MSVC 2017 (refs #3573)

### DIFF
--- a/include/mapnik/renderer_common/render_group_symbolizer.hpp
+++ b/include/mapnik/renderer_common/render_group_symbolizer.hpp
@@ -42,9 +42,9 @@ struct render_thunk_list_dispatch
     {
         offset_ = offset;
 
-        for (render_thunk_ptr const& thunk : thunks)
+        for (render_thunk const& thunk : thunks)
         {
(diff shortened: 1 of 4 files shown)


lightmare:
After looking a bit more into the standard, I have to rectify what I wrote above. Now I think the original code is perfectly valid. The vector shouldn't attempt to use a deleted copy constructor, it should just use the throwing move constructor. More info in the issue referenced above.

That being said, I think the change is worth keeping. The overhead is minimal, and vector of elements with throwing move is kinda dangerous -- although not in this particular use case, it's relatively easy to shoot oneself in the foot.

PR #3594, opened 2017-01-07
"X" EXACT_POSITION should not use dx/dy...

cquest:
When using "X" in simple placement-type, the text is always moved event if "X" is the first displacement in the list...

### DIFF
--- a/src/text/text_layout.cpp
+++ b/src/text/text_layout.cpp
@@ -62,7 +62,7 @@ pixel_position evaluate_displacement(double dx, double dy, directions_e dir)
     switch (dir)
     {
     case EXACT_POSITION:
-        return pixel_position(dx,dy);
+        return pixel_position(0,0);
         break;
     case NORTH:
         return pixel_position(0,-std::abs(dy));


talaj:
I agree that it is not possible to ignore `dx` and `dy`. But `EXACT_POSITION` perhaps means exact application of `dx` and `dy`. I have used new `CENTER` position for that in downstream 83474326fd85b4f300a21fe2235389b47c0a759c.

cquest:
If the meaning of "X" is to apply both displacements by default, the documentation/wiki should be updated to make things clearer because X can be replaced by the corresponding NW/SW/NE/SE equivalent depending on dx/dy to obtain the same result where the "default" position cannot currently be obtained in any way.

Having an option to try to place a label at the original position is very useful.
If "X" behaviour is not to be changed, your "CENTER" is worth a pull-request. Have you proposed one ?

talaj:
> If "X" behaviour is not to be changed, your "CENTER" is worth a pull-request. Have you proposed one ?

I just did in https://github.com/mapnik/mapnik/pull/3596. Maybe `ORIGIN` would be more appropriate. Let maintainers decide.

talaj:
@cquest, https://github.com/mapnik/mapnik/pull/3596 is merged.

artemp:
Closing here as both `CENTER` (pixel_position(0, 0)) and `EXACT` (pixel_position(dx,dy)) options are available in master
@talaj @cquest

PR #3618, opened 2017-02-10
Postgis parameter quoting and interpolation

### LINKED ISSUE #3603: Query with character @ doesn't render
I have this query:
```
<Parameter name="table">(SELECT * FROM dataset WHERE added_by = 'somebody@example.com') AS suquery</Parameter> 
```
which doesn't render.
AFAIK, @ is not a character that needs to be escaped in XML.

I'm running mapnik 3.0.12. Is this a known issue? If so, can you recommend a workaround?

lightmare:
This started out as an attempt to solve #3603 but ended up much broader.

#### Breaking change 1: interpolation of attributes aka render-time `@variables`
Previously
```sql
... WHERE col = @foo AND email = 'bar@sue.me'
-- would be transformed into
... WHERE col = _value_of_foo_ AND email = 'bar_value_of_sue_.me'
```

Now in order for `@variable` to be interpolated, it has to be wrapped in `!`, and the injected value will be quoted:
```sql
... WHERE col = !@foo! AND email = 'bar@sue.me'
-- will be transformed into
... WHERE col = '_value_of_foo_' AND email = 'bar@sue.me'
```

#### Breaking change 2: parameter `geometry_table` quoting
Previously
```sh
<Parameter name="geometry_table">schema-foo.table-bar</Parameter>
```
might've worked. Now it needs to be quoted in the same way it'd need to be quoted in an SQL query, i.e.:

```sh
<Parameter name="geometry_table">"schema-foo"."table+bar"</Parameter>

<Parameter name="geometry_table">simple_identifiers.dont_need_quoting</Parameter>
```

What lead to this:
The `geometry_table_` member variable contained either the value of the datasource parameter `geometry_table`, or only the table name (if the parameter was schema-qualified), or a table name extracted from datasource parameter `table` (where it had to be quoted), or even the whole query in `table` (if `sql_utils::table_from_sql` failed to find a name). It confused me when I was trying to figure out how it should be quoted in various places. So I made `geometry_table_` retain the value from the parameter, putting extracted schema and table name into separate variables. And it made sense to do this extraction in a common way (from either `geometry_table` or `table`) right inside `table_from_sql`, hence the need for `geometry_table` parameter to be "properly SQL-quoted".

#### Side effects
- drops postgis plugin's dependency on `boost::regex`
- same changes applied to pgraster plugin (which didn't have `@var` interpolation at all)
- fixes #3603

### DIFF
--- a/include/mapnik/sql_utils.hpp
+++ b/include/mapnik/sql_utils.hpp
@@ -39,6 +39,41 @@
 
 namespace mapnik { namespace sql_utils {
 
+    struct quoted_string
+    {
+        std::string const* operator-> () const { return &str; }
+        std::string const& str;
+        char const quot;
+    };
+
+    inline quoted_string identifier(std::string const& str)
+    {
+        return { str, '"' };
+    }
+
+    inline quoted_string literal(std::string const& str)
+    {
+        return { str, '\'' };
+    }
+
+    inline std::ostream& operator << (std::ostream& os, quoted_string qs)
+    {
+        std::size_t pos = 0, next;
+
+        os.put(qs.quot);
+        while ((next = qs->find(qs.quot, pos)) != std::string::npos)
+        {
+            os.write(qs->data() + pos, next - pos + 1);
+            os.put(qs.quot);
+            pos = next + 1;
+        }
+        if ((next = qs->size()) > pos)
+        {
+            os.write(qs->data() + pos, next - pos);
+        }
+        return os.put(qs.quot);
+    }
+
     inline std::string unquote_double(std::string const& sql)
     {
         std::string table_name = sql;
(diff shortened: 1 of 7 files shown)


springmeyer:
These breaking changed I've added https://github.com/mapnik/mapnik/wiki/API-changes-between-v3.0-and-v3.1. @lightmare - can you add any others you are aware of?

lightmare:
@springmeyer I made a correction to the note about quoting (it's not just reserved words, it's non-word-like table names in general that need quoting as well), added examples to make it clearer, and added changes to pgraster plugin.

lightmare:
I just realized something important.
Previously it was possible to inject not just values into SQL, but use variable as column name for example:
```
SELECT geom, @colname FROM ...
```
But now this will single-quote the injected value, so will return the name of the column you want, not data from it.
I don't know how useful this feature was, but it was possible and someone might have found a good use for it. Now there's no way to do this.

PR #3636, opened 2017-03-07
Color emoji + updated harfbuzz_shaper and unicode_string_grammar  

artemp:


artemp:
@shawmat 👍

test/unit/text/shaping.cpp
```diff
@@ -65,39 +65,39 @@ TEST_CASE("shaping")
     mapnik::face_manager fm(fl, font_file_mapping, font_memory_cache);
 
     {
-        std::vector<std::tuple<unsigned, unsigned>> expected =
-            {{{0, 0}, {0, 3}, {0, 4}, {0, 7}, {3, 8}, {11, 9}, {68, 10}, {69, 11}, {70, 12}, {12, 13}}};
+        std::vector<std::pair<unsigned, unsigned>> expected =
+            {{0, 0}, {0, 3}, {0, 4}, {0, 7}, {3, 8}, {11, 9}, {68, 10}, {69, 11}, {70, 12}, {12, 13}};
         // with default NotoSans-Regular.ttc and NotoNaskhArabic-Regular.ttf ^^^
```
lightmare: @artemp the reason this works with pair but not tuple is because pair is an aggregate, tuple isn't (because it has a base). If you ever need to do this with more than a pair of values, or another non-aggregate-type, the trick is to specify the type of the first element; subsequent elements don't need the type repeated, so: 
```c++
std::vector<std::tuple<unsigned, unsigned>> expected = {std::make_tuple(0u, 0u), {0, 3}, {0, 4}};
// and they don't even need to be put in a vector,
// you can just capture the `std::initializer_list<T>` by reference
auto & expected = {std::tuple<unsigned, unsigned>{0, 0}, {0, 3}, {0, 4}};
```

lightmare:
It'd be better to not introduce more dependency on `boost::regex`, I've recently submitted two PRs that remove the current dependencies. ICU can do the conversion from UTF-16, and we can't get rid of ICU, but we can easily get rid of `boost::regex`.

artemp:
@lightmare - yes, I saw your PR, thanks!  I have to say I'm not convinced that ditching boost_regex and relying on ICU regex is a right approach in a long term. I have to study this issue more to get back to your on this but meanwhile using boost header is totally fine.

test/unit/text/shaping.cpp
artemp: @lightmare - yep, thanks  `initializer_list` reference sounds great ^

artemp: @lightmare - changed ^ 47763afef31e9d8390d4d1079825143f17e41788

artemp: @lightmare - any ideas ? 
```
^~~~~~
/usr/lib/gcc/x86_64-linux-gnu/5.4.1/../../../../include/c++/5.4.1/tuple:612:19: note: constructor declared here
        constexpr tuple(_U1&& __a1, _U2&& __a2)
                  ^
test/unit/text/shaping.cpp:69:84: error: chosen constructor is explicit in copy-initialization
            {std::tuple<unsigned, unsigned>{0, 0}, {0, 3}, {0, 4}, {0, 7}, {3, 8}, {11, 9}, {68, 10}, {69, 11}, {70, 12}, {12, 13}};
```

artemp: ok, trying `std::tuple<unsigned, unsigned>(0, 0)`  ..

lightmare: My bad, I thought I had used it in another test previously, turns out I had it stashed so not really tested on multiple compilers. Now I checked and it works for me with both gcc and clang (but I think it's using libstdc++ as well).

lightmare: Oh, gcc 5.4 :no_entry: I stopped testing with gcc 5 after recent reports that 5.3 can't compile x3 grammars, so I deemed the gcc-5 series insufficient.

artemp: nope, didn't work across all compilers - reverted back to good old pair 791f7b51bbf99bbd303367965b53364709f58af7

PR #3660, opened 2017-04-23
add twkb support to sqlite plugin

rinigus:
This PR adds TWKB support for SQLite datasource. To use it, convert geometry BLOBs into TWKB and specify twkb as a format in the datasource wkb_format option.


artemp:
thanks @rinigus !

PR #3668, opened 2017-05-09
Raster Overzoom Quest

flippmoke:
First attempt at some code to properly handle the overzoom and scaling problems. Please note that I added a new test case on the [test-visual-data branch](https://github.com/mapnik/test-data-visual/commits/overzoom_fix). It is not complete and we should add a few others to deal with other special cases. 

/cc @artemp @springmeyer

### DIFF
--- a/include/mapnik/raster.hpp
+++ b/include/mapnik/raster.hpp
@@ -40,15 +40,27 @@ class raster : private util::noncopyable
 {
 public:
     box2d<double> ext_;
+    box2d<double> query_ext_;
     image_any data_;
     double filter_factor_;
     boost::optional<double> nodata_;
+    
+    template <typename ImageData>
+    raster(box2d<double> const& ext,
+           box2d<double> const& query_ext,
+           ImageData && data,
+           double filter_factor)
+        : ext_(ext),
+          query_ext_(query_ext),
+          data_(std::move(data)),
+          filter_factor_(filter_factor) {}
 
     template <typename ImageData>
     raster(box2d<double> const& ext,
            ImageData && data,
            double filter_factor)
         : ext_(ext),
+          query_ext_(ext),
           data_(std::move(data)),
           filter_factor_(filter_factor) {}
 
@@ -71,7 +83,6 @@ public:
     {
         filter_factor_ = factor;
     }
-
 };
 }
 
--- a/include/mapnik/renderer_common/process_raster_symbolizer.hpp
+++ b/include/mapnik/renderer_common/process_raster_symbolizer.hpp
@@ -53,6 +53,7 @@ struct image_dispatcher
     image_dispatcher(int start_x, int start_y,
                           int width, int height,
                           double scale_x, double scale_y,
+                          double offset_x, double offset_y,
                           scaling_method_e method, double filter_factor,
                           double opacity, composite_mode_e comp_op,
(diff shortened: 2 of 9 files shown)


artemp:
@flippmoke - looking great!

include/mapnik/raster.hpp
```diff
@@ -40,15 +40,27 @@ class raster : private util::noncopyable
 {
 public:
     box2d<double> ext_;
+    box2d<double> query_ext_;
```
artemp: @flippmoke - can we get away with just passing x/y offsets e.g `point<double>` ?

include/mapnik/renderer_common/process_raster_symbolizer.hpp
```diff
@@ -210,12 +215,17 @@ void render_raster_symbolizer(raster_symbolizer const& sym,
     if (source)
     {
         box2d<double> target_ext = box2d<double>(source->ext_);
-        prj_trans.backward(target_ext, PROJ_ENVELOPE_POINTS);
+        box2d<double> target_query_ext = box2d<double>(source->query_ext_);
+        if (!prj_trans.equal()) {
```
artemp: pls, follow mapnik formatting style :)

plugins/input/raster/raster_featureset.cpp
```diff
@@ -89,33 +89,36 @@ feature_ptr raster_featureset<LookupPolicy>::next()
...
-                        intersect = t.backward(feature_raster_extent);
-                        mapnik::image_any data = reader->read(x_off, y_off, width, height);
-                        mapnik::raster_ptr raster = std::make_shared<mapnik::raster>(intersect, std::move(data), 1.0);
-                        feature->set_raster(raster);
+                    // select minimum raster containing whole ext
+                    int x_off = static_cast<int>(std::floor(ext.minx()));
+                    int y_off = static_cast<int>(std::floor(ext.miny()));
+                    int end_x = static_cast<int>(std::ceil(ext.maxx()));
+                    int end_y = static_cast<int>(std::ceil(ext.maxy()));
+
```
artemp: @flippmoke - did you figure out why we need `floor/ceil` here ?

include/mapnik/raster.hpp
flippmoke: No, because we need to know when it stops on the other side as well, so we need to know two full extents.

plugins/input/raster/raster_featureset.cpp
flippmoke: No I did not, it is something that I still need to dig into -- I probably should just draw it out that always helps the most.

flippmoke:
After a lot of thought, I am not going to be updating the pgraster plugin as part of this issue. The reason. I have left the additional constructor option in the `raster` object so code will continue to operate as expected and I believe nothing should change from the current code in pgraster. After speaking with @springmeyer we considered that we might be pulling the pgraster plugin out in `3.1.x` and the lack of proper tests here, I am fairly confident that we shouldn't make any changes.

springmeyer:
> After speaking with @springmeyer we considered that we might be pulling the pgraster plugin out in 3.1.x and the lack of proper tests here, I am fairly confident that we shouldn't make any changes.

See https://github.com/mapnik/mapnik/issues/3674

Issue #3677, opened 2017-05-14
Polygons no longer closed after on-the-fly reprojection


mitchellgreig:
This seems related to #1399 and possibly even #1339 (at least that is where I ended up when trying to troubleshoot the issue. This may already be noted in the myriad of issues around clipping (if it is, in fact, a clipping issue), but could not find helpful information anywhere. If this is already an issue, at least this may help future googlers.

I am not sure of the Mapnik version, but I am using node-mapnik 3.6.0

Polygons with native projection in 4269 being rendered in a 3400 (an Alberta 10TM projection), lose their closedness and LineSymbolizer fails to render to border to completion (in every case skipping the last line segment of the polygon). 

A workaround is to define the layer as a subquery and have PostGIS do the transformation. I cannot verify if this issue is found in other data sources.

talaj:
I don't think it's a clipping issue, rather it seems like closedness is lost during reprojection. Actually a workaround for me was to set `clip="true"`. The clipping algorithm fixes unclosed polygon.

mitchellgreig:
Wow, thanks, that's a good workaround too and with less fuss. Somewhere in the various eras of information I got the impression that clip was defaulted to true.... also got mixed up in the line-clip vs. clip thing as well.  Thanks again.

mitchellgreig:
As an update, for any future lookers. clip="true" isn't a great choice. If your projection isn't oriented to latitude and longitude it will create artifacts. Essentially, if you have a square that crosses the upper or lower border, it will draw a new line to end it that doesn't line up to the viewing extent. So across the top and bottom of your viewing area it will render boundaries that don't exist. In one case, with water features, it would add a sliver polygon that didn't exist at all.

talaj:
@mitchellgreig Proposed fix is in https://github.com/mapnik/mapnik/pull/3678.

Issue #3666, opened 2017-05-05
Loosing lines during polygon outline rendering with LineSymbolizer and coordinates transformation


alexey-gl:
Hi!
I've found a problem with rendering polygon outline with LineSymbolizer. 
The last line of the polygon was lost.

With several tests I determined that the problem occurred when:
**1.** The layer and map are in the different coordinate systems. 
**2.** The coordinate system has central meridian `lon_0` > 90. This is the most strange for me.

In the example below I use epsg:4326 for layer data and epsg:2330, epsg:2354 for map.

```xml
<?xml version="1.0" encoding="utf-8"?>

 <!-- EPSG:2330 -->
 <!--
 map:   EPSG:2330 (lon_0 > 90)
 layer: EPSG:4326
 Lost points!
 -->
 <Map srs="+proj=tmerc +lat_0=0 +lon_0=93 +k=1 +x_0=16500000 +y_0=0 +a=6378140 +b=6356755.288157528 +units=m +no_defs" background-color="rgba(0,0,0,0.0)" background-image-comp-op="src-over" background-image-opacity="1" buffer-size="128" base="/storage/templates">

 <!--
 map:   EPSG:2354 (lon_0 <= 90)
 layer: EPSG:4326
 OK!
 -->
 <!-- <Map srs="+proj=tmerc +lat_0=0 +lon_0=90 +k=1 +x_0=30500000 +y_0=0 +a=6378140 +b=6356755.288157528 +units=m +no_defs" background-color="rgba(0,0,0,0.0)" background-image-comp-op="src-over" background-image-opacity="1" buffer-size="128" base="/storage/templates"> -->

  <Style name="user_data_style" filter-mode="all" opacity="1" image-filters-inflate="false" comp-op="src-over">
    <Rule>
      <LineSymbolizer stroke="rgb(0,0,0)" stroke-width="3" stroke-linejoin="round" stroke-linecap="round"/>
      <!-- <PolygonSymbolizer /> -->
    </Rule>
  </Style>

  <Layer name="user_data" srs="+init=epsg:4326" status="true" clear-label-cache="false" queryable="false" cache-features="false" group-by="" buffer-size="0">
    <StyleName>user_data_style</StyleName>
    <Datasource>
      <Parameter name="layer">OGRGeoJSON</Parameter>
[...]

Issue #1339, opened 2012-07-23
LineSymbolizer doesn't close polygons


tobwen:
When using LineSymbolizer to draw the outline of a polygon with Cairo, the lines in the polygon doesn't seem to be closed as expected:
![example image](http://imagebin.org/index.php?mode=image&id=221895)

mapnik: v2.0.1
cairo: 1.10.2
cairo-wrapper: 1.8.4
python-cairo: 1.8.8
geos: v3.2.0

Tested on PDF output, viewer is latest Adobe Reader on Windows XP.

Edit: also occures on AGG

mitchellgreig:
Has there ever been any changes or workarounds with this? AGG renderer polygons (in node-mapnik) still are showing up not closed for me. Tried reversing vertices, using line-clip set to false, dynamically recreating polygon geometry on the fly, and definitely checked the underlying polygon validity.

talaj:
Should be fixed by https://github.com/mapnik/mapnik/pull/3678.

PR #3678, opened 2017-05-16
fix closing polygons on reprojection

talaj:
Fixes https://github.com/mapnik/mapnik/issues/3677, https://github.com/mapnik/mapnik/issues/3666, https://github.com/mapnik/mapnik/issues/1339.

Tests will come soon.


flippmoke:
If this fix works, lets also roll this back into `3.0.x` branch.

talaj:
I have added some unit tests.

The problem manifests only with projections for which extent doesn't contain coordinates `(0, 0)` because the `(0, 0)` coordinates are used for closing segment. Reprojecting `(0, 0)` in such projections leads to an error, skipping the closing segment so the polygon is left unclosed.

Also marker placement vertex-last has problem with closing segment. I fixed that and added some unit tests.

artemp:
Thanks @talaj !

Issue #3412, opened 2016-04-26
Rendering tiff file problem


Destanek:
I have rendered with mapnik style earlier and have no problem with it
![0](https://cloud.githubusercontent.com/assets/17759444/14815001/17745e5e-0bd3-11e6-8d85-95e617134856.png)

but when I use MapQuest Style which include mq_world_mercator_x.tif at the low level zoom the output is this
![0](https://cloud.githubusercontent.com/assets/17759444/14815048/51a07784-0bd3-11e6-9834-1f4a374e974c.png)

the source tif file look like this
![world](https://cloud.githubusercontent.com/assets/17759444/14815242/72eebde6-0bd4-11e6-8e2d-408eb2f8bf0c.png)

you can see a xml file of layer-base.xml in [here](https://github.com/MapQuest/MapQuest-Mapnik-Style/blob/master/mapquest_inc/layer-base.xml.inc)
This problem only cause when render a low level zoom that include these tif file 

by the way I use generate_tile.py to render and change the render_size to 512 I don't know If it's cause this problem or not but it's work with mapnik style

PhilHa:
I've got the same problem.
Have you found a solution already?

Destanek:
No, I haven't found yet. I tried to change the Projection but it didn't work so I don't have any idea for now.

lightmare:
Looks like an issue with flipped tile y-coordinates. Where can I get the failing tif? The link at https://github.com/MapQuest/MapQuest-Mapnik-Style doesn't work.

Destanek:
You can download it from here 
[http://aweble.de/downloads/mercator_tiffs.tar.bz2](url)

PhilHa:
But it's not just a reversed y-order.
On Zoomlevel 4 and 5 there are slipped stripes (and correct y-order apart from that).

Zoom 4:
![zoom4](https://cloud.githubusercontent.com/assets/5525659/17920074/6adae330-69d2-11e6-8e61-2fe83d27c86a.PNG)

Zoom 5:
![zoom5](https://cloud.githubusercontent.com/assets/5525659/17920076/6eef6b44-69d2-11e6-92f4-b82580c7a540.PNG)

lightmare:
[...]

PR #3679, opened 2017-05-17
The last stripe of a tiff could be read incorrectly due to a swap

flippmoke:
I was looking at our code currently and how we have been using `std::swap_range` within the tiff reader for the reading of RGBA. I believe that our logic is flawed because the temporary buffer's order could be reversed. However, reading the TIFF reader logic give an idea why we have not seen any issues:

http://www.libtiff.org/man/TIFFReadRGBAStrip.3t.html

> When reading a partial last strip in the file the last line of the image will begin at the beginning of the buffer.

We always expected this to be shifted in the read buffer therefore we could be reading the wrong pixels. However, this has been often producing "good" results because I believe we are simply reading a portion of the result from the previous stripped section we read. Therefore, the last lines would often likely just appear as those they were repeats.

My change I believe makes this work differently for the stripped case of RGBA8, but we still would need to make this change for the tiled case as well. 

/cc @artemp @springmeyer

### DIFF
--- a/src/tiff_reader.cpp
+++ b/src/tiff_reader.cpp
@@ -497,6 +497,7 @@ struct tiff_reader_traits
     using image_type = T;
     using pixel_type = typename image_type::pixel_type;
 
+    constexpr static bool reverse = false;
     static bool read_tile(TIFF * tif, std::size_t x, std::size_t y, pixel_type* buf, std::size_t tile_width, std::size_t tile_height)
     {
         std::uint32_t tile_size = TIFFTileSize(tif);
@@ -513,8 +514,9 @@ struct tiff_reader_traits
 template <>
 struct tiff_reader_traits<image_rgba8>
 {
+    using image_type = image_rgba8;
     using pixel_type = std::uint32_t;
-
+    constexpr static bool reverse = true;
     static bool read_tile(TIFF * tif, std::size_t x0, std::size_t y0, pixel_type* buf, std::size_t tile_width, std::size_t tile_height)
     {
         if (TIFFReadRGBATile(tif, x0, y0, buf) != -1)
@@ -530,16 +532,7 @@ struct tiff_reader_traits<image_rgba8>
 
     static bool read_strip(TIFF * tif, std::size_t y, std::size_t rows_per_strip, std::size_t strip_width, pixel_type * buf)
     {
-        if (TIFFReadRGBAStrip(tif, y, buf) != -1)
-        {
-            /*
-            for (std::size_t y = 0; y < rows_per_strip/2; ++y)
-            {
-                std::swap_ranges(buf + y * strip_width, buf + (y + 1) * strip_width, buf + (rows_per_strip - y - 1) * strip_width);
-            }*/
-            return true;
-        }
-        return false;
+        return (TIFFReadRGBAStrip(tif, y, buf) != 0);
     }
 };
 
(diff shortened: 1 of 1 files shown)


artemp:
@flippmoke - well spotted! Yes, using swap_ranges wasn't correct for stripped reads. The fix you suggested is not great as it brings back code duplication. I'll push a better fix and amend tests to catch this case here. I don't think the logic is flawed for `TIFFReadRGBATile` but I need to check.

lightmare:
> we have not seen any issues

This is not accurate. #3412 clearly shows `read_stripped` was broken. I had a fix almost ready -- not a simple one, I completely rewrote it, abandoning TIFFReadRGBAStrip/Tile entirely, because with those you're flipping top-down (which is the default) images twice -- but didn't have time to go back and finalize tests...

artemp:
@lightmare - if you think it's still broken in master, please, post a test image.

lightmare:
Yes, it's still broken in master. You can use existing visual test like I wrote in the referenced issue -- you'll get differently broken results with different BLOCKYSIZE (rows per strip)
```sh
gdal_translate -co BLOCKYSIZE=122 test/data/raster/river_merc.tiff /tmp/river_merc.tiff
mv /tmp/river_merc.tiff test/data/raster/
test/visual/run --agg --styles=tiff-opaque-edge-raster2
```

BLOCKYSIZE=122 makes the error most apparent, as it cuts the last strip in half
BLOCKYSIZE=1 test failure proves reference images are broken as well

artemp:
@lightmare - I can't seem to replicate this. It's hard to spot any defects using `river_merc.tiff`,could try image generated with
```python
#!/usr/bin/env python3

import mapnik

if __name__ == "__main__":
    #size = [64, 64]
    size = [969, 793]
    tile_size = 122
    index = 0
    colors = [mapnik.Color("red"), mapnik.Color("blue"), mapnik.Color("green"), mapnik.Color("yellow")]
    im = mapnik.Image(size[0], size[1])
    im.fill(mapnik.Color("orange"))
    for j in range(0, int(size[0]/tile_size)):
        index +=1
        for i in range(0, int(size[1]/tile_size)):
            index += 1
            for x in range(0,tile_size):
                for y in range(0, tile_size):
                    im.set_pixel(i*tile_size + x, j*tile_size + y, colors[index % 4])

    im.save('output.tif', 'tiff')
    im.save('output.png')

```


 and post wrong rendering pls?


stripped-bug

artemp:
@lightmare - sorry I was talking about https://github.com/mapnik/mapnik/tree/tiff-multiband-grey not master!!

lightmare:
> It's hard to spot any defects using river_merc.tiff,could try image generated with

It's quite obvious when you use BLOCKYSIZE such that `HEIGHT % BLOCKYSIZE` is more than a few pixels. With `river_merc.tiff` and BLOCKYSIZE=122, the river is clearly disconnected at many points in `tiff-opaque-edge-raster2-*-agg.png`.

The above image with colorful tiles 122x122 plays with the bug when BLOCKYSIZE=122. You have to use BLOCKYSIZE (rows-per-strip) that divides neither image height, nor "visual artifact" height.

This is what I got from `tiff-opaque-edge-raster2` after replacing `river_merc.tiff` with the above image converted to different rows-per-strip tiffs:

`convert -define tiff:rows-per-strip=123 colorful.png river_merc.tiff`
tiff-opaque-edge-raster2-969-793-1 0-agg

`convert -define tiff:rows-per-strip=176 colorful.png river_merc.tiff`
tiff-opaque-edge-raster2-969-793-1 0-agg

lightmare:
alright, seems fixed in tiff-multiband-grey

artemp:
@lightmare - my bad getting mixed up with branches, thanks for verifying!

PhilHa:
Should this be merged into current release (v3.0.14) already?

artemp:
@PhilHa - it should ^.

PhilHa:
Because I still get stuff like that ;-)
grafik
It seems, that it's not merged into 3.0.14

flippmoke:
@PhilHa it would be super helpful here if you could make a simple test case in the visual tests! I would love to dive into this right away.

PhilHa:
As I don't have mapnik running in an IDE, I can't create a test right now,
but you can reproduce it with the open mapquest style from here:
https://github.com/MapQuest/MapQuest-Mapnik-Style
and you'll need the tiff files from here too:
http://aweble.de/downloads/mercator_tiffs.tar.bz2

flippmoke:
@PhilHa how are you building `3.0.14`?

PhilHa:
```
wget "https://github.com/mapnik/mapnik/releases/download/v3.0.14/mapnik-v3.0.14.tar.bz2"
tar xvf mapnik-v3.0.14.tar.bz2
cd mapnik-v3.0.14
./configure CC="gcc" CXX="g++-5" XMLPARSER=libxml2 && JOBS=8 make && make install

**Comment by artemp** [comment]
@PhilHa - yes, I there's a bug in `stripped` TIFF swapping code : https://github.com/mapnik/mapnik/blob/master/src/tiff_reader.cpp#L809

```c++
 std::size_t num_rows = std::min(end_y - y, static_cast<std::size_t>(rows_per_strip_));
```
`end_y` is wrong, it should be actual strip height which is `height_`. 

 I'll will push fix asap and we'll have to do a point release `3.0.15` soon after.
Thanks a bunch for reporting !!

Issue #1956, opened 2013-07-19
Statistics processor to enable details on rendering time at runtime


springmeyer:
We should provide an easy runtime way to get details of how long it takes to query data and render it. This ticket replaces the older #780.

Steps on this path:
- [x] - remove the compile time stderr output in `feature_style_processor_impl.hpp` since it is not flexible, makes reading and refactoring the rendering code harder, and really needs to be a separate renderer/processor to answer more meaningful questions (like how long does it take to query data as separated from rendering it).
- [ ] - Design a new statistics_processor that will return data when called (to be easily aggregated) instead of printing to stderr.

javisantana:
Hey @springmeyer is there any design doc or any idea on how to progress with this? I think this is not only a rendering problem, it makes sense for input plugins also so you can have different metrics.

I see there is a MAPNIK_STATS being used by some input plugins but not sure if that's the way you want to go (https://github.com/mapnik/mapnik/blob/master/plugins/input/postgis/postgis_datasource.cpp#L265)

pramsey:
Any guidance on design/approach you could provide would be much appreciated @springmeyer -- if I go at this myself I'm pretty much guaranteed to make a hash of it (and not a hashtable either, more like a breakfast hash (and not a tasty one either)).

PR #3705, opened 2017-06-19
[WIP] Performance stats

rafatower:
These changes are WIP but I'd really like to hear your opinions to iterate on them. The changes aim at covering some of the concerns of https://github.com/mapnik/mapnik/issues/1956

Basically this PR adds a `stats_timer` class. The objects of that class track the CPU and wall time between either the creation and destruction of them, or up to a call to `stop()`. The main difference from existing `timer` class (from which it inherits) is that they store the metrics in a singleton `timer_stats`, from which the stats can later be recovered.

It also adds a few stat timers to measure things I was particularly interested in, and a standalone test for the feature (as I was having some trouble linking against the singleton).

Any comment (especially re. approach, design and API of the classes) is more than welcomed.

### DIFF
--- a/include/mapnik/performance_stats.hpp
+++ b/include/mapnik/performance_stats.hpp
@@ -61,8 +61,8 @@ public:
     timer_metrics get(std::string const& metric_name);
     void reset(std::string metric_name);
     void reset_all();
-    std::string dump();
-    std::string flush();
+    metrics_hash_t dump();
+    metrics_hash_t flush();
 
 private:
     friend class CreateStatic<timer_stats>;
--- a/include/mapnik/performance_stats.hpp
+++ b/include/mapnik/performance_stats.hpp
@@ -61,8 +61,6 @@ public:
     timer_metrics get(std::string const& metric_name);
     void reset(std::string metric_name);
     void reset_all();
-    metrics_hash_t::iterator begin();
-    metrics_hash_t::iterator end();
     std::string dump();
     std::string flush();
 
--- a/include/mapnik/performance_stats.hpp
+++ b/include/mapnik/performance_stats.hpp
@@ -52,7 +52,7 @@ struct timer_metrics {
     double wall_clock_elapsed;
 };
 
-typedef std::unordered_map<std::string, timer_metrics> metrics_hash_t;
+using metrics_hash_t = std::unordered_map<std::string, timer_metrics>;
 
 class MAPNIK_DECL timer_stats : public singleton<timer_stats, CreateStatic>
 {
--- a/include/mapnik/performance_stats.hpp
+++ b/include/mapnik/performance_stats.hpp
@@ -52,11 +52,11 @@ struct timer_metrics {
     double wall_clock_elapsed;
 };
 
-using metrics_hash_t = std::unordered_map<std::string, timer_metrics>;
-
 class MAPNIK_DECL timer_stats : public singleton<timer_stats, CreateStatic>
 {
 public:
+    using metrics_hash_t = std::unordered_map<std::string, timer_metrics>;
+
     void add(std::string const& metric_name, double cpu_elapsed, double wall_clock_elapsed);
     timer_metrics get(std::string const& metric_name);
     void reset(std::string metric_name);
--- a/include/mapnik/performance_stats.hpp
+++ b/include/mapnik/performance_stats.hpp
@@ -1,0 +1,125 @@
+/*****************************************************************************
+ *
+ * This file is part of Mapnik (c++ mapping toolkit)
+ *
+ * Copyright (C) 2017 Artem Pavlenko
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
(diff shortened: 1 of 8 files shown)


include/mapnik/performance_stats.hpp
```diff
@@ -0,0 +1,94 @@
...
+using mapnik::CreateStatic;
+
+namespace mapnik {
+
+struct timer_metrics {
+    double cpu_elapsed;
+    double wall_clock_elapsed;
+};
+
+using metrics_hash_t = std::unordered_map<std::string, timer_metrics>;
```
flippmoke: move this using into the class as it doesn't need a larger scope.

src/performance_stats.cpp
```diff
@@ -0,0 +1,106 @@
...
+ *
+ *****************************************************************************/
+
+#include <mapnik/performance_stats.hpp>
+
+namespace mapnik {
+
+template class MAPNIK_DECL singleton<timer_stats, CreateStatic>;
+
+void timer_stats::add(std::string const& metric_name, double cpu_elapsed, double wall_clock_elapsed)
```
flippmoke: I am not certain that this going to be a good measure of performance because you will be measuring the time it takes to allocate memory in your `metrics_` unordered map or to simply find it. This is possibly much longer then the section of code you are attempting to measure.

src/performance_stats.cpp
```diff
@@ -0,0 +1,106 @@
...
+{
+#ifdef MAPNIK_THREADSAFE
+    std::lock_guard<std::mutex> lock(metrics_mutex_);
+#endif
+    timer_metrics& metrics = metrics_[metric_name];
+    metrics.cpu_elapsed += cpu_elapsed;
+    metrics.wall_clock_elapsed += wall_clock_elapsed;
+    metrics_[metric_name] = metrics;
```
flippmoke: You are already getting this by reference, there is no need to set it to the value again.

src/performance_stats.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include <mapnik/performance_stats.hpp>
+
+namespace mapnik {
+
+template class MAPNIK_DECL singleton<timer_stats, CreateStatic>;
+
+void timer_stats::add(std::string const& metric_name, double cpu_elapsed, double wall_clock_elapsed)
+{
+#ifdef MAPNIK_THREADSAFE
+    std::lock_guard<std::mutex> lock(metrics_mutex_);
```
flippmoke: The lock guard will also be measured as it will require time to be constructed and deconstructed. Therefore, we are possibly measuring time to acquire the mutex and release it.

src/performance_stats.cpp
```diff
@@ -0,0 +1,106 @@
...
+    if (! stopped_)
+        {
+            stop();
+        }
+}
+
+void stats_timer::stop() const
+{
+    timer::stop();
+    try
```
flippmoke: Why do we need a try catch here?

flippmoke:
This likely should be enabled and disabled by a DEFINE as it is using a global singleton and adding extra code perf testing that would not be wanted in production release.

src/performance_stats.cpp
rafatower: Good point. The time to allocate memory, lock the instance and find within the map will always be an overhead. Whether it is included in the recorded time depends on the order of parameter evaluation here:

https://github.com/CartoDB/mapnik/blob/53e17ac7479470eaa97b21c42f28c2965db8fc93/src/performance_stats.cpp#L95

I'll fix that to make sure the overhead time is not added up to the metrics. OTOH, I'll think of ways of reducing overhead to a minimum.

src/performance_stats.cpp
rafatower: I actually just copied it from the base class: https://github.com/mapnik/mapnik/blob/df0bbe404cf3952081cc512da6c1e3e2726a46ae/include/mapnik/timer.hpp#L147

I guess the original intent of this catch-all is to never make the measured code fail because of the measurements. I'm all for removing it.

src/performance_stats.cpp
rafatower: Actually the time recorded does not depend on the order of evaluation of the parameters, because of the call to `timer::stop();` before adding the metrics to the singleton `timer_stats`:

https://github.com/CartoDB/mapnik/blob/53e17ac7479470eaa97b21c42f28c2965db8fc93/src/performance_stats.cpp#L92

https://github.com/mapnik/mapnik/blob/df0bbe404cf3952081cc512da6c1e3e2726a46ae/include/mapnik/timer.hpp#L82-L109

rafatower:
Thanks a lot for the comments.

> This likely should be enabled and disabled by a DEFINE as it is using a global singleton and adding extra code perf testing that would not be wanted in production release

Actually I'd rather prefer to be able to enable/disable the metrics through configuration, at startup or even at runtime, if possible at all. I see that the overhead can be a major concern and I'll do my best to figure out a lightweight enough approach.

Any advice in that direction is more than welcomed, even some rough guidelines about what the code can and cannot do to make its way to a production build.

flippmoke:
I have a feeling that @artemp will not allow it if it is not optionally enabled by a define. It risks someone putting perf testing code in a tight loop and perhaps making perf much worse. Additionally the mutexes would cause havoc perhaps in multi-threaded environments. In general I would highly suggest against any benchmarking with multiple threads this way. It is much easier to judge multi-threaded performance at the application level rather then within parts of the code. Judging single threaded operations is much better way IMO. 

Additionally, we do a lot of benchmarking using different performance tools such as `perf` or `iprofiler` on OSX. These are often better ways of determining where there are slow downs in the codebase and are far less intrusive.

rafatower:
For benchmarking that's OK. For instance `perf` is less intrusive in the sense of not requiring changes in the source code. OTOH getting useful information from `perf` usually implies compiling with debug symbols, disabling optimizations and the like and the overhead of profiling in practice is far greater than the approach I proposed.

But the goal is to get timing metrics from a production setting.

Another possible approach is to collect and return the metrics per function call (just a few specific ones), returning them to the caller. The caller could even decide when the timing info is to be collected and returned. That would avoid singletons  and mutexes at all. Do you think that's feasible?

rafatower:
Closing this in favor of https://github.com/mapnik/mapnik/pull/3767

PR #3715, opened 2017-06-30
Geobuf.input

artemp:
geobuf.input plug-in (https://github.com/mapbox/geobuf)


src/wkb.cpp
```diff
@@ -351,6 +351,7 @@ struct wkb_reader : util::noncopyable
     {
         int num_polys = read_integer();
         mapnik::geometry::multi_polygon<double> multi_poly;
+        multi_poly.reserve(num_polys);
```
springmeyer: @artemp should this be ported to `master` and `v3.0.x` independently?

springmeyer:
Per chat:
  - [ ] This is not viable for merging until geobuf format is settled (but great to have for experimenting)
  - [ ] The `pbf.hpp` header should be replaced with https://github.com/mapbox/protozero/
  - [ ] Headers should be organized such that pure geobuf related parsing code is separate from conversion to mapnik objects (such that geobuf parsing library can be extracted standalone eventually)

src/wkb.cpp
artemp: @springmeyer ported ( 249d6cc311f90694a74b26be201342225817c752 )

PR #3716, opened 2017-07-03
Set premultiplied flag to the color font glyph bitmap

talaj:
Premultiplied flag has to be set to the glyph bitmap, otherwise [compositing is throwing an error if MAPNIK_DEBUG=True](https://github.com/mapnik/mapnik/blob/333ef9fde145f88339eaccba810305707bae9b0e/src/image_compositing.cpp#L148).

Glyph bitmaps are stored premultiplied:
http://git.savannah.gnu.org/cgit/freetype/freetype2.git/tree/ChangeLog.25?id=ca799e9be52526739691f120285a27b91fd15d68#n4906

Also simple visual test is added:
bitmap-color-font-1-256-256-1 0-agg-reference


artemp:
@talaj - looks good, cheers.

Issue #3717, opened 2017-07-04
Transparent background color is not applied with Cairo renderer


cklein05:
When rendering a map with Cairo, one of the first things done is painting the context with the map's background color. However, the background color gets not painted, if it's fully transparent. This is due to the fact, that drawing with Cairo by default uses the OVER compositing operator, which just does nothing, if the alpha of the context's current color is 0x00. [See this in the docs.](https://www.cairographics.org/FAQ/#clear_a_surface) The result is always a (fully) transparent black image, no matter what background color was set for the map.

Actually, Mapnik should switch the context's comp operator to SOURCE prior to painting the map's background in order to create the desired image (and restore the operator afterwards, of course).

I've verified this for the Cairo renderer, but it may also be an issue with the Agg renderer, which I did not check.

talaj:
I think it's not a problem in case of Agg, where [pixels are directly set to the desired value](https://github.com/mapnik/mapnik/blob/a734ff1e905d72a870bc736e2e6521fedd3e4adc/src/agg/agg_renderer.cpp#L169). I agree that Cairo should do the same. I will take a look on that.

cklein05:
Great! Thank you for fixing that.

PR #3718, opened 2017-07-04
cairo renderer: clear map background with given color as agg renderer does

talaj:
Solves https://github.com/mapnik/mapnik/issues/3717.

@cklein05


artemp:
thanks @talaj !

PR #3725, opened 2017-07-07
remove superfluous line

talaj:
I think that line should not be there. See the `if` below.


springmeyer:
Nice catch. Please feel free to merge once travis is green.

PR #3742, opened 2017-08-18
Bbox index

artemp:
This PR adds more "performant" spatial on-disk index implementation by storing and testing `bounding box` per item. This results in faster `geometries-in-a-bbox` queries but also increases `*.index` file size by factor of 2. On a balance, this increased file size justified by better performance.

There is also new `mapnik-index` option to only index features within provided bounding box:

```
-b [ --bbox ] arg           Only index features within bounding box: 
                              --bbox=minx,miny,maxx,maxy

```


springmeyer:
@artemp can you add a description to this PR that documents why it is important?

PR #3760, opened 2017-09-09
Ensure the svg_parser interface is back compatible with v3.0.x series

springmeyer:
This fixes the API breakages when #3685 was ported to v3.0.x. This allows node-mapnik master to stay compiling against v3.0.x without changes. This was useful for me to test other branches against mapnik core branches off v3.0.x.


springmeyer:
latest node-mapnik master and v3.0.x branch are updated to the new API so this is obsolete. Closing and won't merge.

PR #3474, opened 2016-07-29
Nested layers, layer level comp-op and opacity

talaj:
Enables [grouped compositing](https://github.com/mapnik/mapnik/wiki/Compositing#grouped-compositing) and improves layer organization.

Layers are rendered recursively, creating it's own buffer if `comp-op` or `opacity` properties are set. Otherwise, rendering buffer is inherited from parent layer.

All tests are passing except some SVGs with changes in grouping SVG elements per `Layer`.

Here is a simple example.

``` xml
<Map background-color="darkblue" srs="+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over">
    <Parameters>
        <Parameter name="sizes">512,256</Parameter>
        <Parameter name="bbox">-20037508.3427892476320267,1283343.6938826972618699,20037508.3427892439067364,13365151.3630704730749130</Parameter>
    </Parameters>
    <Style name="world">
        <Rule>
            <PolygonSymbolizer fill="white"/>
        </Rule>
    </Style>
    <Style name="text">
        <Rule>
            <TextSymbolizer size="120" dy="-50" halo-radius="8" face-name="DejaVu Sans Bold" fill="white" vertical-alignment="top" placement="point">WORLD</TextSymbolizer>
        </Rule>
    </Style>

    <Layer name="world" comp-op="src-over" srs="+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over">
        <StyleName>world</StyleName>
        <Datasource>
            <Parameter name="file">../../data/shp/world_merc</Parameter>
            <Parameter name="type">shape</Parameter>
        </Datasource>

        <Layer name="text" comp-op="dst-in" srs="+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs">
            <StyleName>text</StyleName>
            <Datasource>
                <Parameter name="type">csv</Parameter>
                <Parameter name="inline">
                    x, y
                    0, 0
                </Parameter>
            </Datasource>
        </Layer>
    </Layer>
</Map>
```

marker-whole-multi-polygon-512-256-1 0-agg

PR #3764, opened 2017-09-11
fix feature_style_processor::apply_to_layer() after #3474

talaj:
Fixes the buffer-size problem from https://github.com/mapnik/node-mapnik/pull/805.

I will add test coverage soon.

### DIFF
--- a/include/mapnik/feature_style_processor_impl.hpp
+++ b/include/mapnik/feature_style_processor_impl.hpp
@@ -213,8 +213,12 @@ void feature_style_processor<Processor>::apply_to_layer(layer const& lay,
 
     if (!mat.active_styles_.empty())
     {
+        p.start_layer_processing(mat.lay_, mat.layer_ext2_);
+
         render_material(mat,p);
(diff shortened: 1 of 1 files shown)


artemp:
@talaj :+1:

springmeyer:
Great, thanks for fixing @talaj - looking forward to seeing test coverage. While the negative buffer usecase in node-mapnik is not critical I presume this functionality might be critical for other rendering situations in mapnik. Can you figure out which?

talaj:
> I presume this functionality might be critical for other rendering situations in mapnik. Can you figure out which?

This bug affects methods of renderers for single layer rendering. These two methods are affected:
https://github.com/mapnik/mapnik/blob/e97759d01b32a55423db9b5df09b1888e7a31714/include/mapnik/feature_style_processor.hpp#L66-L85

What exactly doesn't work depends on render. With AGG renderer `clear-label-cache="true"` on Layer will not work and `query_extent` will not be corretly adjusted by `buffer-size` and `maximum-extent`.

PR #3767, opened 2017-09-13
[v3.0.x] Add metrics support

Algunenano:
This is an attempt to improve the initial @rafatower 's PR to introduce an internal metrics system in Mapnik (#3705).

Essentially, this adds a new class `metrics` to support a tree-like structure with metrics generated during a request.
Currently metrics are only added in `feature_style_processor.hpp` as it's the common path for grids and images but the idea is to incrementally add them in more places to help us diagnose slow paths.

Initial tests (not in full depth) hasn't shown any kind of performance impact in tile generation times (not enabled at compilation vs enabled and requested vs enabled and not requested. All 3 show equivalent times), but fow now I've set up the metrics as disabled by default. Use `ENABLE_METRICS=True` to enable them in the configure call.

As I was using node-mapnik and it wasn't compatible with current master I've done all the development and testing with the v3.0.15 tag, but the MR over master should be almost automatic (I think right now I'd complain about conflicts with some #includes). Once this is accepted MRs would come to node-mapnik and tilelive-mapnik:
https://github.com/Algunenano/node-mapnik/commits/performance_stats
https://github.com/Algunenano/tilelive-mapnik/commits/performance_stats

When the full package is in action it allow us to extract information like this though Windshaft-CartoDB:
```json
{  
   "authorizedByAPIKey":1,
   "authorize":1,
   "res":5554,
   "getTileOrGrid":5554,
   "render-png":5554,
   "Mapnik":{  
      "Time (us)":5545412,
      "Calls":1,
      "Setup":{  
         "Time (us)":407477,
         "Calls":1,
         "Datasource: Get Features":{  
            "Time (us)":407449,
            "Calls":1
         }
      },
      "Render":{  
         "Time (us)":5137082,
         "Calls":1,
         "Style":{  
            "Time (us)":5137077,
            "Calls":1,
            "features":478587
         }
      }
   },
   "tiles in metatile":1,
   "total":5556
}
```
which now allows to start getting stats about, for example, %time in database vs %time rendering depending on the number of features retrieved.

Please have a look and let me know if I should modified anything or if you prefer the PR against Master directly.

### DIFF
--- a/include/mapnik/feature_style_processor.hpp
+++ b/include/mapnik/feature_style_processor.hpp
@@ -25,11 +25,13 @@
 
 // mapnik
 #include <mapnik/box2d.hpp>
-#include <mapnik/featureset.hpp>
 #include <mapnik/config.hpp>
 #include <mapnik/feature_style_processor_context.hpp>
+#include <mapnik/featureset.hpp>
+#include <mapnik/metrics.hpp>
 
 // stl
+#include <memory>
 #include <set>
 #include <string>
 
@@ -56,6 +58,11 @@ class MAPNIK_DECL feature_style_processor
 public:
     explicit feature_style_processor(Map const& m,
                                      double scale_factor = 1.0);
+#ifdef MAPNIK_METRICS
+    explicit feature_style_processor(Map const& m,
+                                     double scale_factor,
+                                     metrics& metr);
+#endif
 
     /*!
      * \brief apply renderer to all map layers.
@@ -113,7 +120,12 @@ private:
     void render_material(layer_rendering_material const & mat, Processor & p );
 
     Map const& m_;
+
+#ifdef MAPNIK_METRICS
+public:
+    metrics metrics_ = metrics(false);
+#endif
 };
-}
+} //namespace mapnik
 
 #endif // MAPNIK_FEATURE_STYLE_PROCESSOR_HPP
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -1,0 +1,195 @@
+/*****************************************************************************
+ *
+ * This file is part of Mapnik (c++ mapping toolkit)
+ *
+ * Copyright (C) 2017 Artem Pavlenko
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ *****************************************************************************/
+
+#ifdef MAPNIK_METRICS
+
+#include <mapnik/metrics.hpp>
+
+#include <mapnik/make_unique.hpp>
+
+#include <boost/next_prior.hpp>
+#include <sstream>
+#include <utility>
+
+namespace mapnik {
+
+const std::string measurement_str[TOTAL_ENUM_SIZE] =
+{
+    "UNKNOWN",
+    "Time (us)",
+    "Value",
+    "Calls"
+};
+
+measurement::measurement(int64_t value, measurement_t type)
+    : value_(value),
+      type_(type)
+{
+}
+
+autochrono::autochrono(metrics* m, std::string name)
+    : metrics_(m),
+      name_(std::move(name))
+{
+    start_ = clock_.now();
+}
+
+autochrono::~autochrono()
+{
+    auto ns = std::chrono::duration_cast<time_units>(clock_.now() - start_);
+    metrics_->measure_add(name_, ns.count(), measurement_t::TIME_MICROSECONDS);
+}
+
+metrics::metrics(bool enabled)
+    : enabled_(enabled)
+{
+}
+
+metrics::metrics(metrics const &m, std::string prefix)
+    : enabled_(m.enabled_),
+      prefix_(m.prefix_ + prefix),
+      storage_(m.storage_)
+#ifdef MAPNIK_THREADSAFE
+     ,lock_(m.lock_)
+#endif
+{
+}
+
+/* Move constructor */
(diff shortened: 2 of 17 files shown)


pramsey:
Is there a slot in the metadata for the resolution of the output image? Rendering a 1000x1000 image should be slower than a 256x256 image (likely; certainly compressing it will be)

pramsey:
Is this meta-data all-in or for one layer? If it's all-in, having one chunk per layer plus the rolled up stats on the parent would be nice.

talaj:
@pramsey, I think you can add any string-integer pair to the metric like this:

```c++
metric.add_t("image width", 1000, measurement_t::VALUE);
```

talaj:
I used [`MAPNIK_STATS`](https://github.com/mapnik/mapnik/blob/5a87d8fef96685b6289a9151b55a9ceb947251f3/plugins/input/postgis/postgis_datasource.cpp#L103-L105) for that some time ago, but this looks definitely better. If merged, `MAPNIK_STATS` should be replaced by `MAPNIK_METRICS`.

src/metrics.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+autochrono::~autochrono()
+{
+    auto ns = std::chrono::duration_cast<time_units>(clock_.now() - start_);
+    metrics_->add_t(name_, ns.count(), measurement_t::TIME_MICROSECONDS);
+}
+
+std::unique_ptr<autochrono> metrics::time(const std::string& name)
+{
+    return std::make_unique<autochrono>(this, name);
```
talaj: To minimize overhead, I would not construct `autochrono` on the heap like this, but rather on the stack. I think this method is misleading.

include/mapnik/feature_style_processor.hpp
```diff
@@ -113,7 +115,24 @@ class MAPNIK_DECL feature_style_processor
...
     Map const& m_;
+
+#ifdef MAPNIK_METRICS
+public:
+    void enable_metrics(bool status);
+    std::shared_ptr<metrics> get_metrics();
+
+private:
+    bool metrics_enabled_ = false;
+    std::shared_ptr<metrics> metrics_ = nullptr;
```
talaj: Why `std::shared_ptr`? I think this line should be 

```c++
metrics & metrics_;
```

Lifetime of `metrics` object should be responsibility of an observer - caller of Mapnik. Except of semantics, `std::shared_ptr` is complicated thing with some overhead.

talaj: Or `metrics * metrics_;`, if it should be nullable.

Algunenano: Basically, it's a shared pointer to ease working with a metrics object in different threads without needing an strict owner.
Let's say you have 1 thread handling a query that, internally creates 5 tasks (different threads also), 2 of them being non-cancelable. If you want to cancel the query while it's running and the main thread is the only responsible (`metrics *`) then it has to wait for all other threads to stop to free it, if it holds a shared_ptr then it can remove its reference, return immediately and the memory will be freed when the last task stops.

I know right now it's possible to just go for raw pointer but if I have to choose between some overhead and the automatic memory management I'd go for the later.

talaj: OK, it might be right place for shared ownership. Overhead was not my primary concern.

Why is the same code duplicated in `grid.hpp`, `image.hpp`, `image_any.hpp` and  `image_null.hpp`?

Algunenano: I didn't want to use inheritance for such different classes with some of them already inheriting, but I could create a virtual class with the basic members (metrics and metrics enabled) and getter/setters if you prefer it.

src/metrics.cpp
Algunenano: The idea was to have a simple method to calculate time metrics. I might be wrong, but considering the current use in the feature style processor [measure_time](https://github.com/Algunenano/mapnik/blob/a24a6d50a9d6642c0cf8ef410a35860971e87367/include/mapnik/feature_style_processor_impl.hpp#L693) it should be movable all the way up and the overhead to be minimal.

Also, it's difficult to handle optional objects in the stack: 
```c++
if (metrics_enabled_)
{
     autochrono t(metrics_, name);
}
```
This would not work as it would destroy the scope of t. On the other hand you could use boost::optional which I think would have the same overhead, or handle passing nullptr to autochrono but *I think* that would slightly increase the overhead when you are not using metrics.
So I could switch it the calls to:
```c++
autochrono t(metrics_enabled_ ? metrics_ : nullptr, name)
```
What do you think?

Algunenano:
@pramsey as @talaj said you can add whatever you want to the metrics. It accepts any string using "." as separator. For example, doing:
```c++
metrics.time("Mapnik.Render.Layer0");
metrics.inc("Mapnik.Render.Layer0.Features", 1000);
metrics.time("Mapnik.Render.Layer1");
metrics.inc("Mapnik.Render.Layer1.Features", 2000);
```
Should give you something like this:
```json
{
   "Mapnik":{
      "Render":{
         "Layer0":{
            "Time (us)":12531,
            "Calls":1,
            "Features":1000
         },
         "Layer1":{
            "Time (us)":13434,
            "Calls":1,
            "Features":2000
         }
      }
   }
}
```

So as long as you use some kind of ids for layers (as calling the metric with the same name just adds the result) you can measure whatever you want.

Algunenano:
With respect `MAPNIK_STATS`: I started using it but decided to switch to a different flag as it was tied to debug and logging with clog and I wanted to be able to use metrics in release mode.
In the future I'd like to move those stats into metrics and use debug logging instead, so you get the stats through metrics and only get logs with logging=debug.

src/metrics.cpp
talaj: I think it's fine then. I was thinking why should be `metrics` dependent on `autochrono`.

include/mapnik/feature_style_processor.hpp
talaj: What about compositing?

Algunenano: Yeah, that could be a good option. I'll have a look and see if I can set up another class (or modify metrics) to make it public so everything can be accessed directly just by declaring a new member; e.g: `grid.metrics.enable()` or `this.metrics.time("metric.name")`

Algunenano:
I've updated the PR to use composition instead of a shared_ptr to the metrics object and also change the usage of it. The idea is that each class that want to enable stats has a different object but the contents can be shared (with thread safety measures). 
For example, you can start the `feature_style_processor` metrics with the metrics of the image/grid so the internals (tree and lock) are shared between them.
Thanks to this the changes to node-mapnik were simplified a lot, since you can now just enable the metrics in the image and the rest is taking care of in mapnik itself.

PR #3775, opened 2017-10-05
Fix memory leak when color font used

talaj:
The `continue` prevented `FT_Done_Glyph()` from being called.

I was thinking about using `unique_ptr` which would also ensure deallocation in case of exception. Please let me know if it is preferred.


artemp:
:+1:

PR #3782, opened 2017-10-13
Add new angled-point marker placement mode for lines

bmharper:
Details are inside the commit message. I have created a visual test for this, but will rather wait for feedback before sending a PR for that.
We use this kind of rendering for drawing symbols on top of water pipes - for example, when a pipe is actually a valve, or when a pipe is closed, etc. In this case, we always want just a single symbol drawn in the center of the polyline, and we want the symbol's angle to be aligned to the line.

New `angled-point` marker-placement mode:
angled-point

Existing `point` marker-placement mode:
point

### DIFF
--- a/include/mapnik/geom_util.hpp
+++ b/include/mapnik/geom_util.hpp
@@ -35,6 +35,12 @@
 #include <vector>
 #include <algorithm>
 
+// boost
+#pragma GCC diagnostic push
+#include <mapnik/warning_ignore.hpp>
+#include <boost/optional.hpp>
+#pragma GCC diagnostic pop
+
 namespace mapnik
 {
 template <typename T>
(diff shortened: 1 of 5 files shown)


artemp:
@bmharper - great feature, thanks for contributing! My only concern after skimming through the code is the use of pointer logic to select angled placement. Could we use `optional<double>` instead ? 
Please, share visual test so we can merge this PR asap, cheers!

bmharper:
I've modified the code to use `optional<double>`, and also sent a PR for one new visual test that accompanies this new mode: https://github.com/mapnik/test-data-visual/pull/56. Apologies for force pushing to this branch - I know that makes incremental reviews harder.
Thanks for the warm welcome!

PR #3792, opened 2017-11-09
Glibcxx fix

springmeyer:
Currently mapnik, when built with libstdc++-5-dev on linux (and the existing boost mason packages) ends up depending on these GLIBCXX symbols:

```
                 U _ZNSt11logic_errorC1EPKc@@GLIBCXX_3.4.21
                 U _ZNSt12length_errorC1EPKc@@GLIBCXX_3.4.21
                 U _ZNSt12out_of_rangeC1EPKc@@GLIBCXX_3.4.21
                 U _ZNSt13runtime_errorC1EPKc@@GLIBCXX_3.4.21
                 U _ZNSt13runtime_errorC2EPKc@@GLIBCXX_3.4.21
                 U _ZNSt16invalid_argumentC1EPKc@@GLIBCXX_3.4.21
                 U _ZSt24__throw_out_of_range_fmtPKcz@@GLIBCXX_3.4.20
                 U _ZNSt11logic_errorC1EPKc@@GLIBCXX_3.4.21
                 U _ZNSt12length_errorC1EPKc@@GLIBCXX_3.4.21
                 U _ZNSt12out_of_rangeC1EPKc@@GLIBCXX_3.4.21
                 U _ZNSt13runtime_errorC1EPKc@@GLIBCXX_3.4.21
                 U _ZNSt13runtime_errorC2EPKc@@GLIBCXX_3.4.21
                 U _ZNSt16invalid_argumentC1EPKc@@GLIBCXX_3.4.21
                 U _ZSt24__throw_out_of_range_fmtPKcz@@GLIBCXX_3.4.20
```

These symbols are not available in the libstdc++6.so versions on ubuntu trusty or amazon linux 2017.09 and older (they are only available in the default libstdc++6.so installed on ubuntu xenial and above).

In order to allow mapnik binaries to run on AWS lambda we need to drop the dependence on these symbols, which can be done by:

 - building boost libraries against libstdc++-4.9-dev instead of libstdc++-5-dev
 - building mapnik against libstdc++-4.9-dev instead of libstdc++-5-dev
 - applying a shim to avoid `throw_out_of_range_fmt@GLIBCXX_3.4.20` getting linked (to avoid needing to build against libstdc++-4.8-dev which has known bugs)

PR #3796, opened 2017-11-16
Glibcxx fix redux

springmeyer:
This finishes the work from #3792 by actually adding the workaround and adding a script that tests the symbols we'd expect are gone are actually gone - if not it fails.


springmeyer:
This looks good on travis. Merging.

PR #3809, opened 2017-11-28
Export all text_placements_ classes from mapnik.dll

bmharper:
Previously, only text_placements_dummy had a MAPNIK_DECL annotation.
This commit adds that annotation to text_placements_simple and
text_placements_list, so that they are also exposed from the Windows DLL
build.


artemp:
@bmharper - thanks!

PR #3828, opened 2018-01-11
Replaced old-style exception for Python 3 compatibility.

mikejcorey:
I'm not sure if this is a one-time orphan or a widespread convention in the codebase, but I had to switch from the older comma style to "as" style to make build.py work on Python 3.


artemp:
@mikejcorey thanks for spotting!

PR #3833, opened 2018-01-17
Fix crash in case of empty ring

talaj:
The `polygon_vertex_processor` which is responsible for constructing `mapnik::geometry::polygon` out of vertex stream was crashing when the vertex stream contained only `SEG_CLOSE` command.


artemp:
:+1:

PR #3834, opened 2018-01-17
Interior, Polylabel: cover empty polygon or exterior ring

talaj:
Interior and Polylabel algorithms are throwing an out of bounds exception or misbehave if run against empty polygon or a polygon with empty exterior ring. It is a good practice to not throw outside of Mapnik during rendering.

I've fixed these cases and changed interface of these algorithms to be similar to the interface of `mapnik::geometry::centroid()`.


artemp:
looking good :+1: !

PR #3836, opened 2018-01-18
Avoid unnecessary copy of features with "cache-features"

Algunenano:
Unless I'm missing something this is unnecessary. Could it be a copy and paste error from the `group_by` option that does need to use a cache array?


talaj:
I think that the [`cache-features`](https://github.com/mapnik/mapnik-reference/blob/3f76a593650d16d9c295549fb57a788233f98447/3.0.6/reference.json#L185-L190) option is all about copying features into memory.

Algunenano:
Closing this as it was in fact necessary. Sorry for the inconvenience!

Issue #3041, opened 2015-08-25
Offset artifacts


talaj:
I have quite a lot of polygons where offsetting produces following artifacts after https://github.com/mapnik/mapnik/pull/2904.

![polygon-offset-1-800-800-1 0-agg](https://cloud.githubusercontent.com/assets/1950911/9466741/4cbadfb4-4b37-11e5-9aad-9dc0e2891b0d.png)

Before https://github.com/mapnik/mapnik/pull/2904:
![polygon-offset-1-800-800-1 0-agg-orig](https://cloud.githubusercontent.com/assets/1950911/9466742/4cbee88e-4b37-11e5-8947-643a9905d185.png)

Complete test is in https://github.com/mapycz/test-data-visual/commit/4a0db3f8b6d4658ee23ea398804a8f208b962f1f.

artemp:
@flippmoke ^

talaj:
It also makes labeling such lines problematic as there are huge angle differences between adjacent characters due to those spikes.

flippmoke:
Nightmare on offset street. This entire algorithm needs to be revisited, however, it would be a heck of a lot easier if it was able to be done in a mapnik::geometry object rather then in the agg vertex format...

talaj:
> it would be a heck of a lot easier if it was able to be done in a mapnik::geometry object rather then in the agg vertex format

And that is problem because all the [vertex converters](https://github.com/mapnik/mapnik/blob/master/include/mapnik/vertex_converters.hpp) operates on the agg vertex format?

flippmoke:
@talaj yes, the other problem is that it needs to be quick and many of the other offset algorithm as very slow, I will probably have to think long and hard about this one.

talaj:
I can see the improvements in visual tests but I moved back to the old proven algorithm. It breaks existing maps and only solutions I see currently are workarounds. I might find some spare time to dive into the offsetting code.

Also I'm not convinced that inversion of offset direction on lines was worth the loss of compatibility.

winni159:
Hi! I'm working on this issue and I have some nice results for the polygon above and the others:
a)
![after-polygon-offset-1-800-800-1 0-agg](https://cloud.githubusercontent.com/assets/3541452/10863030/6006fe64-7fc0-11e5-9be3-62de3e250bf1.png)
b) San Marino before:
[...]

PR #3843, opened 2018-01-26
Fix filtering of vertices in offset_converter

talaj:
The `offset_converter` generates many possible candidate vertices for the offsetted path which are then filtered to get rid of ["the curls"](https://github.com/mapnik/mapnik/issues/3041).

This filtration can erroneously remove vertices with `SEG_MOVETO` or `SEG_CLOSE` commands and lead to inadvertent joining of lines or rings.

A visual test before the fix:
multipolygon-two-holes-geojson-clip-300-300-2 0-agg-reference

After the fix:
multipolygon-two-holes-geojson-clip-300-300-2 0-agg-reference2


artemp:
thanks @talaj !

PR #3844, opened 2018-02-03
Interior, polylabel: Scale precision by polygon size 

talaj:
It was a mistake to use a constant value as the precision parameter for interior and polylabel algorithms. These algorithms work on polygons in map (viewport) coordinates, thus the precision should scale with the polygon, to get consistent positions across zoom levels, for example.

This pull request fixes this by setting the precision to 1% of the polygon size.

Here is a demonstration of the problem:

polygon-placement-scale-invariant-256-256-1 0-agg-reference
polygon-placement-scale-invariant-128-128-1 0-agg-reference

After the fix:
polygon-placement-scale-invariant-256-256-1 0-agg
polygon-placement-scale-invariant-128-128-1 0-agg

cc @kocio-pl

### DIFF
--- a/src/geometry/interior.cpp
+++ b/src/geometry/interior.cpp
@@ -145,15 +145,8 @@ struct cell
 };
 
 template <class T>
-boost::optional<point<T>> polylabel(polygon<T> const& polygon, T precision = 1)
+point<T> polylabel(polygon<T> const& polygon, box2d<T> const& bbox , T precision = 1)
 {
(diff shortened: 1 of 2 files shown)


talaj:
Travis fails due to unrelated problem which is fixed in https://github.com/mapnik/mapnik/pull/3845.

talaj:
Thanks @artemp . I will also make a pull request against v3.0.x.

artemp:
👍

Issue #3854, opened 2018-02-13
grid_vertex_converter - illegal memory access (empty pod_array) 


artemp:
```DEBUG = True DEBUG_UNDEFINED = True DEBUG_SANITIZE = True``` reveals 

```
* thread #1, name = 'run', stop reason = signal SIGILL: illegal instruction operand
    frame #0: run`agg::pod_array<unsigned char*>::operator[](this=0x00007fffffff7168, i=0) at agg_array.h:144
   141 	
   142 	        unsigned size() const { return m_size; }
   143 	        const T& operator [] (unsigned i) const { return m_array[i]; }
-> 144 	              T& operator [] (unsigned i)       { return m_array[i]; }
   145 	        const T& at(unsigned i) const           { return m_array[i]; }
   146 	              T& at(unsigned i)                 { return m_array[i]; }
   147 	        T  value_at(unsigned i) const           { return m_array[i]; }
```

**Comment by talaj** [comment]
@artemp - I see what is the problem, will send a fix.

**Comment by artemp** [comment]
@talaj great thanks!

I have a fix locally which is a bit long winded, maybe you can come up with something better : 

```
diff --git a/include/mapnik/grid_vertex_converter.hpp b/include/mapnik/grid_vertex_converter.hpp
index 7d75d570d..dbf4c8c8f 100644
--- a/include/mapnik/grid_vertex_converter.hpp
+++ b/include/mapnik/grid_vertex_converter.hpp
@@ -154,7 +154,7 @@ private:
     }
 
     grid_vertex_converter(polygon<T> const& poly, T dx, T dy, double scale_factor, box2d<T> const& envelope)
-        : hit_bitmap_scale_(get_hit_bitmap_scale(envelope)),
+        : hit_bitmap_scale_(get_hit_bitmap_scale(envelope) > 0 ? get_hit_bitmap_scale(envelope) : 1.0),
           dx_(dx * hit_bitmap_scale_),
           dy_(dy * hit_bitmap_scale_),
           vt_(envelope.valid() ? (envelope.width() * hit_bitmap_scale_) : 0,
@@ -183,27 +183,28 @@ private:
     // The polygon is rendered to a bitmap for fast hit-testing.
     image_gray8 create_hit_bitmap(polygon<T> const& poly) const
     {
-        polygon_vertex_adapter<T> va(poly);
-        view_transform_agg_adapter vta{ vt_ };
[...]

PR #3855, opened 2018-02-13
grid_vertex_converter: Fix handling of empty polygon

talaj:
Addresses https://github.com/mapnik/mapnik/issues/3854.

Fixes edge case in `agg::row_ptr_cache` when the number of rows is zero.

Add tests for `envelope.valid()` to prevent counting with NaN.


artemp:
thanks @talaj !

PR #3866, opened 2018-02-28
Remove sanitizer flags to configure step

springmeyer:
I think it is ideal to enable the sanitizers manually, because the flags have changed across different compilers. So, I recommend something like this

```
./configure CUSTOM_CXXFLAGS="-fsanitize=address" CUSTOM_LDFLAGS="-fsanitize=address"
```

As far as additional flags, see https://github.com/mapbox/cpp/issues/41


talaj:
One should not forget that `./configure CUSTOM_CXXFLAGS="..."` overrides `CUSTOM_CXXFLAGS` from `bootstrap.sh`:

https://github.com/mapnik/mapnik/blob/abce91a1808afd95603e570d72482d032648ca5d/bootstrap.sh#L88

artemp:
@talaj good point, yes.

PR #3871, opened 2018-03-14
static libc++ (+sanitizer support)

artemp:


SConstruct
```diff
@@ -1297,7 +1297,8 @@ if not preconfigured:
     env['QUERIED_PROJ_LIB'] = None
     env['QUERIED_ICU_DATA'] = None
     env['QUERIED_GDAL_DATA'] = None
-
+    #'LINKCOM': '$LINK -o $TARGET $LINKFLAGS $__RPATH $SOURCES $_LIBDIRFLAGS $_LIBFLAGS'
+    env['LINKCOM'] = '$LINK -o $TARGET $SOURCES $LINKFLAGS $__RPATH $_LIBDIRFLAGS $_LIBFLAGS'
```
artemp: Re-ordering at linking stage required when using statically linked `libc++` /cc @springmeyer

demo/c++/build.py
```diff
@@ -36,14 +36,15 @@
 demo_env['CXXFLAGS'] = copy(env['LIBMAPNIK_CXXFLAGS'])
 demo_env.Append(CPPDEFINES = env['LIBMAPNIK_DEFINES'])
 
+demo_env['LINKFLAGS'] = '-l%s ' % env['MAPNIK_NAME'] +  demo_env['LINKFLAGS']
```
artemp: Ensure `-lmapnik` appears before `libc++.a` so all std:: symbols resolved. 
NOTE: @springmeyer mentioned that [lld](https://lld.llvm.org/) can resolve symbols in any order (needs checking to see if this is a future solution)

include/mapnik/csv/csv_grammar_x3_def.hpp
```diff
@@ -38,9 +38,9 @@ using x3::lit;
 using x3::lexeme;
 using ascii::char_;
 
-struct csv_unesc_chars_ : x3::symbols<char>
```
artemp: ha, this got overwritten wrongly ^ , will fix right away.

include/mapnik/json/unicode_string_grammar_x3.hpp
```diff
@@ -34,14 +34,12 @@ namespace mapnik { namespace json { namespace grammar {
...
 using escaped_unicode_type = x3::rule<class escaped_unicode_tag, std::string>;
-
 BOOST_SPIRIT_DECLARE(unicode_string_grammar_type);
 BOOST_SPIRIT_DECLARE(escaped_unicode_type);
-
 }
 
 grammar::unicode_string_grammar_type const& unicode_string_grammar();
-grammar::escaped_unicode_type const& escaped_unicode_grammar();
+__attribute__((visibility("default"))) grammar::escaped_unicode_type const& escaped_unicode_grammar();
```
artemp: setting visibility to "default" to make `escaped_unicode_grammar()` visible during linking stage.  I'm not sure if we need to expose more to make `libmapnik-json` more usable, lets see.

/cc @springmeyer

artemp: I'm seeing that merge went wrong and some changes got overwritten, hmm.. on the case
/cc @springmeyer

artemp: I'm having second thoughts on re-arranging grammars. The issue with duplicate symbols only manifest itself when linking plug-ins statically. Adding `escaped_unicode_grammar()` helps but as a result requires linking to `mapnik-json.a` in rundemo and `mapnik-render`. I might revert this change all together. 
/cc @springmeyer

src/geometry/interior.cpp
```diff
@@ -85,15 +85,19 @@ auto point_to_polygon_dist(const point<T>& point, const polygon<T>& polygon)
 
     for (const auto& ring : polygon)
     {
-        for (std::size_t i = 0, len = ring.size(), j = len - 1; i < len; j = i++)
+        std::size_t length = ring.size();
+        if (length > 1)
```
artemp: without this check, `len - 1` overflows when `len == 0`

src/plugin.cpp
```diff
@@ -59,27 +60,27 @@ PluginInfo::PluginInfo(std::string const& filename,
...
                 callable_returning_void init_once = reinterpret_cast<callable_returning_void>(dlsym(module_->dl, "on_plugin_load"));
-                if (init_once) {
-                    init_once();
-                }
+                if (init_once)init_once();
           }
-  #else
+          else
+          {
+              throw std::runtime_error(dlerror());
```
artemp: New behaviour - throw at runtime with detailed message from `dlerror` if `dlopen` fails. 
/cc @springmeyer

src/geometry/interior.cpp
talaj: I don't think this is a problem. That overflown `len - 1` value is not used when the ring is empty.

artemp: @talaj - yes, true, but sanitizer wasn't happy so I added the check. I'm off on vacation right now, feel free to revert or work-around differently.

talaj: I think it's fine, thanks for the info.

PR #3872, opened 2018-03-20
GDAL Driver Overview Fix and Memory Reduction

### LINKED ISSUE #3822: GDAL: Overviews no longer work after "Raster Overzoom Quest"
To fetch a lower resolution image from an overview, `RasterIO()` has to be called with appropriate parameters indicating lower resolution request, such that this condition is met:

https://github.com/OSGeo/gdal/blob/b2d44355766dce069a027777f6bbd5c31bd98c56/gdal/frmts/gtiff/geotiff.cpp#L4114

This has been broken in https://github.com/mapnik/mapnik/pull/3668, where image buffer size and  requested region size are made equal:

https://github.com/mapnik/mapnik/blob/43d7278352e95b2e7a2371f2b4d56dfd95ab566a/plugins/input/gdal/gdal_featureset.cpp#L218-L223

Initially reported in https://github.com/mapnik/python-mapnik/issues/175.

flippmoke:
A fix for two distinct issues associated with gdal featuresets, the first is overviews were not properly being utilized based on the resolution of the final image requested. The second is that allocation of far too much memory could be possible in GDAL to attempt to do resampling internally in mapnik. This now has a hard cap so that we allocate less memory in these situations but are still able to resample internally in mapnik.

Should resolve #3822 

Artem is out for a bit could you please review @talaj

### DIFF
--- a/plugins/input/gdal/gdal_featureset.cpp
+++ b/plugins/input/gdal/gdal_featureset.cpp
@@ -155,6 +155,8 @@ feature_ptr gdal_featureset::get_feature(mapnik::query const& q)
     //size of resized output pixel in source image domain
     double margin_x = 1.0 / (std::fabs(dx_) * std::get<0>(q.resolution()));
     double margin_y = 1.0 / (std::fabs(dy_) * std::get<1>(q.resolution()));
+    MAPNIK_LOG_DEBUG(gdal) << "gdal_featureset: margin_x=" << margin_x;
+    MAPNIK_LOG_DEBUG(gdal) << "gdal_featureset: margin_y=" << margin_y;
     if (margin_x < 1)
     {
         margin_x = 1.0;
@@ -169,6 +171,10 @@ feature_ptr gdal_featureset::get_feature(mapnik::query const& q)
     int y_off = rint(box.miny() - margin_y);
     int end_x = rint(box.maxx() + margin_x);
     int end_y = rint(box.maxy() + margin_y);
+    MAPNIK_LOG_DEBUG(gdal) << "gdal_featureset: x_off=" << x_off;
+    MAPNIK_LOG_DEBUG(gdal) << "gdal_featureset: y_off=" << y_off;
+    MAPNIK_LOG_DEBUG(gdal) << "gdal_featureset: end_x=" << end_x;
+    MAPNIK_LOG_DEBUG(gdal) << "gdal_featureset: end_y=" << end_y;
 
     //clip to available data
     if (x_off < 0)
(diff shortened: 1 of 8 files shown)


flippmoke:
I believe the only test where I am concerned about overviews still working properly is a situation where an overview is applied on a geotiff with different projections from the resulting image. These situations seem a bit more complex. I have not added a test such as this to our tests here but we might consider adding one.

flippmoke:
Also it looks like we need an overview test on a single band tiff where we specify the band specifically (as this is a slightly different code path currently).

talaj:
Thank you @flippmoke. I will take a look.

talaj:
> Also it looks like we need an overview test on a single band tiff where we specify the band specifically (as this is a slightly different code path currently).

I've prepared such test and will make a pull request soon. The test is not passing as expected yet, but I think the problem is unrelated to this pull request. I have to investigate.

I like the approach of getting the raster feature in native resolution, leaving the responsibility for resampling to Mapnik, where you can control interpolation filter. That was my main difficulty with original implementation and the hotfix I made.

plugins/input/gdal/gdal_featureset.cpp
```diff
@@ -187,21 +193,139 @@ feature_ptr gdal_featureset::get_feature(mapnik::query const& q)
...
+            im_offset_y = adjusted_ratio * im_offset_y;
+            im_height = adjusted_height;
+        }
+    }
+    
+    // Maximum memory limitation for image will be simply based on the maximum
+    // area we allow for an image. The true memory footprint therefore will vary based
+    // on the type of imagery that exists. 
+    // max_im_area based on 50 mb limit for RGBA
+    constexpr int64_t max_im_area = (50*1024*1024) / 4;
```
talaj: Would be nice to make this value a datasource parameter, at least to allow to make a test case for this code path without need to store a large raster file in the test data.

flippmoke: @talaj we could definitely do that, perhaps with a default value here.

talaj:
Here is the promised test with single band raster: https://github.com/mapnik/test-data-visual/pull/66

Unfortunately this is resulting image:

gdal-overview-gray8-256-256-1 0-agg

The problem is unrelated to this pull request and can be fixed by this change:

```diff
--- a/include/mapnik/renderer_common/process_raster_symbolizer.hpp
+++ b/include/mapnik/renderer_common/process_raster_symbolizer.hpp
@@ -266,7 +266,9 @@ void render_raster_symbolizer(raster_symbolizer const& sym,
                 bool scale = (std::fabs(image_ratio_x - 1.0) > eps) ||
                      (std::fabs(image_ratio_y - 1.0) > eps) ||
                      (std::abs(start_x) > eps) ||
-                     (std::abs(start_y) > eps);
+                     (std::abs(start_y) > eps) ||
+                     (raster_width != source->data_.width()) ||
+                     (raster_height != source->data_.height());
                 detail::image_dispatcher<F> dispatcher(start_x, start_y, raster_width, raster_height,
                                                             image_ratio_x, image_ratio_y,
                                                             offset_x, offset_y,
```

talaj:
Yeah, if I understand it correctly there should be the same logic concerning raster size calculation as in the `gdal_featureset`. It is not currently, the `margin_x` and `margin_y` are not taken into account. Also rounding is done differently.

flippmoke:
I am also wondering if the rounding here isn't correct. https://github.com/mapnik/mapnik/blob/9bcd733def2059e6755d498957c2cc183ff43c8e/include/mapnik/renderer_common/process_raster_symbolizer.hpp#L226-L229

talaj:
There is wrong assumption that no scaling means source raster (output of GDAL datasource) has the same resolution as destination raster (a map or it's part). But source raster has additional margin pixels so the size is bigger even the scale is the same.

flippmoke:
@talaj sorry I haven't been able to finish up my work here, but I wanted to get your thoughts on what else needs to be done besides just merging this in

talaj:
I can only think of test for the `if (im_area > max_im_area)` now. I can prepare the test.

flippmoke:
What about possibly making it datasource parameter?

talaj:
I think the datasource parameter is necessary if we want to make a test case out of existing raster files from test data.

You mentioned concerns about situation when projection differs between raster and resulting map. I think this is not directly related to overviews. There is no re-projection going on in the GDAL datasource, if I'm correct.

flippmoke:
There is not any reprojection currently done through GDAL, all of it is done at later steps inside mapnik. At least that is my understanding :joy: . I will attempt to make this a datasource parameter soon to make testing possible.

flippmoke:
Attempted to merge in raster fixes and still broken 😒 -- clearly doing something wrong here still. 
Should be:
image
but instead is:
image

flippmoke:
Disregard! I simply forgot to recompile locally 👍

PR #3875, opened 2018-03-23
Raster colorizer: check image bounds

talaj:
I think this is correct fix for https://github.com/mapnik/mapnik/pull/3872#issuecomment-375324545.

cc @flippmoke

### DIFF
--- a/src/raster_colorizer.cpp
+++ b/src/raster_colorizer.cpp
@@ -129,20 +129,25 @@ void raster_colorizer::colorize(image_rgba8 & out, T const& in,
 {
     using image_type = T;
     using pixel_type = typename image_type::pixel_type;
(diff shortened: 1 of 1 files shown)


flippmoke:
@talaj do we want to merge this into `gdal_overview_fix_v3.0.x` and `gdal_overview_fix` branches first?

talaj:
@flippmoke Not necessarily, but if it's more convenient for you, I can add the commit to those branches and close this pull request.

flippmoke:
@talaj I am just wanting to make sure that we have a valid test for this prior. Do we hit this normally with out the changes in gdal_overview? I haven't had time to fully review the code around this change yet, but doing it right now.

talaj:
@flippmoke There is no test which cover this in master currently. I will prepare one.

PR #3876, opened 2018-03-26
Fix build issue with MAPNIK_THREADSAFE disabled

Algunenano:
This is just a couple of necessary includes so it builds when `MAPNIK_THREADSAFE` is disabled. Required by `std::once_flag`.

Let me know if you want a PR for the 3.x branch too (should be portable by straight cherry-pick).


springmeyer:
👍 Thanks for catching this.

PR #3890, opened 2018-04-27
Improve the allow_overlap optimization when loading the map

Algunenano:
This PR does 2 things:
* Use the optimization in `load_map_string`, which is used for example when loading a map through a xml string with `node-mapnik`.
* Use the optimization also for point and text symbolizer.

### DIFF
--- a/src/load_map.cpp
+++ b/src/load_map.cpp
@@ -149,51 +149,69 @@ private:
     std::string xml_base_path_;
 };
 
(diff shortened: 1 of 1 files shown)


artemp:
@Algunenano - looks good, thanks!

lightmare:
This is a potentionally breaking change. If you modify styles/layers after `load_map`, may result in unexpected overlaps. Also presence of additional attributes in `save_map` output may be unexpected.

PR #3891, opened 2018-04-30
don't error oddly if compiler is not found

springmeyer:
Before this change if `c++` did not exist on the system you'd see this confusing error:

```
./configure 
Inheriting from mapnik-settings.env
scons: Reading SConscript files ...

Welcome to Mapnik...

Configuring build environment...
SCons CONFIG found: 'config.py', variables will be inherited...
Configuring on Linux in *release mode*...
AttributeError: 'NoneType' object has no attribute 'decode':
  File "/root/mapnik/SConstruct", line 1278:
    color_print(5, "CXX %s" % cxx_version.decode("utf8"))

real	0m0.393s
user	0m0.294s
sys	0m0.043s
```


artemp:
:+1:

Issue #2516, opened 2014-10-10
immutable map object during rendering


springmeyer:
Maps and the style objects they hold should not be mutated during rendering in order to be safe to share between rendering threads. Style immutability was working in Mapnik 2.3.x but appears to have regressed in 3.x at some point because I'm seeing crashes (https://github.com/mapnik/mapnik/issues/2504)  in text properties.

Note: we are working toward the same promise for datasources too (#2521), such that any and all structures held by a map could be shared between threads if you wish.

Back in the early days of 2.3.x we fixed one instance of this problem (https://github.com/mapnik/mapnik/pull/1723) which unlocked being able to share Maps across threads when a custom thread-safe  datasource is used (e.g. vector tiles) (#1737).

This ticket stands to track ensuring no state is altered on a map during rendering.
- [x] - Needs review: [this shared_ptr is copied](https://github.com/mapnik/mapnik/blob/536b23bfde5b45a6450a106144d9131afcf2d087/src/text/text_properties.cpp#L101) and not the underlying `font_feature_settings` instance effectively mutating state on the map. One solution would be to pass `font_feature_settings` around by value so it is properly copied. Work in progress at https://github.com/mapnik/mapnik/tree/ff_settings_by_value. Done in #2522
- [x] - Needs review: is it safe to take a pointer (or after 536b23bfde5 a ref) to glyph_info or should we pass by value? Seems fine as long as the original owner (`text_item`) stays in scope (which it should).

PR #3900, opened 2018-05-22
[WIP] Marker symbolizer: Cache marker images for faster rendering

Algunenano:
This adds an image cache to speed up the marker symbolizer.

What is does is, if the marker symbolizer is cacheable (for now it just checks that it doesn't use expressions), it caches the associated `svg::path_attributes`. Then the `svg::path_attributes` contain a vector of images associated with that marker (the same marker rendered with `sampling_rate` (8) precision).

This PR comes from the [Carto fork](https://github.com/CartoDB/mapnik). Apart from some minor changes, there we also have metrics and a flag to enable or disable it per map (currently it's always on).

Main drawbacks I see:
* Breaks `const` since the cached is held internally (in the attributes and the symbolizer). We tried having a global cache (as with the marker_cache singleton) but it required to be pretty big (thousands of images) and to use synchronization (mutexes for multithread support) and it'd still work poorly in high contention scenarios. Also, having it associated with the symbolizer means that when the it disappears so do the associated images (no need to hack around a LRU policy).
* Some subpixel changes (check failed tests).
* The `symbolizer` variant grows in size. I haven't seen anywhere something like an array of thousands of symbolizers, so it shouldn't have any impact.

Main benefits:
* Faster rendering. Here are some measurements from our fork (Response time for the complete stack) requesting tiles with N points with the same style: [cache_off_vs_cache_on.pdf](https://github.com/mapnik/mapnik/files/2026900/cache_off_vs_cache_on.pdf)

Please let me know of any comments or improvements that could be made to have this accepted.

### DIFF
--- a/src/agg/process_markers_symbolizer.cpp
+++ b/src/agg/process_markers_symbolizer.cpp
@@ -58,7 +58,7 @@ struct agg_markers_renderer_context : markers_renderer_context
     using vertex_source_type = typename SvgRenderer::vertex_source_type;
     using pixfmt_type = typename renderer_base::pixfmt_type;
 
-    agg_markers_renderer_context(symbolizer_base const& sym,
+    agg_markers_renderer_context(markers_symbolizer const& sym,
                                  feature_impl const& feature,
                                  attributes const& vars,
                                  BufferType & buf,
@@ -66,7 +66,8 @@ struct agg_markers_renderer_context : markers_renderer_context
       : buf_(buf),
         pixf_(buf_),
         renb_(pixf_),
-        ras_(ras)
+        ras_(ras),
+        sym_cacheable(sym.cacheable == markers_symbolizer::cache_status::CACHEABLE)
     {
         auto comp_op = get<composite_mode_e, keys::comp_op>(sym, feature, vars);
         pixf_.comp_op(static_cast<agg::comp_op_e>(comp_op));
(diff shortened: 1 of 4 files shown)


lightmare:
I don't think this can be accepted as is. One problem I see is casting away `const` from a function argument. That's breach of contract.
(aside note: drawing from STL convention of naming things like `std::cref, std::cbegin, ...`, I wouldn't name const-deprived variables `cattr, csym`; rather `mattr, msym` for mutable)

More important issue, though, is that `Map` should be safe to use concurrently. Maybe we're not there, yet, but that's the goal (#2516). So a cache has to be either global or in `mutable` members and used in a thread-safe manner.

Algunenano:
> I don't think this can be accepted as is. One problem I see is casting away const from a function argument. That's breach of contract.
(aside note: drawing from STL convention of naming things like std::cref, std::cbegin, ..., I wouldn't name const-deprived variables cattr, csym; rather mattr, msym for mutable)

Any other ideas on how to handle this? I did some investigation around this and I couldn't find a good solution, for example changing the interface to non const means hundreds of changes.

> More important issue, though, is that Map should be safe to use concurrently. Maybe we're not there, yet, but that's the goal (#2516). So a cache has to be either global or in mutable members and used in a thread-safe manner.

We started with a full thread safe implementation and optimized from there. The only point with issues (I'd have to check the standard) is the resize call; for anything else the worst it can happen is to have some value calculated twice and discarded once (handled by the atomics of shared_ptr afaik).

We also tried the global cache and it was both more complex and performed way worst (this one did require mutexes).

lightmare:
> Any other ideas on how to handle this? I did some investigation around this and I couldn't find a good solution, for example changing the interface to non const means hundreds of changes.

- `path_attributes`
  ```c++
  mutable std::unique_ptr<std::array<cache_line, sampling_rate * sampling_rate>>
      cached_images;
  ```
  `mutable` because you want to modify it inside const object
  `std::unique_ptr` because `vector` is usually 4x the size of pointer
  `std::array` because the size is fixed, let's avoid the `vector` storage allocation

- `markers_symbolizer`
  make the added cache members `mutable`


Now the harder part.

> (handled by the atomics of shared_ptr afaik)

Unfortunately not. Only the "control block", that's the dynamically allocated refcount, is guaranteed atomic. That means multiple threads copying a `shared_ptr` is fine -- they each atomically increase the refcount, and don't modify the actual stored pointer. But writing (that is, assignment or swap) to a `shared_ptr` while another thread also writes or reads is undefined behavior as far as the C++ standard goes (https://en.cppreference.com/w/cpp/memory/shared_ptr) (and the same in Boost https://www.boost.org/doc/libs/1_65_0/libs/smart_ptr/doc/html/smart_ptr.html#shared_ptr_thread_safety)

You can picture `shared_ptr<T>` as a smarter `std::pair<T*, std::atomic<size_t>*>`. Don't quote me on this, it's just a simplification meant to illustrate that only the `refcount` is "smart". If you want to assign to such pair, i.e. modify its `.first` and `.second` members, you must do it under lock. Same for `shared_ptr`.

src/agg/process_markers_symbolizer.cpp
```diff
@@ -78,9 +79,144 @@ struct agg_markers_renderer_context : markers_renderer_context
...
+
+                    agg::rendering_buffer buf(fill_img->bytes(), fill_img->width(), fill_img->height(), fill_img->row_size());
+                    pixfmt_type pixf(buf);
+                    renderer_base renb(pixf);
+
+                    auto cattr_copy = cattr;
+                    cattr_copy.stroke_flag = false;
+                    cattr_copy.stroke_gradient.set_gradient_type(NO_GRADIENT);
+                    svg_attribute_type attrs_copy;
+                    attrs_copy.push_back(cattr_copy);
```
lightmare:
```c++
attrs_copy.push_back(attr);
auto& attr_copy = attrs_copy.back();
// and then do the changes
// also, the copy constructors shouldn't copy `cached_images` imo

File: src/agg/process_markers_symbolizer.cpp [hunk 2]
```diff
@@ -78,9 +79,144 @@ struct agg_markers_renderer_context : markers_renderer_context
...
+                                         marker_tr_copy, 1.0, params.snap_to_pixels);
+
+                    if (std::all_of(stroke_img->begin(), stroke_img->end(), [](uint32_t val) { return val == 0; }))
+                    {
+                        stroke_img.reset(); // optimization, can ignore image
+                    }
+                }
+
+                // Update cache with the new images
+                cl.set = true;
```
lightmare: this should come after the pointers are set and be atomic

Algunenano:
>  But writing (that is, assignment or swap) to a shared_ptr while another thread also writes or reads is undefined behavior as far as the C++ standard goes (https://en.cppreference.com/w/cpp/memory/shared_ptr) (and the same in Boost https://www.boost.org/doc/libs/1_65_0/libs/smart_ptr/doc/html/smart_ptr.html#shared_ptr_thread_safety)

Thanks, I didn't know data races were undefined behaviour; I was assuming that if 2 threads were writing the value it was undefined which one would end up stored (which is fine in this case) but one of the 2 (N) would. I'll have a look around and check was better performance wise, atomics, mutex per cache line or mutex per attributes.

lightmare:
I'd start with 1 global mutex for the cached images array allocation, and 1 global mutex + condition variable for the rendering, plus atomic flags `cl.set` and `cl.rendering`. That'll eliminate the possible duplication of work, and when only access to the flags is locked, shouldn't cause too much contention. Something like this:
```c++
if (!cl.set) {
    std::unique_lock<std::mutex> lk(global_mtx_render());

    while (cl.rendering) { // some other thread is rendering this `cl`
        global_cv_render().wait(lk); // wait on condition variable
    }

    if (!cl.set) {
        cl.rendering = true;
        lk.unlock(); // unlock the mutex; this allows rendering
                     // different `cl` in parallel
        try {
            // TODO render now, set image ptrs
        }
        catch (...) {
            // if one thread failed to render, other threads with
            // the same input are likely to fail as well;
            // just log warning and keep going, pretend empty images are ok
        }
        cl.set = true;
        cl.rendering = false;
        global_cv_render().notify_all();
    }
}
```

Algunenano:
I'm closing this for now. No point on keeping it open until I'm able to work on it again.

PR #3909, opened 2018-06-06
Improve padding calculation

Algunenano:
Improve how the extra padding is calculated when clipping polygons and lines. This gets rid of the magical `padding *= std::fabs(offset) * 1.2;` but, since the offset converter relies in another magical number (`threshold_`) it has to reference that one to properly calculate the padding.

This fixes 2 main issues:
* An offseted tile border might be shown in the image under certain conditions.
* The offset of a polygon vertex might be omitted from the render if the distance from the internal vertex to the clipped size is too close (under the threshold limit).

These 2 examples show the changes (before and after): 
image.

I don't see any noticeable performance impact (it varies depending if the formula increases or reduces the clip size): 
[pre_offset_fix_vs_post_offset_fix.pdf](https://github.com/mapnik/mapnik/files/2077317/pre_offset_fix_vs_post_offset_fix.pdf)

PR with the visual changes and a couple of extra tests: https://github.com/mapnik/test-data-visual/pull/68

### DIFF
--- a/src/cairo/process_line_pattern_symbolizer.cpp
+++ b/src/cairo/process_line_pattern_symbolizer.cpp
@@ -133,13 +133,11 @@ void cairo_renderer<T>::process(line_pattern_symbolizer const& sym,
     box2d<double> clipping_extent = common_.query_extent_;
     if (clip)
     {
-        double padding = (double)(common_.query_extent_.width()/common_.width_);
-        double half_stroke = width/2.0;
-        if (half_stroke > 1)
-            padding *= half_stroke;
-        if (std::fabs(offset) > 0)
-            padding *= std::fabs(offset) * 1.2;
-        padding *= common_.scale_factor_;
+        double pad_per_pixel = static_cast<double>(common_.query_extent_.width()/common_.width_);
+        double pixels = std::ceil(std::max(width / 2.0 + std::fabs(offset),
+                                          (std::fabs(offset) * offset_converter_default_threshold)));
+        double padding = pad_per_pixel * pixels * common_.scale_factor_;
+
         clipping_extent.pad(padding);
     }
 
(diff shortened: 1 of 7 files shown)


artemp:
Thanks @Algunenano !

Algunenano:
I've updated the visual data and the tests are passing but 2 of the builds failed due to network issues. If needed they can be relaunched manually by someone with the necessary rights.

Algunenano:
I see that it was merged manually. Thanks!

lightmare:
Does anyone have an idea what's the purpose of those `(double)(extent_.width / width)` casts? I mean, if `extent_` is always `box2d<double>`, then the width already is a `double`, and if it's not, then I'd expect the cast on just the numerator to force floating-point division, not on the result.

Algunenano:
Yep, you are right. I didn't think much about it and just kept the cast as it was.

src/cairo/process_line_pattern_symbolizer.cpp
```diff
@@ -133,13 +133,11 @@ void cairo_renderer<T>::process(line_pattern_symbolizer const& sym,
...
     if (clip)
     {
-        double padding = (double)(common_.query_extent_.width()/common_.width_);
-        double half_stroke = width/2.0;
-        if (half_stroke > 1)
-            padding *= half_stroke;
-        if (std::fabs(offset) > 0)
-            padding *= std::fabs(offset) * 1.2;
-        padding *= common_.scale_factor_;
+        double pad_per_pixel = static_cast<double>(common_.query_extent_.width()/common_.width_);
```
artemp: the cast is redundant (thanks  digging @lightmare). Just looking at identical code blocks - how about a helper function to calculate `padding` ?

PR #3912, opened 2018-06-11
GDAL: Fixes of overviews

talaj:
Fixes issues with GDAL rasters reported in https://github.com/mapnik/python-mapnik/issues/186. It's basically a follow-up to https://github.com/mapnik/mapnik/pull/3873/.

Following issues should be fixed:
* Calculation of ideal overview resolution was wrong, causing rendering from lowest available resolution.
* Raster extent size was being shrunk by rounding, causing blank space between tiles.
* `filter-factor` hasn't been taken into account, causing blurry results.

Here is an example from visual tests before and after the fix:

tiff-reprojection-2-tiled-1024-1024-4x4-1 0-agg

tiff-reprojection-2-tiled-1024-1024-4x4-1 0-agg-reference

### DIFF
--- a/plugins/input/gdal/gdal_featureset.cpp
+++ b/plugins/input/gdal/gdal_featureset.cpp
@@ -124,6 +124,33 @@ feature_ptr gdal_featureset::next()
     return feature_ptr();
 }
 
+void gdal_featureset::find_best_overview(int bandNumber,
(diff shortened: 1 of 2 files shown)


flippmoke:
Thanks @talaj for your continued work here!

PR #3913, opened 2018-06-12
Add test to assert correct value types

springmeyer:
We were not testing the types coming out of the geojson datasource. This adds tests. In particular I was curious if the boolean type was actually parsed as a `mapnik::value_bool`.


lightmare:
The same-type check could be added to the macro `REQUIRE_ATTRIBUTES` that already checks the values from list. This has the benefit of immediately adding this test to topojson and csv as well with no code duplication. I've found one wrong test in csv this way :) Will make a PR for master soon...

springmeyer:
#3925 handled this

PR #3924, opened 2018-06-30
Changelog autolinks

lightmare:
I got tired of manually looking up issues and commits mentioned in changelog, so I wrote a little script to automate turning them into proper links (GitHub doesn't do this automatically when viewing documents in repository).

I had to un-indent the lists in recent entries, not because I didn't like that they weren't line-wrapping, but because links (or any other markdown) are not interpreted in code blocks.

The script should be idempotent, i.e. it shouldn't mess up anything when applied repeatedly on the same document.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,33 +10,33 @@ For a complete change history, see the git log.
 
 Released: March 06, 2018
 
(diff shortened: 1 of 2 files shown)


lightmare:
rebased, missed some linkable hashes

PR #3752, opened 2017-09-04
Backport Python test of bbox reprojection

talaj:
This test more belongs here than to the bindings.

Originaly as Python test from https://github.com/mapnik/mapnik/pull/2657.

Issue #2438, opened 2014-09-25
tiff reprojection differences on osx vs linux


springmeyer:
for `tiff-reprojection-2` the expected image looks like:

![tiff-reprojection-2-250-250-1 0-agg-reference](https://cloud.githubusercontent.com/assets/20300/4398113/7b7f3146-444f-11e4-8fb8-88d5191c3af2.png)

While on ubuntu precise (with mapnik master built with mapnik-packaging - which should include the same gdal version) I'm seeing:

![tiff-reprojection-2-250-250-1 0-agg](https://cloud.githubusercontent.com/assets/20300/4398115/8fa75e00-444f-11e4-8e3a-927dc988454e.png)

Which is the cause of these test failures:

```
Visual rendering: 4 failed / 1502 passed
1) 1821 different pixels:
    /tmp/mapnik-visual-images/tiff-reprojection-2-250-250-1.0-agg.png (actual)
    tests/visual_tests/images/tiff-reprojection-2-250-250-1.0-agg-reference.png (expected)
2) 1821 different pixels:
    /tmp/mapnik-visual-images/tiff-reprojection-2-250-250-1.0-cairo.png (actual)
    tests/visual_tests/images/tiff-reprojection-2-250-250-1.0-cairo-reference.png (expected)
3) 1821 different pixels:
    /tmp/mapnik-visual-images/tiff-reprojection-2-250-250-2.0-agg.png (actual)
    tests/visual_tests/images/tiff-reprojection-2-250-250-2.0-agg-reference.png (expected)
4) 1821 different pixels:
    /tmp/mapnik-visual-images/tiff-reprojection-2-250-250-2.0-cairo.png (actual)
    tests/visual_tests/images/tiff-reprojection-2-250-250-2.0-cairo-reference.png (expected)
```

The test was enabled in https://github.com/mapnik/mapnik/commit/7fdae08ce1e1e67dbde8e9a24c5b7ad30db72043#diff-592b7c68836bded85ad1fb606511547e (and previous images were made up)

BergWerkGIS:
Just a little reminder, with the new visual diff in the test results `index.html`, it is easy to see where the differences are:

![image](https://cloud.githubusercontent.com/assets/4549888/10364839/4685518a-6dae-11e5-99ac-2db2f8e713fa.png)

springmeyer:
[...]

PR #3935, opened 2018-07-10
Fix bbox reprojection

lightmare:
Contains rebased #3752 and fixes bug discovered thanks to that test.

Function `calculate_bbox` could skip the first point because the order of evaluation of arguments to function call is indeterminate:
```c++
box2d<double> env(*it, *(++it));
```

I replaced `calculate_bbox` with `boost::geometry::envelope`.

Also while at it, I rewrote `envelope_points` to generate the requested number of points, instead of snapping to a multiple of 4, with an extra point added first to the top, then bottom, then right side.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -106,10 +106,7 @@ script:
  # (and might work) for the next build
  - DURATION=2400
  - scripts/travis-command-wrapper.py -s "date" -i 120 --deadline=$(( $(date +%s) + ${DURATION} )) make
- - RESULT=0
- - make test || RESULT=$?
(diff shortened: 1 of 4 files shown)


talaj:
Also https://github.com/mapnik/mapnik/issues/2438.

talaj:
Thank you!

lightmare:
Thanks for finding those related issues, had no idea what the actual impact was :bow_and_arrow:

springmeyer:
wow, incredible work @lightmare 🙏 🎉

talaj:
Would be great to port this to [v3.0.x](https://github.com/mapnik/mapnik/tree/v3.0.x).

PR #3944, opened 2018-07-20
remove old render_box function declaration from cairo_renderer.hpp

### LINKED ISSUE #3943: mapnik::cairo_renderer::render_box(box2d) lacks implementation.
Trying to use the [render_box](https://github.com/mapnik/mapnik/blob/master/include/mapnik/cairo/cairo_renderer.hpp#L174) function in cairo_renderer.hpp creates a linker error since the function is not implemented.

The commit creating this regression is c285cfa1, during the refactor of cairo_renderer the declaration of render_box remained but the implementation was removed.

Either remove the render_box declaration or implement it again.

EmilGedda:
fixes #3943


talaj:
Thanks!

Issue #3386, opened 2016-03-26
Add Ubuntu Trusty to Travis CI


knutole:
Would be great to include more Ubuntu versions in Travis CI. We're using Ubuntu Trusty, and over the last two years I've experienced issues with building Mapnik more times than I care to remember.

@lightmare mentioned that Ubuntu Precise is currently included in build tests, but would be great to test build on official Docker Ubuntu images also, https://hub.docker.com/_/ubuntu/, like Ubuntu Trusty.

I believe this could catch a lot of build problems on Ubuntu 14 and up?

PR #3945, opened 2018-07-20
travis: switch to dist: trusty

lightmare:
Precise should've been discontinued in April (https://blog.travis-ci.com/2017-08-31-trusty-as-default-status). Don't know when they'll finally remove it, but why wait for it?

Also bumped postgres version to 9.5, not that it was needed (on trusty they have 9.2 up to 9.6 preinstalled).


lightmare:
The coverage report doesn't make much sense. It probably got confused by the distro change.

lightmare:
Closes #3386

PR #3958, opened 2018-07-31
Fix buffer-size with scale-factor

talaj:
One would expect that rendering with `scale-factor` other than 1.0 has no effect on buffered extent of the map. This pull request corrects this behaviour.

For example, when rendering "retina" tile with `scale-factor` 2.0, `Map::width()` is 512 instead of 256, [Map::scale()](https://github.com/mapnik/mapnik/blob/3683c305577760033f902188957154e980c6f362/src/map.cpp#L676-L681) is half the `Map::scale()` with `scale-factor` 1.0. This causes to multiply map extent buffer by factor of 0.5.

### DIFF
--- a/include/mapnik/feature_style_processor_impl.hpp
+++ b/include/mapnik/feature_style_processor_impl.hpp
@@ -260,7 +260,7 @@ void feature_style_processor<Processor>::prepare_layer(layer_rendering_material 
     box2d<double> query_ext = extent; // unbuffered
     box2d<double> buffered_query_ext(query_ext);  // buffered
 
-    double buffer_padding = 2.0 * scale;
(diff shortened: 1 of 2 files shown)


lightmare:
Isn't it strange that this doesn't change any visual test results?

talaj:
No. There are just four visual tests with buffer-size:

```console
$ git grep -l buffer.size
image-filters-galore.xml
image-filters-multi-blur-inflate.xml
image-filters-multi-blur.xml
tiff-non-alpha-mask-band.xml
```

It affected some reference images in the downstream: https://github.com/mapycz/test-data-visual/commit/5fa22fe1abc6e81593a4d37d14e658e8b11f793f

PR #3962, opened 2018-08-01
bootstrap.sh improvements

### LINKED ISSUE #3954: bootstrap.sh `setup_mason` fails
Refs #3743 

https://github.com/mapnik/mapnik/blob/5aebd821f6107cb12d9c8199a9e3734cc07e7872/bootstrap.sh#L17-L23

The else branch is flawed. When `MASON_VERSION` is a commit hash, not a branch, you can't pull from it; the script will abort with an error like `fatal: Couldn't find remote ref e4c1746`.

Why pull anyway? I have a suspicion the pull was added because you used a branch name for `MASON_VERSION`, let `git checkout` create a local branch of that name and later `git checkout` would (correctly) checkout the local without fetching remote updates. But IMO `bootstrap.sh` shouln't mess with local branches at all, i.e. not create the local branch to begin with, and not merge into any local branch later.

```sh
    if [[ ! -d ./.mason ]]; then
        git clone https://github.com/mapbox/mason.git ./.mason
    else
        git -C .mason fetch --all
    fi
    git -C .mason checkout --detach "${MASON_VERSION}"
```

edit: forgot to add that when a branch name is used, it should include the remote name:
```sh
MASON_VERSION="origin/twig"
```

lightmare:
1. `source bootstrap.sh` shouldn't close the terminal when something goes wrong, it's not nice to users. Even [Travis doesn't like that](https://docs.travis-ci.com/user/customizing-the-build/#implementing-complex-build-steps), and he's a machine. So I wrapped the part that needs `set -e` from `install_mason_deps` onwards in a subshell.

2. Environment overrides have different syntax for `./bootstrap.sh` and `./configure`:
    ```sh
    CXX='c++' ./bootstrap.sh
    # versus
    ./configure CXX='c++'
    ```
    Added the ability to accept the second form, so now you can write:
    ```sh
    ./bootstrap.sh CXX='ccache c++'
    ```

3. prepend `.mason` to `PATH` only if it's not already there

4. change the way `.mason` is checked out
    - only fetch from remote(s) if the revision is unknown
    - checkout detached HEAD, don't create local branches

    Closes #3954

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,8 +1,5 @@
 #!/usr/bin/env bash
 
-set -eu
-set -o pipefail
-
 : '
 
 todo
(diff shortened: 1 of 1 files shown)


talaj:
Very nice.

talaj:
Thanks a lot, these issues were annoying.

Issue #3956, opened 2018-07-30
BuildingSymbolizer scale factor inconsistent across renderers


lightmare:
Creating this to document current behaviour. Each of the 3 renderers calculates building height differently.

I'm not going to fix this separately, because I have more changes in the works, and the extra update to reference images would unnecessarily bloat the `test-data-visual` repository.

AGG multiplies by scale factor:
https://github.com/mapnik/mapnik/blob/3683c305577760033f902188957154e980c6f362/src/agg/process_building_symbolizer.cpp#L86
![building-agg](https://github.com/mapnik/test-data-visual/raw/4d101ded5c18034629543401338889e9d8257e44/images/building-symbolizer-2-300-300-2.0-agg-reference.png)

Cairo doesn't multiply by scale factor:
https://github.com/mapnik/mapnik/blob/3683c305577760033f902188957154e980c6f362/src/cairo/process_building_symbolizer.cpp#L50
![building-cairo](https://github.com/mapnik/test-data-visual/raw/4d101ded5c18034629543401338889e9d8257e44/images/building-symbolizer-2-300-300-2.0-cairo-reference.png)

Grid doesn't multiply by scale factor and hard-codes the default:
https://github.com/mapnik/mapnik/blob/3683c305577760033f902188957154e980c6f362/src/grid/process_building_symbolizer.cpp#L70
![building-grid](https://github.com/mapnik/test-data-visual/raw/4d101ded5c18034629543401338889e9d8257e44/images/building-symbolizer-2-300-300-2.0-grid-reference.png)

Issue #1643, opened 2012-12-12
building symbolizer extrusion wall fill


bengolder:
Currently the building symbolizer uses a darkened version of "building-fill" in order to color the walls of the extruded form. It just multiplies the building fill color by 0.8.

See [this line](https://github.com/mapnik/mapnik/blob/master/src/agg/process_building_symbolizer.cpp#L122) and [this line](https://github.com/mapnik/mapnik/blob/master/src/agg/process_building_symbolizer.cpp#L151).

It would be great if there were a way to designate the fill color of the extrusion walls, perhaps with something like "building-wall-fill".

PR #3969, opened 2018-08-09
[WIP] BuildingSymbolizer styling

lightmare:
Refs #1643, #3956. These issues are just a tiny part, though. The rest is a major rewrite of BuildingSymbolizer rendering algorithm.

Fully opaque buildings should look the same.
For transparent buildings this is a noticeable change in appearance, often not for the better (yet).

#### New properties

- `side-fill = 80% fill`
  Suggestions for a better name are welcome. I wanted something more generic than `wall-fill`, because it needs a new mapnik-global enumeration value.

- `side-stroke = stroke // side-fill`
  Color for ground outline and vertical wall edges. Verticals are not implemented here, so this should probably be renamed to `ground-stroke`, `base-stroke` or something.
  I'm not happy with the double-fallback default, might change it to `stroke` only.

- `stroke = 80% fill`
  Color for roof outline.

- `stroke-opacity = 1`
  Previously `fill-opacity` was also used for strokes. I consider that a bug and that's one reason why `stroke-opacity` defaults to 1 and not `fill-opacity` for backward compatibility. Another reason is that taking the default from `fill-opacity` would be really unexpected if `stroke` color is also specified, and having rules saying when it defaults to what would just be confusing.

- `stroke-width = 1`
  Applies to both roof and ground outlines.

Better styling is nice, however that's not how this PR started.

#### Z-order

After fixing one broken polygon in test data I noticed an issue with a multipolygon:
building-symbolizer-2-300-300-2 0-agg
I considered it minor because BuildingSymbolizer was probably not intended for such high heights. There's no easy way to do proper z-order across features, so solving it within a multipolygon would only be a minor improvement. But it was so easy to fix that I jumped on it -- first render all walls, then all roofs. We can also save projected vertices during the first pass, so that we don't need to do the expensive transformation twice. And then I wanted to delve deeper...

#### Walls in the old algorithm

Every line segment produced a 4-point patch to fill and stroke, that's at least 12 vertices to rasterize per 1 vertex of input. When the fill==stroke color is opaque, some walls are pure overhead (because they're completely covered), and about half of stroke lines as well. When the fill==stroke color is transparent, twice-stroked "vertical" edges look quite good on buildings with clean geometry and sharp corners; on round walls it depends on the density and regularity of the linear approximation; and on non-simplified geometry even apparently straight walls have multiple edges:
building-symbolizer-138-200-200-1 0-agg building-symbolizer-134-200-200-1 0-agg building-symbolizer-179-200-200-1 0-agg

Below is the new implementation without vertical edges (a bit bland, isn't it?):
building-symbolizer-138-200-200-1 0-agg-reference building-symbolizer-134-200-200-1 0-agg-reference building-symbolizer-179-200-200-1 0-agg-reference

Much later, when comparing samples from the old and new algorithm, I noticed that the old way of stroking the 4-point patch with invalid miterlimit caused bleeding on jaggy geometries:
building-symbolizer-143-200-200-1 0-agg building-symbolizer-143-200-200-1 0-grid
This is fixed in one of the commits preceding the complete overhaul, in case the whole thing doesn't get merged.

#### Missing verticals

I'm not rendering vertical wall edges, at least for now. Fully opaque buildings don't care, but for transparent buildings this makes a huge difference. Currently I'm evaluating ways to re-introduce "important" edges that actually enhance the appearance.

One problem I have created (yet haven't realized soon enough) is that if `side-stroke` is not the same color as `side-fill` (doesn't matter if opaque or partially transparent), the walls and edges would need to be rendered in back-to-front order to give the expected result.

I didn't want to go back to sorting line segments. However, if we do decide to sort, then we can try crazier stuff -- like shading the walls according to angle (light direction), and then edges won't need to be stroked as they'll pop out naturally.


lightmare:
#### New algorithm

It starts with the ground outline. The four SVG images below are hand-made, not renderer output. They show the source geometry as a light gray polygon with thin black outline. The thick gray line is the ground stroke that's rendered:
ground

Next it renders contiguous wall "strips". For example walls in the "hole" were previously rendered as five 4-point patches; now it's two patches: one 8-point and one 6-point. Back walls are shown in red, although in actual rendering they're the same color, and not rendered at all if wall and roof fill colors are opaque:
walls

Finally the roof is put on top:
roof

Combining the three steps yields:
combined

<details>
<summary>Irrelevant miter join ramble...</summary>

I intentionally used line joins exceeding miter limit in these illustrations to show a general issue with miter joins.

SVG "miter" join never looks good. That's `CAIRO_LINE_JOIN_MITER` and `agg::miter_join_revert`.

SVG2 defines "miter-clip", which is what `agg::miter_join` does. Cairo doesn't know this, so we can't have nice miters and similar appearance in agg and cairo at the same time.

See https://www.w3.org/TR/svg-strokes/#StrokeMiterlimitProperty
miterlimit

</details>


#### Miter woes

As you can see in the combined image, ground/roof strokes stretch too far outwards. The illustration is exaggerated, but the spikes are clearly visible even with stroke-width=1. The previous solution to this were vertical edge strokes, which filled the gaps, although in a weird way. Drawing the vertical edge properly requires a 3-way join -- the ground outline makes its join, and the vertical sits on it. In order for the vertical to line up precisely, the ground join must be "round", otherwise the distance that the stroke/miter goes along x-axis depends on segment angles.

Another option would be to place walls on the outer edge of the stroke, instead of directly on the source geometry. At first I thought this would be overkill, so I dismissed the idea. Note to self: maybe worth revisiting; it shouldn't require running a full-fledged offset converter, on-the-fly "miter-clip"-like offset should suffice.

The solution I chose is adding a thin (`stroke-width/2`) rectangle on both sides of every wall "strip". Here's the final result (actual rendering):
building-symbolizer-3-420-350-1 0-agg-reference
Under close inspection this doesn't look "right", as walls inexplicably cut into the outline, but with thin stroke it looks fine, and only costs 4 extra vertices per wall strip.

lightmare:
Rebased on master, renamed new properties, and re-introduced vertical edge strokes at corners.

I still didn't want to add an edge on every joint, so I used a simple heuristic -- draw an edge only if the next segment diverges by more than 5 degrees from the direction from the previous edge. This eliminates some edges on practically straight walls:
building-symbolizer-152-200-200-1 0-agg-reference building-symbolizer-152-200-200-1 0-agg

... and also coarsens round walls made from many short segments:
building-symbolizer-150-200-200-1 0-agg-reference building-symbolizer-150-200-200-1 0-agg

However, sometimes originally dense round wall approximation becomes irregular:
building-symbolizer-191-200-200-1 0-agg-reference building-symbolizer-191-200-200-1 0-agg

The threshold angle is currently hardcoded, but can be tweaked. Perhaps we should expose it to style xml, defaulting to zero for least user surprise.

I also snap vertical edges' x-coordinate to pixel center to make them crisp, I think it looks better:
building-symbolizer-126-200-200-1 0-agg-reference building-symbolizer-126-200-200-1 0-agg

lightmare:
In case anyone's interested in benchmarks, I took one simple (`building-symbolizer-1.xml` without the background pattern, only color), and one complex building (roughly 300 points).

<details>
<summary>Before:</summary>

```sh
$ test/visual/run -o -vd -i 2000 -s 1.0 --agg --cairo --grid --styles test/data-visual/building-*.xml
"building-complex-solid-200-200-1.0" with agg... OK (3416 milliseconds)
"building-complex-solid-200-200-1.0" with cairo... OK (6691 milliseconds)
"building-complex-solid-200-200-1.0" with grid... OK (3672 milliseconds)
"building-complex-transparent-200-200-1.0" with agg... OK (3404 milliseconds)
"building-complex-transparent-200-200-1.0" with cairo... OK (7446 milliseconds)
"building-complex-transparent-200-200-1.0" with grid... OK (3688 milliseconds)
"building-simple-solid-256-256-1.0" with agg... OK (498 milliseconds)
"building-simple-solid-256-256-1.0" with cairo... OK (2650 milliseconds)
"building-simple-solid-256-256-1.0" with grid... OK (826 milliseconds)
"building-simple-transparent-256-256-1.0" with agg... OK (562 milliseconds)
"building-simple-transparent-256-256-1.0" with cairo... OK (2868 milliseconds)
"building-simple-transparent-256-256-1.0" with grid... OK (823 milliseconds)

Visual rendering: 0 failed / 12 passed / 0 overwritten / 0 errors
agg: 	7881 milliseconds
cairo: 	19656 milliseconds
grid: 	9011 milliseconds
total: 	36548 milliseconds
```
</details>

<details>
<summary>After:</summary>

```sh
$ test/visual/run -o -vd -i 2000 -s 1.0 --agg --cairo --grid --styles test/data-visual/building-*.xml
"building-complex-solid-200-200-1.0" with agg... OK (1944 milliseconds)
"building-complex-solid-200-200-1.0" with cairo... OK (4254 milliseconds)
"building-complex-solid-200-200-1.0" with grid... OK (2101 milliseconds)
"building-complex-transparent-200-200-1.0" with agg... OK (2415 milliseconds)
"building-complex-transparent-200-200-1.0" with cairo... OK (6043 milliseconds)
"building-complex-transparent-200-200-1.0" with grid... OK (2087 milliseconds)
"building-simple-solid-256-256-1.0" with agg... OK (424 milliseconds)
"building-simple-solid-256-256-1.0" with cairo... OK (2277 milliseconds)
"building-simple-solid-256-256-1.0" with grid... OK (750 milliseconds)
"building-simple-transparent-256-256-1.0" with agg... OK (542 milliseconds)
"building-simple-transparent-256-256-1.0" with cairo... OK (2684 milliseconds)
"building-simple-transparent-256-256-1.0" with grid... OK (754 milliseconds)

Visual rendering: 0 failed / 12 passed / 0 overwritten / 0 errors
agg: 	5326 milliseconds
cairo: 	15260 milliseconds
grid: 	5694 milliseconds
total: 	26281 milliseconds
```
</details>

andrzej-r:
Is the Z-order now fixed? In #3956 the same shape rendered correctly. Without a fix building symbolizers are not really useful, I can see a lot of such errors in build-up areas.

What I'd also like to have is an ability to make all buildings opaque and _then_ apply transparency to whole layer. That is, I'd like the buildings to be semi-transparent in such a way that layers under them are visible but not their internal/back walls. Is such feature available?

lightmare:
@andrzej-r 
> Is the Z-order now fixed? In #3956 the same shape rendered correctly. Without a fix building symbolizers are not really useful, I can see a lot of such errors in build-up areas.

The image in #3956 looks correct only because the geometry in mapnik/test-data is currently broken -- according to OGC Simple Features specification, it is not valid -- it has a hole in a hole. Once the geometry in test-data is fixed, it'll render as shown here.

Then, BuildingSymbolizer alone cannot solve Z-order across features (separate Polygons or MultiPolygons). Mapnik assumes only 2D layers, and renders features one by one, as they come out of a datasource. BuildingSymbolizer is deliberately cheating to create a cheap illusion of height on this 2D sheet. It's not supposed to produce true 3D effect.

If you're thinking about rendering buildings with their actual height, I doubt BuildingSymbolizer was meant to be used that way. You'd need a fully 3D map for that to make sense, which is not feasible with Mapnik. Imagine a tall building with a road at its ground level and another road 3 storeys above the first. Where do you render the high road? Building height doesn't have much meaning if both roads are rendered at its ground level.


> What I'd also like to have is an ability to make all buildings opaque and _then_ apply transparency to whole layer. That is, I'd like the buildings to be semi-transparent in such a way that layers under them are visible but not their internal/back walls. Is such feature available?

Yes, that's possible with `Layer` attribute `opacity`.

andrzej-r:
Is there any chance of fixing/improving this mechanism? Indeed all buildings would have to be rendered together, not one by one. Unfortunately without doing it the building symbolizer is broken in a way that cannot be fixed in the code using it (by sorting buildings before rendering).

The layer opacity works very well indeed, thank you.

I get the point about "full 3d" maps, but that's not my goal. 2.5D for buildings only is fine but it should work with real data (i. e. data from osm).

andrzej-r:
I suppose the right question to ask is: _is it doable at all_? The buildings indeed would have to be rendered all at once by somehow queuing features belonging to the same layer. Unfortunately, without that building symbolizer cannot be used for rendering buildings.

I have no problem grouping buildings in a single layer. That's, in fact, what I already had to do in https://github.com/gravitystorm/openstreetmap-carto/pull/3427 (at present there are two building layers) to sort "special buildings" together with other buildings. Unfortunately, sorting by coordinates only goes so far, there are many cases where it fails either because of irregular building shapes or because of ordering conflicts.

lightmare:
> I suppose the right question to ask is: _is it doable at all_?

Not without some changes outside of BuildingSymbolizer itself. It would have to collect edges (as you pointed out, sorting features/polygons is not sufficient), and then in a separate final step render walls in proper order. And it would get more complicated with multiple symbolizers (e.g. with varying properties or passing different rule filters) in one style/layer. For a single symbolizer, or several with identical properties, it'd be easier -- I've pondered the idea of multi-pass symbolizers while working on this, but no proof of concept exists. It's certainly outside the scope of this PR and I won't have time to implement that any time soon.

andrzej-r:
I'm sorry for hijacking this PR, please let me know if there is a better place for this discussion.

It looks like a non-trivial task indeed. I don't think the solution would have to be generic - as you pointed out earlier, a truly generic solution would have to be a 3d projection of an entire scene, including multi-level roads, POIs at different floors etc. That generally needs some level of interactivity to be useful, so it would not work well if rendered statically anyway.

A question about multiple symbolizers, what does it mean in practice? If I render some buildings in a different colour (as I do) does it mean they are using a different symbolizer from other buildings? Or is it about mixing e. g. buildings and roads in a single layer? I think it would be fair to assume there is only one layer with buildings and that it doesn't contain any non-building features.

lightmare:
> If I render some buildings in a different colour (as I do) does it mean they are using a different symbolizer from other buildings?

If they pass different filters, then yes. Check the xml style if you're unsure how the css translates into that.

PR #3976, opened 2018-08-24
[WIP] experimental Spirit X3 hacks

lightmare:
This is an experiment I started because I hate how Spirit X3 separates rules from parsing expressions with the `_def` suffix. Jumping between where a rule is used in the grammar and where it's defined is awkward because they're different identifiers. Well, I could make vim do that with a keystroke, but I chose not to write vimscript for this.

C++ allows a name to refer to a type and a variable (or function) at the same time. They don't shadow each other. Although the variable/function takes precedence where it's syntactically allowed, you can always get to the type with an explicit *class-key* (`struct`, `enum`, ...). So I use the same name for the `Tag` type used to differentiate `x3::rule` specializations, and for the variable used in parsing expressions.

Rule parsing expressions are stored in a variable template specialized on the `Tag` type. After testing a few variations of this approach, I found it's already been explored (https://github.com/djowel/spirit_x3/issues/17).

So there are no suffixes, everything related to a rule is designated by a single name.

```c++
struct MyRule { static MyRule proxy; ... }; // type named MyRule
struct MyRule const MyRule = MyRule::proxy; // variable named MyRule
MyRule MyRule::proxy = "Some nonterminal";
template <> auto const _rule_def_<MyRule> = ... parsing expression ...
```

The `MyRule` variable could've been simply an `x3::rule` instance, but that would suffer from initialization order issues because `x3::rule` constructor is not `constexpr`. So I instead made the `struct MyRule` type usable in parsing expressions (it converts to an `x3::rule`).

So far I've only converted TopoJSON grammar to this style (despite my opinion that Spirit is not the right tool to parse JSON-based protocols and that this grammar is flawed, I had my hands on it when I got the idea :)). With gcc-6, `topojson_grammar_x3.o` size dropped by ~30k (~10%), that accounts for ~1% of `libmapnik-json.a`. I'm curious whether other grammars will also shrink. I kinda expected the opposite, as I made `x3::rule` type signatures longer. Perhaps the reduction is due to `BOOST_SPIRIT_DEFINE` putting parsing expressions in function-local static variables, whose names are really long, because they include both the function type and the variable type.

### DIFF
--- a/include/mapnik/json/topojson_grammar_x3.hpp
+++ b/include/mapnik/json/topojson_grammar_x3.hpp
@@ -25,24 +25,21 @@
 
 // mapnik
 #include <mapnik/json/topology.hpp>
+#include <mapnik/util/spirit_rule.hpp>
 
-#pragma GCC diagnostic push
-#include <mapnik/warning_ignore.hpp>
-#include <boost/spirit/home/x3.hpp>
-#pragma GCC diagnostic pop
+namespace mapnik { namespace topojson { namespace grammar {
 
-namespace mapnik { namespace json { namespace grammar {
+MAPNIK_SPIRIT_EXTERN_RULE(start_rule, topojson::topology);
 
-namespace x3 = boost::spirit::x3;
+}}} // namespace mapnik::topojson::grammar
 
-using topojson_grammar_type = x3::rule<class topojson_rule_tag, topojson::topology>;
+namespace mapnik { namespace json {
 
-BOOST_SPIRIT_DECLARE(topojson_grammar_type);
-
+inline auto topojson_grammar()
+{
+    return topojson::grammar::start_rule;
 }
 
-grammar::topojson_grammar_type const& topojson_grammar();
-
-}}
+}} // namespace mapnik::json
 
 #endif //MAPNIK_TOPOJSON_GRAMMAR_X3_HPP
--- a/include/mapnik/json/topojson_grammar_x3_def.hpp
+++ b/include/mapnik/json/topojson_grammar_x3_def.hpp
@@ -42,7 +42,7 @@ BOOST_FUSION_ADAPT_STRUCT(
 
 BOOST_FUSION_ADAPT_STRUCT(
     mapnik::topojson::arc,
-    (std::list<mapnik::topojson::coordinate>, coordinates)
+    (mapnik::topojson::position_array, coordinates)
     )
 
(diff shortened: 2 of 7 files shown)


lightmare:
I'll try to explain how this works on a simple grammar. Let's parse a subset of JSON that only knows booleans, numbers, and arrays.

Rules that are not involved in recursive cycles, and don't need to be used from other sources, are defined with `MAPNIK_SPIRIT_LOCAL_RULE(Tag[, Attribute])`. The `Attribute` argument is optional and defaults to `x3::unused_type`.
```c++
// grammar_def.hpp

MAPNIK_SPIRIT_LOCAL_RULE(r_boolean, bool)
    = "false" >> x3::attr(false)
    | "true" >> x3::attr(true)
    ;
```

When there is a recursive cycle, some rule(s) must be introduced first with `MAPNIK_SPIRIT_RULE(Tag[, Attribute])` and later specialized with `MAPNIK_SPIRIT_RULE_DEF(Tag)`.
```c++
// grammar_def.hpp

MAPNIK_SPIRIT_RULE(r_value, my_value_type);

MAPNIK_SPIRIT_LOCAL_RULE(r_array, my_array_type)
    = '[' > -(r_value % ',') > ']'
    ;

MAPNIK_SPIRIT_RULE_DEF(r_value)
    = x3::double_
    | r_boolean
    | r_array
    ;
```

When a rule is to be used from other grammars/sources, it goes like this:

```c++
// grammar.hpp

MAPNIK_SPIRIT_EXTERN_RULE(r_start_rule, boost::optional<my_value_type>);

// grammar_def.hpp

MAPNIK_SPIRIT_EXTERN_RULE_DEF(r_start_rule)
    = r_value
    | "null"
    ;

// grammar.cpp

MAPNIK_SPIRIT_INSTANTIATE(r_start_rule, iterator_type, context_type);
```

And finally, all rules must have their fancy names (those that appear in `x3::expectation_failure` exception) defined once:
```c++
// grammar.cpp

MAPNIK_SPIRIT_RULE_NAME(r_boolean) = "Simpleson boolean";
MAPNIK_SPIRIT_RULE_NAME(r_value) = "Simpleson value";
MAPNIK_SPIRIT_RULE_NAME(r_array) = "Simpleson array";
MAPNIK_SPIRIT_RULE_NAME(r_start_rule) = "Simpleson message";
```

PR #3979, opened 2018-08-31
sqlite: Add !pixel_width! and !pixel_height!

rbray89:
Adding basic support for some additional tokens.


talaj:
Travis reports some failing visual tests. All are using sqlite datasource. Can you, please, take a look on them? I think that this change should not have any effect on these visual tests.

rbray89:
That's what I thought when I looked at the tests. The only possible logical change would be https://github.com/mapnik/mapnik/pull/3979/commits/7d843b255f2a53ee5a67093b225e0a99332aab09#r214278495

Though that would only affect the visual test if there was a query with `!intersects!` in it.

lightmare:
> Though that would only affect the visual test if there was a query with !intersects! in it.

You're passing `table_` to `populate_tokens`, but the `query` may have been modified in the conditional branch.

lightmare:
Sorry, I should've explained better.
The branch calling `sqlite_utils::apply_spatial_filter` should come first, because it replaces `!intersects!` with an actual check involving subquery.
`populate_tokens` replaces `!intersects!` with a dummy check `1=1` (to make the query valid if the filter is not applied), so it should be called second.

Tests passing with the change that swaps the order show that we have no test for the functionality of `sqlite_utils::apply_spatial_filter` (actually no test for sqlite plugin, all were in python I guess). This needs to be a unit test on what features come out of the datasource query. Visual tests, as evidenced, are not affected when the filter is skipped. (Just making notes, not asking for this test to be included in this PR).

rbray89:
As far as test-cases, as far as I can tell, it seems there are no test queries covering pixel_width at all (including for postgis). I assume If I wanted to add some, I'd add a new image to the data-visual repo and a new xml file that generates it? I'm a little new to  pull requests with submodule changes. I assume the test repo would have to be updated first and then I'd repoint to the latest?

talaj:
@rbray89 Thanks for the update. I think it can be merged now.

If you want to make a test, unit test would be more appropriate. There is no unit test for sqlite datasource yet.

Regarding your question, you can send pull request to the repository with test data alongside this one. Don't forget to update reference to the submodule in this repo. Tests in Github will run as expected and it can be merged without further changes.

alsaleem00:
thanks for adding these two tokens. 
It would be also better to add the !bbox! token.
The reason is that it should be easier to switch to sqlite from posgis using same select (xml)

artemp:
@alsaleem00 thanks for the suggestion. It should be possible to add '!bbox!' token, I'll try to find time to look into this, thanks!

alsaleem00:
@artemp  I managed to add support for spatialite (mod_spatiallite) in sqlite driver (input plugin). It works.
I noticed that mapnik does not emit the plugin error (in case it did not find required so files). Maybe it is sqlite problem!
I had to test existence of all required so libraries through writing a separate code.

artemp:
> @artemp I managed to add support for spatialite (mod_spatiallite) in sqlite driver (input plugin). It works.

:+1:

 I noticed that mapnik does not emit the plugin error (in case it did not find required so files). Maybe it is sqlite problem! I had to test existence of all required so libraries through writing a separate code. If you think it is a good addition, i can make PR

Can you describe briefly how are you proposing to address this, pls?

alsaleem00:
Mapnik code is clear and well organized.
First, User may add additional tags to the xml file
`<Parameter name="sqlite_extn_1"><![CDATA[mod_spatialite]]></Parameter>`
`<Parameter name="extn_1_init"><![CDATA[sqlite3_mod_spatialite_init]]></Parameter>`
The second tag is optional if the extension does not follow sqlite3 rules. It is the library entry point.

Then, in (**sqlite_datasource**) constructor we get these tags and store/pass them to **sqlite_connection** constructor.
In **sqlite_connection**, we use **load_extension** to load the extension after opening the database.

I actually added support for two exentions, namely, e.g. spatiallite and regexp. I prefer to add support for three/four since sqlite is limited and there are plenty of extensions that can be used.

I also added support for !bbox! in the table query.

Issue #3977, opened 2018-08-25
How to render streets with a patters that behaves similar to “polygon-pattern”?


sommerluk:
Hello.

At https://github.com/gravitystorm/openstreetmap-carto we are trying to render unpaved roads different from paved ones.

Normally, I would avoid to ask this type of questions here on the issue tracker, but I started to write the code for openstreetmap-carto in May 2017, since than the solution is in continuous development, we have tested a lot of possible solutions, but I still struggle to have a well-performing solution. That’s why now, after one year, finally I ask here.

# What is the desired rendering?

## Paved roads
We want to render paved roads still like they do currently in openstreetmap-carto: with a plain colour.

## Unpaved roads
Unpaved roads should however be rendered from now on with a fine, irregular-looking pattern. [At this image](https://user-images.githubusercontent.com/6830724/39374051-2c22b3a4-4a39-11e8-8fdc-06eb89698583.png) you can see the rendering we want to get. The irregular pattern fills exactly the space that would have been occupied for a paved road. Neither at crossings between two unpaved roads nor at crossings between a paved and an unpaved road there is any visual artefact.

## Further requirements
We are using round line caps for roads to avoid visual artefacts at places where OSM ways are split.

# line-pattern?
The natural choise would be Mapnik’s line-pattern feature. Mapnik renders this pattern following the geometry of the OSM way, blending it at angles within the OSM way. [See here](https://github.com/sommerluk/roadpatternrendering#pattern-rendering-with-line-pattern) for a minimalistic example. But I do not get the desired rendering because line-pattern seems to have no support for round line caps, so this leads to ugly rendering artefacts on every single angle in the line, and especially where OSM ways end. (Furthermore, we need many pattern files: one for each combination of road colour and width. But the main problem with this solution are the rendering artefacts.)

# dash-array?
We could use various dash-arrays to [try to get a similar visual effect](https://github.com/sommerluk/roadpatternrendering#pattern-rendering-with-dash-array). But also here, we have rendering artefacts.

# polygon-pattern on line geometries?
[...]

PR #3980, opened 2018-08-31
WIP: Polygon pattern on line geometries

talaj:
Addresses https://github.com/mapnik/mapnik/issues/3977.

I'm not sure if such functionality belongs to PolygonPatternSymbolizer, but given simplicity of that change, why not?

I set polygon filling-rule to `agg::fill_non_zero` for line geometries to allow lines cross themselves. 

There is one failing visual test where PolygonPatternSymbolizer is already used on line string geometry.

It should not be a problem to implement it also for Cairo backend.

text-halign-adjust-800-800-1 0-agg_2

```xml
<Map background-color="white" srs="+proj=latlong +datum=WGS84">
    <Parameters>
        <Parameter name="sizes">800, 800</Parameter>
        <Parameter name="bbox">-0.05, -0.01, 0.95, 0.01</Parameter>
    </Parameters>

    <Layer name="layer" srs="+proj=latlong +datum=WGS84">
        <StyleName>My Style</StyleName>
        <Datasource>
            <Parameter name="file">../../data/csv/lines2.csv</Parameter>
        </Datasource>
    </Layer>

    <Style name="My Style">
        <Rule>
            <PolygonPatternSymbolizer
                stroke-width="20"
                stroke-linecap="round"
                file="../../data/svg/clippath.svg"
                transform="scale(0.1)"
                />
        </Rule>
    </Style>
</Map>
```

text-halign-adjust-800-800-1 0-agg

```xml
<PolygonPatternSymbolizer
    stroke-width="20"
    stroke-linecap="round"
    file="../../data/svg/stripes_pattern.png"
    />
```

@kocio-pl @sommerluk


lightmare:
This is so cool! :snowflake:

sommerluk:
Having this in Mapnik would be really great! Thanks a lot!

My questions:

- I suppose implement this in the `line` symbolizer (as a parameter that “overwrites” `line-color`) would be considerably more complicate than having it in the `polygon-pattern` symbolizer, wouldn’t it?
- `stroke-dasharray` and `stroke-offset` are not available in your PR. Indeed that’s something we do not need at [openstreetmap-carto](https://github.com/gravitystorm/openstreetmap-carto/), but if it’s easy to implement, it might be worth to add it also.

Thanks again for implementing this!

talaj:
> I suppose implement this in the line symbolizer (as a parameter that “overwrites” line-color) would be considerably more complicate than having it in the polygon-pattern symbolizer, wouldn’t it?

That is the main problem of this pull request. It would not be complicated to implement it as part of LineSymbolizer. It is a design question. Does existence of PolygonPatternSymbolizer even make sense if we could set a patten to PolygonSymbolizer?

I think the functionality should be part of LinePatternSymbolizer or PolygonPatternSymbolizer. I feel it more appropriate for LinePatternSymbolizer, but it would need completely new option to select between the two types of line patterns. PolygonPatternSymbolizer doesn't need such option. This is why I tried to implemented it there first. 

> stroke-dasharray and stroke-offset are not available in your PR. Indeed that’s something we do not need at openstreetmap-carto, but if it’s easy to implement, it might be worth to add it also.

Yes, these can be added easily.

sommerluk:
> It is a design question.

Yes, I see your point.

`PolygonPatternSymbolizer` would work well for openstreetmap-carto also.

sommerluk:
An additional question: At openstreetmap-carto, we try to use the pattern rendering style not only for roads (line geometries), but we do the same also for turning circles (point geometries, that we render with a `marker` symbolizer as circles in the same colour as the road, and with varying diameters). Would the new PolygonPatternSymbolizer also accept point geometries, just rendering as circles with the diameter specified in stroke-width?

(We could of course generate a set of round SVG images at the corresponding diameters and colours, but this would not give us the benefit of globally aligned patterns.)

talaj:
@sommerluk I'm implementing it as part of LinePatternSymbolizer finally. The last reason why I decided so is to allow rendering line on polygon geometries. It is important use case often used in cartography. It would not be possible to render a line on polygon with PolygonPatternSymbolizer.

I agree having globally aligned patterns on points can be useful, but LinePatternSymbolizer is not right tool for such job.

You can use [ST_Buffer](http://postgis.net/docs/ST_Buffer.html) with PolygonPatternSymbolizer as a workaround.

lightmare:
> I agree having globally aligned patterns on points can be useful, but LinePatternSymbolizer is not right tool for such job.

In cairo, zero-length line with round cap is rendered as a round dot: https://www.cairographics.org/manual/cairo-cairo-t.html#cairo-stroke
Unfortunatelly, to get the same effect in agg, the line must be "long enough" -- I guess to not fall in the same pixel sampling cell.

> Now realizing that `ST_Buffer` could be also used as a workaround for line patterns you are currently doing by comp-ops.

That's only possible where `pixel_width:pixel_height` is close to `1`, elsewhere N-S roads would end up with different thickness than W-E roads. `ST_Buffer` operates in map coordinates.

sommerluk:
> ST_Buffer could be also used as a workaround for line patterns you are currently doing by comp-ops.

Well, suppose this would be a no-go for us also because we have for the different road types far more than a dozen different road widths _per zoom level_. Multiplying this with about 15 zoom levels that actually show roads, this would be a so big number of SQL code for these geometries that it would be impossible to maintain ;-)

talaj:
> In cairo, zero-length line with round cap is rendered as a round dot: https://www.cairographics.org/manual/cairo-cairo-t.html#cairo-stroke

Then we should consider it also for `agg::conv_stroke`.

talaj:
I've added commits which implement the functionality on LinePatternSymbolizer for AGG, Cairo and Grid renderers.

LinePatternSymbolizer has new option `line-pattern` with following values:

- `warp` - the original line pattern functionality; default value
- `repeat` - works the same way as patterns on polygons

LinePatternSymbolizer now supports also these options known from LineSymbolizer and PolygonPatternSymbolizer:

- `stroke-width`
- `stroke-linecap`
- `stroke-linejoin`
- `stroke-miterlimit`
- `stroke-dasharray`
- `alignment`

I tried to share pattern rendering code with PolygonPatternSymbolizer. There was and still is duplicate code between renderers, but I intentionally didn't try to fix it as part of this issue/pull request.

artemp:
@talaj - this is great! Would you like to merge?

talaj:
@artemp Yes, please.

kocio-pl:
Great! I'm very happy with making real, generic, upstream solution instead of risky workarounds in downstream.

I guess what we at OSM Carto need for development is porting it to 3.0.x, release + npm package, so Kosmtik could use it (some more things are needed for production deployment).

talaj:
@kocio-pl Yes, I will try to merge it to 3.0.x.

I would also like to make it work on point geometries as described above.

I'm also working on fixing https://github.com/gravitystorm/openstreetmap-carto/issues/937. I have it fixed in Mapnik for Mapy.cz so I'm only porting it to the upstream. It's unfortunate such bug was not fixed earlier.

sommerluk:
@talaj I’ve taken a look the the options list, and it seems fine for me. All our use cases for line geometries would be covered with that. Also great to hear that backporting to 3.0.x is planned.

Thanks a lot for implementing this so fast! This will really help us a lot at openstreetmap-carto. That’s really great work, guys!

sommerluk:
Again, thanks a lot for fixing this bug!

It has taken a long time, but finally since January 14, 2023 we are using this feature actually at openstreetmap.org. Great work!

PR #3993, opened 2018-09-29
Compatibility patch for Boost GIL v1.68

tdubrownik:
A number of basic typedef were removed in the Boost GIL 1.68 release,
notably `bits32` and `bits32f`. Additionally the
`GIL_DEFINE_ALL_TYPEDEFS` macro signature changed.

This commit introduces compatibility type aliases and conditionally
calls the new macro if `BOOST_VERSION` is high enough.

Based on the output of:
```
% grep -r bits32 . | cut -d: -f1 | sort -u 
./deps/boost/gil/extension/toolbox/hsl.hpp
./deps/boost/gil/extension/toolbox/hsv.hpp
./include/mapnik/image_filter.hpp
``` 
…it seems this covers all the uses of at least the bits32* typedefs.


talaj:
Thank you!

PR #4004, opened 2018-10-23
v3.0.x: Line pattern "repeat"

talaj:
Port of https://github.com/mapnik/mapnik/pull/3980 to v3.0.x.

### DIFF
--- a/include/mapnik/agg/render_polygon_pattern.hpp
+++ b/include/mapnik/agg/render_polygon_pattern.hpp
@@ -1,0 +1,156 @@
+/*****************************************************************************
+ *
+ * This file is part of Mapnik (c++ mapping toolkit)
+ *
+ * Copyright (C) 2017 Artem Pavlenko
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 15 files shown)


sommerluk:
@talaj Thanks for the port!

sommerluk:
@nebulon42 This PR adds more options to an existing symbolizer. To use it in CartoCSS, would it work right out of the box? Or does it require also an update of https://github.com/mapnik/mapnik-reference/ ? Or does it require even a new CartoCSS release?

talaj:
@sommerluk I've prepared https://github.com/mapnik/mapnik-reference/pull/153 for that.

sommerluk:
@talaj That’s great! Thanks.

nebulon42:
Yes, it would need an update of mapnik-reference as @talaj already pointed out. It also requires a new carto release if mapnik-reference increases its version in a way that is not covered by the pattern in the [package.json](https://github.com/mapbox/carto/blob/master/package.json#L40). The pattern is ~ so that means that whenever mapnik-reference goes to 8.10.x we need a new release there. This should, however, not influence the versioning decisions. This change, according to semver, would IMO require to go to 8.10.0. A carto release is not much work, it would be a patch release, even if its impact is a bit larger.

sommerluk:
@talaj Could you also update the [npm package](https://www.npmjs.com/package/mapnik-reference) (currently 8.9.2)?

talaj:
@sommerluk, yes, I'm going to do that.

talaj:
@sommerluk Just published [mapnik-reference 8.10.0](https://www.npmjs.com/package/mapnik-reference/v/8.10.0).

sommerluk:
@talaj That’s great, thanks!

PR #4006, opened 2018-10-30
fix: makes serveral header definitions static to support static library builds

nikhedonia:


artemp:
applied in e583063ab55a064c4aea781b48407df3684dbaaf

Issue #3749, opened 2017-09-01
SVG output file is just a wrapper for a PNG image!?


alexandervlpl:
I want to generate SVG maps in Python using Mapnik 3.0.13 (built from source). I'm using version 3.0.13 of the Python bindings as well, built with PYCAIRO as described by @artemp in #3438. My code:

```
f = open('/tmp/test.svg', 'w')
surface = cairo.SVGSurface(f.name, map.width, map.height)
mapnik.render(map, surface)
surface.finish()
```
The output SVG file looks like this:
```
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="4961pt" height="3508pt" viewBox="0 0 4961 3508" version="1.1">
<defs>
<image id="image3198" width="20671" height="14617" xlink:href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAUL8A ...
```
Followed by ~200 MB of base64-encoded PNG garbage. :fearful:

Guys, I've been using your amazing library for years. **Is this a joke?** There's a secret undocumented way to create an _actual_ SVG, isn't there? What is it? How do I use Cairo or whatever `svg_renderer` is to get something usable here?

alexandervlpl:
Update: after recompiling my library and bindings to use the experimental `svg_renderer`, I'm getting the exact same result.

talaj:
The experimental `svg_renderer` is something different than SVG from Cairo backend. The first one is not usable at all and no development is going there.

The Cairo SVG rendering should work fine. Raster may cause problems. Isn't that 200 MB PNG garbage just raster in the background? Vector only maps should look fine with reasonably sized output.

alexandervlpl:
Thanks for the tip, switching back to Cairo.

No, this isn't a raster in the background. I'm experimenting with the default OpenStreetMap style. I'm getting a single huge PNG with all the streets, buildings, etc. The PNG is base64 encoded and embedded in the SVG. Nothing is actually rendered as vectors.

alexandervlpl:
Here's an example output file:
https://inkatlas.com/custom/test.svg

It's ~75MB, best not to open it in a browser.

alexandervlpl:
The image ends up as the only item in a group (defined at the bottom of the file):

```
<g id="surface1">
<use xlink:href="#image8" transform="matrix(0.24,0,0,0.24,0,0)"/>
</g>
[...]

PR #4029, opened 2019-02-08
Cairo: set version of SVG to 1.2

talaj:
This partially solves https://github.com/mapnik/mapnik/issues/3749.

The problem was that Cairo SVG surface is using SVG version 1.1 by default, where `comp-op` is not supported:
https://github.com/behdad/cairo/blob/7a923b1ff5bf3e2b79ce4f2fe14d352faef5bccb/src/cairo-svg-surface.c#L276

SVG version can be upgraded to 1.2 by calling the following function on the surface:

```c++
cairo_svg_surface_restrict_to_version(surface, CAIRO_SVG_VERSION_1_2);
```

Partial solution it is because the SVG surface can be created outside of Mapnik. In such case responsibility for setting the SVG version is on the caller of Mapnik.

### DIFF
--- a/src/cairo_io.cpp
+++ b/src/cairo_io.cpp
@@ -82,6 +82,7 @@ void save_to_cairo_file(mapnik::Map const& map,
         else if (type == "svg")
         {
             surface = cairo_surface_ptr(cairo_svg_surface_create(filename.c_str(),width,height),cairo_surface_closer());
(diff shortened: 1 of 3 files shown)


hholzgra:
Interesting ... I'll try that with my MapOSMatic fork, as there the surface is created by the caller, not Mapnik, indeed.

PR #4039, opened 2019-03-22
Ditch rint

### LINKED ISSUE #2392: ditch rint
We use `rint` in two places: gdal_featureset.cpp and one line in value.hpp. But we use `std::floor(<val> + .5)` many places. Need to review this situation to make sure we're consistently rounding.

mdholloway:
Removes all usages of rint from the codebase, in favor of the standard
rounding and casting formula used throughout the codebase:

  static_cast<int>(std::floor(<val> + .5))

Fixes #2392


artemp:
good catch, thanks @mdholloway

PR #4042, opened 2019-04-04
Address undefined behaviour in agg

Algunenano:
This fixes multiple warnings that appear when running under clang's undefined behaviour sanitizer, most of them related to integer overflows so most of the time I've chosen to cast to double when necessary (if it were using `int32_t` then I'd have chosen `int64_t`, but since as far as I can remember `int` doesn't have a guaranteed size It could happen that int == long int in some architectures).

To test I'm building with clang `8.0.0` and:
```
  python scons/scons.py configure \
    PREFIX="/usr" \
    DESTDIR="$pkgdir" \
    INPUT_PLUGINS=all \
    ENABLE_STATS=False \
    CPP_TESTS=True \
    DEBUG=False \
    ENABLE_LOG=False \
    OPTIMIZATION=0 \
    CUSTOM_CXXFLAGS="-march=native -mtune=native -O0 -pipe -fno-plt -Wall -Wextra -g3 -gdwarf-4 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -fsanitize=undefined -fsanitize-undefined-trap-on-error" \
    CUSTOM_DEFINES="-DACCEPT_USE_OF_DEPRECATED_PROJ_API_H=1" \
    CC="clang" \
    CXX="clang++"
```

The vital part is the `-fsanitize=undefined -fsanitize-undefined-trap-on-error"` so the visual tests crash and generate a coredump and `OPTIMIZATION=0` so you get the exact point of the crash instead of only an aproximate line. I've also run this under gcc (8.2) after this patch is applied and found nothing.

I've tried to add a task in travis to build with ubsan but failed miserably, but it'd be nice to have it run automatically as done in node-mapnik. keep in mind that by using "trap on error" you don't need the ubsan library during runtime, which makes it simpler to build and test.

In my branch these changes required updating some reference images (https://github.com/CartoDB/test-data-visual/commit/804aa4b52e7bdda050f0c2fdcce62abca158b84a).

### DIFF
--- a/deps/agg/include/agg_basics.h
+++ b/deps/agg/include/agg_basics.h
@@ -16,7 +16,10 @@
 #ifndef AGG_BASICS_INCLUDED
 #define AGG_BASICS_INCLUDED
 
+#include <algorithm>
 #include <cmath>
+#include <limits>
+
 #include "agg_config.h"
 
 //---------------------------------------------------------AGG_CUSTOM_ALLOCATOR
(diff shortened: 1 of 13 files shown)


Algunenano:
Seems this crashes node-mapnik test in https://github.com/mapnik/node-mapnik/blame/master/test/data/vector_tile/errors.svg#L29

I'm looking int it, so please do not merge this yet.

Algunenano:
While I was investigating the issue described in my previous comment (it ended up being a shift of a negative number), I found several other places where the code was relying on undefined behaviour.

The issues found are, either detected with ubsan running the tests from mapnik and node-mapnik or by extrapolating fixes to found issues to other places that use a similar code.
For example, in 042093cf2229e0fed24abe78044f4d251972b415 I've addressed an issue that appeared after my initial modifications where negative int values where being shifted (which is undefined) and causing a crash in node-mapnik. A similar issue was addressed in the past in 6815ac28699a1b5b5d983e615f4271fee33e94d6.

These commits introduce a **breaking change** in the debug symbolizer (https://github.com/mapnik/mapnik/blob/0f08588b49030356ad676b6102c7f123c81fa8e5/src/agg/process_debug_symbolizer.cpp#L54) as now negative values in the bounding box limits are clipped to 0. Before the change it was using `int` and implicitly casting to `size_t` (unsigned) in the call to `set_pixel`, which was undefined. You can see what this change meant in the reference images in https://github.com/CartoDB/test-data-visual/commit/c74c51ca2571270df097d4c02b1cbd7ab890be0f.

PR #4047, opened 2019-04-14
libxml2: add support for huge XMLs

### LINKED ISSUE #4038: Add XML_PARSE_HUGE for libxml2
Since version 2.7.3 libxml2 has a maximum size of a single textnode by 10 MB.
There is a Parser-Flag `XML_PARSE_HUGE`, which will relax hardcoded limits.

Would like to add `XML_PARSE_HUGE` to
https://github.com/mapnik/mapnik/blob/46478d4b5a17e4a66bd81af59175e21c493712af/src/libxml2_loader.cpp#L42

What do you think?

tarvip:
Decided to add another options as well as it is also related to big XMLs.

`XML_PARSE_HUGE` relaxes any hardcoded limit from the parser. This affects limits like maximum depth of a document or the entity recursion, as well as limits of the size of text nodes.

`XML_PARSE_BIG_LINES` adds support for big line numbers (bigger than 65535).

My C++ skill is not very good, maybe something like
```
#if LIBXML_VERSION > 20632
        options_ |= XML_PARSE_HUGE;
#endif

#if LIBXML_VERSION > 20800
        options_ |= XML_PARSE_BIG_LINES;
#endif
```

would be better?

Closes #4038


artemp:
thanks @tarvip !

PR #4058, opened 2019-05-17
New raster colorizer mode for Terrain-RGB

talaj:
Allows to easily generate Terrain-RGB from a raster:

```xml
<RasterSymbolizer scaling="bicubic">
    <RasterColorizer default-mode="linear-bgra">
        <stop value="-10000" color="#000000"/>
        <stop value="1667721.5" color="#FFFFFF"/>
    </RasterColorizer>
</RasterSymbolizer>
```
These stops are computed from `height = -10000 + ((R * 256 * 256 + G * 256 + B) * 0.1)`.

The `linear-rgba` was added for complementarity, but has no application curretly.


artemp:
:+1:

PR #4072, opened 2019-07-01
Replace MAPNIK_INIT_PRIORITY workaround

Kojoley:
Spirit rules are just placeholders. They are cheap to construct and by doing this static initialization order fiasco problem is effectively avoided.

Also had to fix name clash between geojson and generic json parsers.


artemp:
@Kojoley - thanks!

PR #4085, opened 2019-09-19
Revamp mapnik-config generation

lightmare:
`utils/mapnik-config/mapnik-config` script consisted of two parts: the first part was a format string in `utils/mapnik-config/build.py`, interpolated with values from a dictionary; the second part was literal bash script in `utils/mapnik-config/mapnik-config.template.sh`.

This PR changes how `mapnik-config` script is generated:
- the whole template is moved into `mapnik-config.template.sh`
- string formatting `%(token)s` in the template replaced with default values, and instead of interpolating `config_template % config_dict`, use regexp substitution to insert SCons environment variables into shell variable assignments in the template
- fixes bug in `CONFIG_GIT_{REVISION,DESCRIBE}` interpolation with Python 3: because `Popen.communicate` returns `bytes` by default, and `str(b'foo') == "b'foo'"`, resulting strings had an extra `b` prepended
- fixes bug in `--prefix` path calculation: `cd $(dirname $(dirname $0)) && pwd` was incorrect if you `cd` into prefix/bin dir and run `./mapnik-config`
- reverts two old commits prepending `${NODE_CONFIG_PREFIX}` to include paths; if this extra prefix is actually needed, it should be added via `${parameter//pattern/substitution}` in the script template, because `build.py` now shell-escapes all substituted values

### DIFF
--- a/utils/mapnik-config/build.py
+++ b/utils/mapnik-config/build.py
@@ -87,22 +87,17 @@ template_env['CXX'] = re.sub(r'^ccache +', '', cxx_cleaned)
 
 template_env['DEFINES'] = ' '.join(config_env['LIBMAPNIK_DEFINES'])
 
-dep_includes = ''.join([' -I%s' % i for i in config_env['CPPPATH'] if not i.startswith('#')])
-
-dep_includes += ' '
-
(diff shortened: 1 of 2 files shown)


artemp:
:+1:

PR #4094, opened 2019-10-21
Boost-1.71

artemp:
NOTE: requires patching boost_1_71 with latest bug fix https://github.com/boostorg/spirit/pull/549/commits/10d027fd5cfd573471e80e2e3c2d0cb0616a6d65

### DIFF
--- a/include/mapnik/json/feature_generator_grammar.hpp
+++ b/include/mapnik/json/feature_generator_grammar.hpp
@@ -63,6 +63,17 @@ struct end_container<mapnik::feature_impl const>
     }
 };
 
+template <>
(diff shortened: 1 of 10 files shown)


include/mapnik/json/feature_generator_grammar.hpp
```diff
@@ -63,6 +63,18 @@ struct end_container<mapnik::feature_impl const>
     }
 };
 
+#if BOOST_VERSION >= 106900
```
manthey: Should this be `>= 107000`?  As it stands, building with Boost 1.69.0 fails for me.

artemp: Hmm.. I just checked and it works for me on linux+clang. Could you share your setup, pls?

PR #4096, opened 2019-11-01
avoid potential out-of-bounds array access (undefined behaviour)

artemp:
Also, add  c++ `C-array` size implementation


src/text/scrptrun.cpp
```diff
@@ -156,7 +161,8 @@ UBool ScriptRun::next()
         // characters above it on the stack will be poped.
         if (pairIndex >= 0) {
             if ((pairIndex & 1) == 0) {
-                parenStack[++parenSP].pairIndex = pairIndex;
+                parenSP = (++parenSP) & STACK_MASK; // avoid out-of-bounds access
```
ericfischer: This has a preincrement and an assignment in the same expression. I think this instead needs to be

```
parenSP = (parenSP + 1) & STACK_MASK;
```

With this change, it looks like this should prevent the crash, since the corresponding code that decrements `parenSP` checks for 0 instead of blindly decrementing.

It will still have strange behavior when punctuation is nested this deeply, since the punctuation will no longer be treated as paired, but incorrect rendering is a lot better than crashing. I don't know if it would be practical from a performance perspective to use a dynamically-allocated vector for the stack instead of an array, but if it is fast enough, that would ultimately be better.

artemp: > This has a preincrement and an assignment in the same expression. I think this instead needs to be
> 
> ```
> parenSP = (parenSP + 1) & STACK_MASK;
> ```
yep, fixed, cheers!

> With this change, it looks like this should prevent the crash, since the corresponding code that decrements `parenSP` checks for 0 instead of blindly decrementing.
> 
> It will still have strange behavior when punctuation is nested this deeply, since the punctuation will no longer be treated as paired, but incorrect rendering is a lot better than crashing. I don't know if it would be practical from a performance perspective to use a dynamically-allocated vector for the stack instead of an array, but if it is fast enough, that would ultimately be better.

This is rather unusual situation to have > 128 nested paired punctuation chars:). Also I thought about not diverging too far from the original ICU code (for future references). But yes, I'm thinking about std::vector too.

test/unit/text/script_runs.cpp
```diff
@@ -0,0 +1,33 @@
...
+        size += runs.getScriptEnd() - runs.getScriptStart();
+        ++count;
+    }
+    REQUIRE(count == 7);
+    REQUIRE(size == text.length());
+}
+
+TEST_CASE("many punctuation chars")
+{
+    mapnik::value_unicode_string text("(((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((test"); // more than 128 "paired" chars
```
talaj:
```c++
mapnik::value_unicode_string text((std::string(791, '(') + "test").data());
```

talaj:
I have another crash case: `()()()x()`
In this case the array is being accessed with negative indices.
I fixed it by adding the following line:
```diff
diff --git a/src/text/scrptrun.cpp b/src/text/scrptrun.cpp
index 750998764..488e57651 100644
--- a/src/text/scrptrun.cpp
+++ b/src/text/scrptrun.cpp
@@ -158,6 +158,7 @@ UBool ScriptRun::next()
             if ((pairIndex & 1) == 0) {
                 parenStack[++parenSP].pairIndex = pairIndex;
                 parenStack[parenSP].scriptCode  = scriptCode;
+                startSP = parenSP;
             } else if (parenSP >= 0) {
                 int32_t pi = pairIndex & ~1;
```
Which makes this comment true:
https://github.com/mapnik/mapnik/blob/0842d0af6fb766e1dc242584b0035af3a6bc3d74/src/text/scrptrun.cpp#L120

I haven't tested it yet with your improvements. It probably wont crash anymore, but that assigment should be there in my opinion.

artemp:
> I have another crash case: `()()()x()`
> In this case the array is being accessed with negative indices.
> I fixed it by adding the following line:
> 
> ```diff
> diff --git a/src/text/scrptrun.cpp b/src/text/scrptrun.cpp
> index 750998764..488e57651 100644
> --- a/src/text/scrptrun.cpp
> +++ b/src/text/scrptrun.cpp
> @@ -158,6 +158,7 @@ UBool ScriptRun::next()
>              if ((pairIndex & 1) == 0) {
>                  parenStack[++parenSP].pairIndex = pairIndex;
>                  parenStack[parenSP].scriptCode  = scriptCode;
> +                startSP = parenSP;
>              } else if (parenSP >= 0) {
>                  int32_t pi = pairIndex & ~1;
> ```
> 
> Which makes this comment true:
> 
> https://github.com/mapnik/mapnik/blob/0842d0af6fb766e1dc242584b0035af3a6bc3d74/src/text/scrptrun.cpp#L120
> 
> I haven't tested it yet with your improvements. It probably wont crash anymore, but that assigment should be there in my opinion.

Thanks @talaj! I'll test and add your test (and fix if required) ASAP (tomorrow)

artemp:
@talaj - I applied your fix and added unit test, thanks a lot!

test/unit/text/script_runs.cpp
artemp: @talaj  :+1:

test/build.py
```diff
@@ -34,6 +35,7 @@
     # unit tests
     sources = glob.glob('./unit/*/*.cpp')
     sources.extend(glob.glob('./unit/*.cpp'))
+    sources.append('../src/text/scrptrun.cpp')
```
springmeyer: Why is this needed? Is it s visibility thing? In that case shouldn't the header be modified so that the right functions are visible?

artemp: First I thought, `ScriptRun` is internal class so we shouldn't expose it. But now I've changed my mind 9e82006314da0997c576fcdffbd2777b68ec38a0 (one less file to compile)

include/mapnik/text/scrptrun.hpp
```diff
@@ -17,12 +17,16 @@
...
+#include <mapnik/config.hpp>
 #pragma GCC diagnostic push
 #include <mapnik/warning_ignore.hpp>
 #include <unicode/utypes.h>
 #include <unicode/uobject.h>
 #include <unicode/uscript.h>
 #pragma GCC diagnostic pop
+#include <vector>
+
+const unsigned int STACK_SIZE = 1 << 7; // 2^n
```
lightmare: this doesn't belong to global scope

src/text/scrptrun.cpp
```diff
@@ -156,8 +160,9 @@ UBool ScriptRun::next()
         // characters above it on the stack will be poped.
         if (pairIndex >= 0) {
             if ((pairIndex & 1) == 0) {
-                parenStack[++parenSP].pairIndex = pairIndex;
-                parenStack[parenSP].scriptCode  = scriptCode;
+                ++parenSP;
+                parenStack.emplace_back(pairIndex, scriptCode);
```
lightmare: This looks wrong. How do I construct a test case that'd fail because of this?
There is no corresponding `pop_back` anywhere in the code. Popping or clearing the stack is done by simply decreasing `parenSP`, but the vector maintains its size, and so items emplaced after a pop will not be at position `parenSP`. Either there need to be actual vector pops everywhere `parenSP` decreases (and consequently `parenSP` made redundant as it will always be equal to `parenStack.size() - 1`), or keep this simplistic raw-array-in-a-vector approach by only pushing to vector when `parenSP >= size()`
```c++
++parenSP;
if ((size_t)parenSP < parenStack.size())
    parenStack[parenSP] = { pairIndex, scriptCode };
else
    parenStack.emplace_back(pairIndex, scriptCode);
```

include/mapnik/text/scrptrun.hpp
lightmare: this doesn't belong to global scope (and is perhaps unnecessarily large)

src/text/scrptrun.cpp
artemp: Good catch ^ thanks @lightmare !

include/mapnik/text/scrptrun.hpp
artemp: What do you think is a reasonable starting size would be here @lightmare ? How about `16` ? Btw, this code including stack size is copy/pasted from ICU.

src/text/scrptrun.cpp
artemp: https://github.com/mapnik/mapnik/commit/9dac8dcc34f57f6529219b28f06d05185c24840b

include/mapnik/text/scrptrun.hpp
lightmare: I was just guessing they chose a large value to cover even crazy cases, because the size was fixed. Most text passing through mapnik are short labels, anyway, so 16 sounds reasonable to me.

Ad copy/paste: I noticed the file comment, was going to submit separate issue/PR. The link to license at icu-project.org is dead, now it's [at github](https://github.com/unicode-org/icu/blob/master/icu4c/LICENSE); mere link is not sufficient to comply with the license, though:
_"... provided that the above
copyright notice(s) and this permission notice appear in all copies of
the Software and that both the above copyright notice(s) and this
permission notice appear in supporting documentation."_

artemp: > I was just guessing they chose a large value to cover even crazy cases, because the size was fixed. Most text passing through mapnik are short labels, anyway, so 16 sounds reasonable to me.

sounds good to me too.

> Ad copy/paste: I noticed the file comment, was going to submit separate issue/PR. The link to license at icu-project.org is dead, now it's [at github](https://github.com/unicode-org/icu/blob/master/icu4c/LICENSE); mere link is not sufficient to comply with the license, though:
> _"... provided that the above copyright notice(s) and this permission notice appear in all copies of the Software and that both the above copyright notice(s) and this permission notice appear in supporting documentation."_

yes, pls ^

PR #4113, opened 2020-01-21
Svg fixes [WIP]

artemp:
Fixing various SVG rendering/parsing glitches

### DIFF
--- a/utils/svg2png/svg2png.cpp
+++ b/utils/svg2png/svg2png.cpp
@@ -54,9 +54,11 @@
 struct main_marker_visitor
 {
     main_marker_visitor(std::string const& svg_name,
+                        double scale_factor,
                         bool verbose,
                         bool auto_open)
         : svg_name_(svg_name),
+          scale_factor_(scale_factor),
           verbose_(verbose),
           auto_open_(auto_open) {}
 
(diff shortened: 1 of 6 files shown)


utils/svg2png/svg2png.cpp
```diff
@@ -71,17 +73,34 @@ struct main_marker_visitor
         double opacity = 1;
         double w, h;
         std::tie(w, h) = marker.dimensions();
+
+
+        double svg_width = w * scale_factor_ + 0.5;
+        double svg_height = h * scale_factor_ + 0.5;
```
lightmare: adding 0.5 here seems strange, see below...

utils/svg2png/svg2png.cpp
```diff
@@ -71,17 +73,34 @@ struct main_marker_visitor
...
 
-        mapnik::box2d<double> const& bbox = {0, 0, w, h};
-        agg::trans_affine mtx = {};
+        mapnik::box2d<double> const& bbox = {0, 0, svg_width, svg_height};
+        // center the svg marker on '0,0'
+        agg::trans_affine mtx = agg::trans_affine_translation(-0.5 * w, -0.5 * h);
+        // Scale the image
+        mtx.scale(scale_factor_);
+        // render the marker at the center of the marker box
+        mtx.translate(0.5 * svg_width, 0.5 * svg_height);
```
lightmare: If I'm reading this correctly, this matrix results in scaling and translation by 0.25px: it shifts by half width, then scales, then shifts back by half scaled width. If svg_width didn't have 0.5 added to it, the matrix would be equivalent to just scaling.

artemp: I copied it from https://github.com/mapnik/node-mapnik/blob/master/src/mapnik_image.cpp#L2833-L2838  as I wanted exactly the same behaviour to start with. But, you're right it seems redundant. Great to have your eyes looking at this, cheers!

utils/svg2png/svg2png.cpp
artemp: The idea is to have output image in pixels with dimensions equal or slightly large then size reported by SVG which is floating point numbers.

utils/svg2png/svg2png.cpp
artemp: I created an issue to look into this #4116

utils/svg2png/svg2png.cpp
lightmare: I should've put emphasis on *here*. Was wondering primarily why 0.5 was added *here* to the floating-point scaled width, rather than only inside the rounding expression. But now with your explanation, it sounds like it was meant to be rounding up, instead of rounding to nearest:
```c++
double svg_width = w * scale_factor_;
int output_width = static_cast<int>(std::ceil(svg_width));
```

artemp: My comment disappeared, doh.. retyping:  Agree with moving up/down 0.5 scaling inside rounding expression.
std::ceil is not what others (checked on Ubuntu) are doing see details at #4116

Issue #4109, opened 2020-01-21
Fails to build with libxml2 2.9.10 (uses xml2-config)


sebastic:
As reported by Mattia Rizzolo in [Debian Bug #949412](https://bugs.debian.org/949412):
> your package is using `xml2-config` to detect and use libxml2.  I'm
> removing that script, so please update your build system to use
> pkg-config instead.
>
> The libxml2 package in experimental already doesn't ship the xml2-config
> script.
>
> Attached is the full build log, hopefully relevant excerpt follows:
>
> ```
> Checking for xml2-config... no
> ...
> Exiting... the following required dependencies were not found:
>    - xml2-config (xml2-config program | try setting XML2_CONFIG SCons option or avoid the need for xml2-config command by configuring with XML2_LIBS & XML2_INCLUDES)
>    - libxml2 (libxml2 library | try setting XML2_CONFIG SCons option to point to location of xml2-config program or configure with XML2_LIBS & XML2_INCLUDES)[0m
>
> See '/build/mapnik-3.0.22+ds1/build-python/config.log' for details on possible problems.[0m
>
> Set custom paths to these libraries and header files on the command-line or in a file called 'config.py'[0m
>     ie. $ python scons/scons.py BOOST_INCLUDES=/usr/local/include BOOST_LIBS=/usr/local/lib[0m
>
> Once all required dependencies are found a local 'config.py' will be saved and then install:[0m
>     $ sudo python scons/scons.py install[0m
>
> To view available path variables:
>     $ python scons/scons.py --help or -h[0m
>
> To view overall SCons help options:
>     $ python scons/scons.py --help-options or -H
>
> More info: https://github.com/mapnik/mapnik/wiki/Mapnik-Installation[0m
> make[1]: *** [debian/rules:66: debian/stamps/configure-python] Error 1
> make[1]: Leaving directory '/build/mapnik-3.0.22+ds1'
> make: *** [debian/rules:53: build] Error 2
> dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
> ```

PR #4114, opened 2020-01-21
Use pkg-config for libxml2.

sebastic:
The changes from this PR have been included as a patch in the Debian package to use pkg-config for libxml 2.9.10+.

Fixes: #4109


artemp:
@sebastic - looks great, thanks!

Issue #4120, opened 2020-01-30
env.ParseFlags + SCons.Util.CLVar.__str__  


artemp:
env.ParseFlags() introduced in a25f53e667c6e4ec123e3968e097804e0d847987 is not handling `CUSTOM_LDFLAGS` as expected. Consider `config.py` entry on macOS 

```
CUSTOM_LDFLAGS = '-Wl,-search_paths_first -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk -arch x86_64 -mmacosx-version-min=10.8'
```
This gets parsed as 
```
{'ASFLAGS': [], 'CFLAGS': [], 'CCFLAGS': ['-stdlib=libc++', ('-isysroot', '/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk'), ('-arch', 'x86_64'), '-mmacosx-version-min=10.8'], 'CXXFLAGS': [], 'CPPDEFINES': [], 'CPPFLAGS': [], 'CPPPATH': [], 'FRAMEWORKPATH': [], 'FRAMEWORKS': [], 'LIBPATH': [], 'LIBS': [], 'LINKFLAGS': ['-Wl,-search_paths_first', ('-isysroot', '/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk'), ('-arch', 'x86_64')], 'RPATH': []}
```
Note, `LINKFLAGS` is a mixed list of strings and tuples. This in return causes `SCons.Util.CLVar.__str_` method to fail with `TypeError: sequence item 3: expected string, tuple found:` (https://travis-ci.org/mapbox/mason/jobs/643495562?utm_medium=notification&utm_source=github_status) 
The `__str__` method is implemented as: 

```python
def __str__(self):
        return ' '.join(self.data)
``` 
and doesn't expect tuples in `self.data` 

/cc @lightmare @springmeyer

lightmare:
The problem is not in the use of `env.ParseFlags`, SCons uses that internally and interpolates those tuples into strings just fine. The culprit is this naive conversion in `write_config`:
https://github.com/mapnik/mapnik/blob/a3086c06a5f01ee30c078513b4072c337fa08643/utils/mapnik-config/build.py#L66

[...]

PR #4121, opened 2020-01-30
mapnik-config: fix substitution of environment variables

lightmare:
Refs #4120


artemp:
@lightmare - thanks for the speedy fix!

PR #4122, opened 2020-01-30
scons: prevent duplication of linker flags

lightmare:
Refs #4120 

There's a minor issue with flags that `env.ParseFlags` appends to both `CCFLAGS` and `LINKFLAGS`. Simply concatenating those lists then duplicates such flags (e.g. `-arch`).
https://github.com/mapnik/mapnik/blob/5732df452cda20b42c26f1946e3b770b142f403e/scons/scons-local-3.0.1/SCons/Environment.py#L722-L725


artemp:
:+1:

PR #4128, opened 2020-02-27
Update SConstruct to fix osx build

dirkvdb:
Fixed framework presence check to avoid TypeError:
TypeError: a bytes-like object is required, not 'str':


lightmare:
I wonder why `value` is not a `str` in the first place. It's coming from a function that is supposed to return a `str`.

PR #4132, opened 2020-03-23
add support for spacing-offset

zdila:
This pull request adds support for offset for `spacing` attribute of `marker-symbolizer`. It is useful for putting two marker symbolizers on the same line where one is shifted. Example:

image

Original question is here: https://github.com/mapnik/mapnik-support/issues/137

Please help me with adding tests and/or documentation.


artemp:
@zdila - thanks!

PR #4159, opened 2020-07-13
Stop setting SourceCode attribute

tomhughes:
SCons 4 has dropped dropped support for the SourceCode attribute.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -141,7 +141,6 @@ PLUGINS = { # plugins with external dependencies
 
 def init_environment(env):
     env.Decider('MD5-timestamp')
-    env.SourceCode(".", None)
     env['ORIGIN'] = Literal('$ORIGIN')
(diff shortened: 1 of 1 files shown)


lightmare:
Not knowing what it was for, I found this note in [Scons 3 doc](https://www.scons.org/doc/3.1.2/HTML/scons-man.html#f-SourceCode):
>  scons will, by default, fetch files from SCCS or RCS subdirectories without explicit configuration. This takes some extra processing time to search for the necessary source code management files on disk. You can avoid these extra searches and speed up your build a little by disabling these searches as follows:
> 
> env.SourceCode('.', None)

I never knew there was such a default. Is that extra search insignificant, or would it be more appropriate to keep fallback for SCons 3? Something like:
```py
if getattr(env, 'SourceCode'): env.SourceCode('.', None)
```

PR #4202, opened 2021-01-15
Upgrade to new Proj APIs

artemp:
https://proj.org/development/migration.html 

#### TODO: 
- [x] - Fix performance issues with proj7 initialisation (sqlite3) 
       * Addressed by caching `proj_transform` objects in https://github.com/mapnik/mapnik/commit/5c086b0cd5140e701824a744b958e81055527e65
- [x] - Fix `PROJ_LIB` detection logic (fails on travis-ci) 
- [x] - Add minimum proj version (>=7.0.0) check in `Sconstruct`
- [x] - Add support for proj-8 (define `PROJ_VERSION_MAJOR` at configure time (?))  
- [x] - test python-mapnik WIP 
- [x] - test node-mapnik WIP


include/mapnik/map.hpp
```diff
@@ -277,11 +271,6 @@ class MAPNIK_DECL Map : boost::equality_comparable<Map>
      */
     std::vector<layer> const& layers() const;
 
-    /*! \brief Get all layers.
-     *  @return Non-constant layers.
-     */
-    std::vector<layer> & layers();
-
```
tomhughes, line 283: Removing this method breaks `demo/viewer` as it makes at least two calls to this, at least one of which is actually trying to modify the layer list...

tomhughes:
Is the python-mapnik WIP mentioned above available somewhere? Only I don't see a branch or PR for it in https://github.com/mapnik/python-mapnik?

include/mapnik/map.hpp
artemp: You're right, removing non-const access to layers is problematic in viewer (same goes for Python bindings). I'm going to look into this again. Thanks for highlighting this issue.

artemp:
> Is the python-mapnik WIP mentioned above available somewhere? Only I don't see a branch or PR for it in https://github.com/mapnik/python-mapnik? 
As above, removing non-const access breaks expected behaviour in python bindings. Going to ponder on this.

artemp:
Ok, latest implementation using `thread_local` storage for proj_transform cache is looking good to me. 
@tomhughes @lightmare @springmeyer

tomhughes:
I backported the latest changes from this branch into the Fedora build of 3.1.0 and that seems to have been successful.

What I do still need is the python-mapnik changes so I can get that building again in Fedora.

artemp:
> I backported the latest changes from this branch into the Fedora build of 3.1.0 and that seems to have been successful.
> 
> What I do still need is the python-mapnik changes so I can get that building again in Fedora.

https://github.com/mapnik/python-mapnik/tree/proj6 - first cut, some tests need updating to new `libproj` syntax 

NOTE: it's important to initialise SRS with `epsg:4326` or 'epsg:3857` syntax in order to fallback on internal geographic WGS84 and "webmercator" implementations.  What version of `libproj` are you testing with? I added requirement for `7.2.0` as PROJ_MIN_VERSION but anything >= 6.0.0 should work. In principal. But I didn't have chance to test.

src/well_known_srs.cpp
```diff
@@ -36,57 +36,41 @@ MAPNIK_DISABLE_WARNING_POP
...
-                          (srs.find("+proj=latlon") != std::string::npos)
-                         )
-        {
-            return boost::optional<bool>(true);
-        }
-        else
-        {
-            return boost::optional<bool>(false);
-        }
+        return boost::optional<bool>(true);
```
lightmare, line 73: @artemp 
This was previously false. Looks like a typo caused by swapping the order of checks (was webmercator first).

artemp: @lightmare great catch as usual! I'll fix it and add unit test, thanks.

artemp: Fixed in 3a302667a1caf6831b6dc96c14e0729debbc1286 
Test added in 2bb5bdb9fe9e4f38e52f664e6cfda4dae8d94551

tomhughes:
With proj 8 there is a new failure:

```
In file included from src/projection.cpp:33:
/usr/include/proj.h:345:23: error: conflicting declaration ‘typedef struct pj_ctx PJ_CONTEXT’
  345 | typedef struct pj_ctx PJ_CONTEXT;
      |                       ^~~~~~~~~~
In file included from src/projection.cpp:24:
include/mapnik/projection.hpp:46:7: note: previous declaration as ‘using PJ_CONTEXT = struct projCtx_t’
   46 | using PJ_CONTEXT = struct projCtx_t;
      |       ^~~~~~~~~~
```

It seems that proj 8 is already defining `PJ_CONTEXT` itself...

tomhughes:
Ah no the problem is the header relies on PROJ_MAJOR_VERSION but proj.h is not included yet, but that's kind of the whole point when it's doing a forward declaration...

artemp:
> Ah no the problem is the header relies on PROJ_MAJOR_VERSION but proj.h is not included yet, but that's kind of the whole point when it's doing a forward declaration...

Oh doh, I think I forgot to fix this. SConstruct is setting `PROJ_VERSION`  https://github.com/mapnik/mapnik/blob/proj6/SConstruct#L1723  which should be used to figure out if proj >=8 or not, thanks

tomhughes:
Yes I hadn't backported your most recent changes yet - just testing that now.

mathisloge:
> Ok, latest implementation using `thread_local` storage for proj_transform cache is looking good to me.

This would not work for windows dlls, since thread_local storage can't be used in exported classes. 
https://docs.microsoft.com/de-de/cpp/error-messages/compiler-errors-1/compiler-error-c2492?view=msvc-160

\cc @artemp

mathisloge:
maybe a pimpl solution would work here? 
define a wrapper class which holds the thread_local static cache and will be instantiated by map?

artemp:
> maybe a pimpl solution would work here?
> define a wrapper class which holds the thread_local static cache and will be instantiated by map?

I'm going to look into this ^ I think moving `proj_cache_` out of `mapnik::Map` and initialising in `map.cpp` might work. I'll need your help testing on Windows, though :)

artemp:
@mathisloge  https://github.com/mapnik/mapnik/commit/503b9c5bbfe0c9e07247629d8a45a1c1e62c34ca <-- Can VC++ handle this?

mathisloge:
yes. that works. 

currently qt5 is being built and so i can't test the viewer yet, but it compiles and tests are running.

artemp:
> yes. that works.
> 
> currently qt5 is being built and so i can't test the viewer yet, but the rest compiles and tests are running.

The real test would be to run shared map benchmark : 

```
./benchmark/out/test_rendering_shared_map --name="shared map"  --map=map.xml --iterations=10 --threads=4

```

mathisloge:
@artemp which map.xml do you usually use?

mathisloge:
I've runned the benchmark with the roads.xml. 
but I needed to change this in the csv plugin:

```diff
diff --git a/plugins/input/csv/csv_utils.cpp b/plugins/input/csv/csv_utils.cpp
index edae079a1..dc5eaf6df 100644
--- a/plugins/input/csv/csv_utils.cpp
+++ b/plugins/input/csv/csv_utils.cpp
@@ -233,7 +233,7 @@ struct ignore_case_equal_pred
 bool ignore_case_equal(std::string const& s0, std::string const& s1)
 {
     return std::equal(s0.begin(), s0.end(),
-                      s1.begin(), ignore_case_equal_pred());
+                      s1.begin(), s1.end(), ignore_case_equal_pred());
 }

 void csv_file_parser::add_feature(mapnik::value_integer, mapnik::csv_line const & )
```

since this would fail if s0 is greater then s1

artemp:
> @artemp which map.xml do you usually use?
I'm using `map.xml` generated by './demo/c++/rundemo' . Or `mapnik-demo` with CMake build.

artemp:
> I've runned the benchmark with the roads.xml.
> but I needed to change this in the csv plugin:
> 
> ```diff
> diff --git a/plugins/input/csv/csv_utils.cpp b/plugins/input/csv/csv_utils.cpp
> index edae079a1..dc5eaf6df 100644
> --- a/plugins/input/csv/csv_utils.cpp
> +++ b/plugins/input/csv/csv_utils.cpp
> @@ -233,7 +233,7 @@ struct ignore_case_equal_pred
>  bool ignore_case_equal(std::string const& s0, std::string const& s1)
>  {
>      return std::equal(s0.begin(), s0.end(),
> -                      s1.begin(), ignore_case_equal_pred());
> +                      s1.begin(), s1.end(), ignore_case_equal_pred());
>  }
> 
>  void csv_file_parser::add_feature(mapnik::value_integer, mapnik::csv_line const & )
> ```
> 
> since this would fail if s0 is greater then s1

Interesting, looks like a bug. I think we need an issue so it doesn't get lost, could you create one or let me know and I'll do it, cheers

mathisloge:
Results with map.xml from mapnik-demo
```
"D:\dev\mapnik\build\demo\c++>"D:\dev\mapnik\build\benchmark\mapnik-benchmark-test_rendering_shared_map.exe" --name="shared map" --map="D:\dev\mapnik\build\demo\c++\map.xml" --threads=4 --iterations=100 
Defaulting to max extent box2d(-10608803.2816615924239159,4028651.6142304763197899,-4323716.1968731693923473,10313738.6990188993513584)
    (pass --extent=<minx,miny,maxx,maxy> to restrict bounds)
Loaded 'C:\Windows\System32\kernel.appcore.dll'. 
"shared map"                                  4 workers    10 iters   3098 milliseconds      3 i/t/s
```

artemp:
> Results with map.xml from mapnik-demo
> 
> ```
> "D:\dev\mapnik\build\demo\c++>"D:\dev\mapnik\build\benchmark\mapnik-benchmark-test_rendering_shared_map.exe" --name="shared map" --map="D:\dev\mapnik\build\demo\c++\map.xml" --threads=4 --iterations=10 
> Defaulting to max extent box2d(-10608803.2816615924239159,4028651.6142304763197899,-4323716.1968731693923473,10313738.6990188993513584)
>     (pass --extent=<minx,miny,maxx,maxy> to restrict bounds)
> Loaded 'C:\Windows\System32\kernel.appcore.dll'. 
> "shared map"                                  4 workers    10 iters   3098 milliseconds      3 i/t/s
> ```

Looks as expected. As long as this test doesn't segfault under stress with many threads running in parallel - all is good

artemp:
I'm planning to merge this branch into master and continue on any remaining issues with `python-mapnik` and 'node-mapnik` bindings 
/cc @tomhughes @lightmare @mathisloge

PR #4191, opened 2020-11-20
CMake build support

mathisloge:
closes #4190 

This PR would add CMake support to mapnik

Remaining work: 

- [x] copy all plugins and fonts in appropiate folders when building  
- [x] windows: copy plugin dlls and dependency dlls into executable dir
- [x] install: check if all targets are installed correctly. 
- [x] install, windows: do we install dependency dlls too? or do we let the user handle that? 
- [x] check visual studio cmake integration
- [x] set name from mapnik to libmapnik and append `libmapnik-` to wkt and json
- [x] add option for MAPNIK_THREADSAFE (also add `mt` to lib name) and set then Boost_USE_MULTITHREADED
- [x] ↑ should `mt` be added to the library name `libmapnik.[so]/[.dll]` when using MAPNIK_THREADSAFE? eg. `libmapnik-mt.[so]/[.dll]`
- [x] add MAPNIK_NO_ATEXIT. (what is the default value?)
- [x] add MAPNIK_NO_DLCLOSE (what is the default value?)
- [x] add MAPNIK_ENABLE_GLIBC_WORKAROUND (what is the default value?)
- [x] add MAPNIK_DEBUG (option will be false by default. But if using the CMakePresets (debug config) will set this to ON)
- [x] add option for MAPNIK_LOG and MAPNIK_DEFAULT_LOG_SEVERITY
- [x] rewrite FindWebP 
- [x] add MAPNIK_STATS

PR #4220, opened 2021-04-06
Simplify proj_transform_cache implementation 

artemp:
https://github.com/mapnik/mapnik/pull/4191#issuecomment-813377082

### DIFF
--- a/include/mapnik/feature_style_processor_impl.hpp
+++ b/include/mapnik/feature_style_processor_impl.hpp
@@ -254,7 +254,7 @@ void feature_style_processor<Processor>::prepare_layer(layer_rendering_material 
     }
 
     processor_context_ptr current_ctx = ds->get_context(ctx_map);
-    proj_transform const* proj_trans_ptr = m_.get_proj_transform(mat.proj0_.params(), mat.proj1_.params());
+    proj_transform const* proj_trans_ptr = proj_transform_cache::get(mat.proj0_.params(), mat.proj1_.params());
     box2d<double> query_ext = extent; // unbuffered
     box2d<double> buffered_query_ext(query_ext);  // buffered
 
@@ -493,7 +493,7 @@ void feature_style_processor<Processor>::render_material(layer_rendering_materia
     layer const& lay = mat.lay_;
 
     std::vector<rule_cache> const & rule_caches = mat.rule_caches_;
-    proj_transform const* proj_trans_ptr = m_.get_proj_transform(mat.proj0_.params(), mat.proj1_.params());
+    proj_transform const* proj_trans_ptr = proj_transform_cache::get(mat.proj0_.params(), mat.proj1_.params());
     bool cache_features = lay.cache_features() && active_styles.size() > 1;
 
     datasource_ptr ds = lay.datasource();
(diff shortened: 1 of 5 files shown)


artemp:
> although I do have an idea which I think is better than thread_local
@lightmare  what is it?

mathisloge:
Have tested on windows. But I needed to move MAPNIK_DECL to the front. 
Normally both should work, but I think the recommended way is to write it before the return parameter. E.g. in this case it leads to a syntactically incorrect program.
```diff
diff --git a/include/mapnik/proj_transform_cache.hpp b/include/mapnik/proj_transform_cache.hpp
index 6bd8581fe..99486e0d5 100644
--- a/include/mapnik/proj_transform_cache.hpp
+++ b/include/mapnik/proj_transform_cache.hpp
@@ -31,8 +31,8 @@ namespace mapnik  {
 class proj_transform; // fwd decl
 namespace proj_transform_cache {

-void MAPNIK_DECL init(std::string const& source, std::string const& dest);
-proj_transform const* MAPNIK_DECL get(std::string const& source, std::string const& dest);
+MAPNIK_DECL void init(std::string const& source, std::string const& dest);
+MAPNIK_DECL proj_transform const* get(std::string const& source, std::string const& dest);

 } // namespace proj_transform_cache
 } // mamespace mapnik
```

But since these changes the demo/c++ application and the shared map benchmark are running smoothly. 

Note that the outputs are from a debug build.
```
Defaulting to max extent box2d(-10608803.2816615924239159,4028651.6142304763197899,-4323716.1968731693923473,10313738.6990188993513584)
    (pass --extent=<minx,miny,maxx,maxy> to restrict bounds)
"shared map"                                  4 workers    10 iters  19970 milliseconds      1 i/t/s
```

artemp:
:+1: moving MAPNIK_DECL to the front is fine. I'll update master, thanks

artemp:
6bab4e56fd591a530e6f020b9ad4fd1633823b40

lightmare:
@artemp 
My idea may very well be another case of over-engineering, as I was pondering whether thread local cache is actually good, or just a quick stepping stone. I also haven't thought it through entirely, wish I found some time to try and implement it to see how it performs.

The benefit of thread local cache -- that each thread has its own -- is also a drawback: you can't use values computed by other threads. If you have multi-threaded rendering, chances are they use some common transforms. And if you have one thread that loads the map, and then passes it to worker threads that render tiles, the initialization of transforms in the loading thread is kinda wasted.

So that was my motivation: how to avoid constructing the same `proj_transform` in each thread without giving up concurrent construction of different transforms. It obviously requires synchronized access to the cache, and another map for keeping track of transforms currently under construction. To keep it relatively simple, I'm not going for out-of-order-construction of transforms.

It goes like this: acquire lockA, search the transform cache, if found return. Otherwise acquire lockB, release lockA, search the under-construction hashmap. If it contains a future for the key, wait for it to fulfil (lock released in wait), then return its value. Otherwise add a new future to the map, release lockB, construct the transform, re-acquire lockA, add transform to the cache map, acquire lockB, release lockA, fulfil the future, remove it from map, return. It's also doable with a single lock, but then a thread can hold the lock for up to 3 lookups (one being insertion), and it's not much simpler, so I prefer 2 locks.

artemp:
@lightmare  thanks, looks like `thread_local` fits well here, unless there are some unforeseen issues. Avoiding dealing with locks is big plus I think.

Issue #4227, opened 2021-05-14
GeoJSON plugin has no option to read Feature.id from source


montzkie18:
Related to this issue: https://github.com/mapnik/node-mapnik/issues/707

I have a use case where I create VectorTiles from a GeoJSON source, using node-mapnik. 
Unfortunately mapnik's geojson datasource ignores `Feature.id` by design to assign an incremental id instead.

It would be great to have a flag to force the GeoJSON plugin to read the ids from the source instead, giving users the responsibility of making those ids unique.

The flag makes it opt in and disabled by default to maintain the current behavior.

YousefED:
Agree this would be useful!

PR #4228, opened 2021-05-14
Add flag for reading ids from geojson source

montzkie18:
GeoJSON: Expose `use_id_from_source` datasource parameter to load features with their original ids from the source. Should address #4227 

Things to note here:
* Added a separate Spirit X3 rule for features that have an ID. This means by default parsing without any flags set will continue to ignore `Feature.id`. But once the flag is set, it will throw an error for invalid ids (e.g. string, null, etc...)
* Used `feature.use_id_from_source()` to check which grammar to use. I figured it can also be used by other plugins if they want to make it as an option.
* All instances where this flag is required is set to false by default. It should not change behavior for downstream packages. 
* Since this is opt-in, I don't think it breaks the original design decision to ensure `Feature.id`s as unique internally.

Depends on this PR: https://github.com/mapnik/test-data/pull/23

### DIFF
--- a/include/mapnik/feature.hpp
+++ b/include/mapnik/feature.hpp
@@ -101,15 +101,17 @@ public:
     using cont_type = std::vector<value_type>;
     using iterator = feature_kv_iterator;
 
-    feature_impl(context_ptr const& ctx, mapnik::value_integer _id)
+    feature_impl(context_ptr const& ctx, mapnik::value_integer _id, bool _use_id_from_source = false)
         : id_(_id),
         ctx_(ctx),
         data_(ctx_->mapping_.size()),
         geom_(geometry::geometry_empty()),
-        raster_() {}
+        raster_(),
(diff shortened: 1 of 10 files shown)


montzkie18:
hi @artemp - any chance I can get a review for this one?

artemp:
> hi @artemp - any chance I can get a review for this one?

@montzkie18 - Yes, I'll find time asap. Thanks for your contribution!

artemp:
@montzkie18 - Could you check test data is updated, I'm getting 
```
could not get file mapping for ./test/data/json//feature-with-id.json
```
,thanks.

montzkie18:
Hi @artemp  - thanks for looking into this.

> Could you check test data is updated

The updated test data is pushed to this PR: https://github.com/mapnik/test-data/pull/23
I'm not sure how to best handle submodules update here without that PR getting merged first. 
Should I point the submodule in this PR temporarily to the forked repo and branch?

PR #4239, opened 2021-08-09
Fix bounding box in svg patterns

pitdicker:
In SVG patterns, Mapnik currently ignores the viewBox of the svg, and determines the bounding box by rendering all elements inside the pattern and looking at the extends. Those are then scaled and centered to the pattern tile size.

Looking through the issues and PRs this has caused quite some difficulties for https://github.com/gravitystorm/openstreetmap-carto (see https://github.com/gravitystorm/openstreetmap-carto/issues/2750), and is blocking the conversion to vector patterns and vector tile output (https://github.com/gravitystorm/openstreetmap-carto/issues/2045). Quite a few PRs in https://github.com/gravitystorm/openstreetmap-carto couldn't get merged or were backed out because of this issue (https://github.com/gravitystorm/openstreetmap-carto/pull/2759, https://github.com/gravitystorm/openstreetmap-carto/pull/2727). Some patterns were clipped and/or given a dummy geometry as a workaround (https://github.com/gravitystorm/openstreetmap-carto/pull/2748, https://github.com/gravitystorm/openstreetmap-carto/pull/2714, https://github.com/gravitystorm/openstreetmap-carto/pull/2747, https://github.com/gravitystorm/openstreetmap-carto/pull/2683, https://github.com/gravitystorm/openstreetmap-carto/pull/4059).

As I understand it, the only change necessary is to stop determining the bounding box by looking at the extend of the elements inside the svg, and just use the reported with and height. I must admit my testing was limited, but can confirm it fixes the problem for all patterns I locally have. It also doesn't break scaling when the svg contains a `viewBox` with different dimensions from its width or height.

### DIFF
--- a/src/marker_cache.cpp
+++ b/src/marker_cache.cpp
@@ -187,9 +187,7 @@ std::shared_ptr<mapnik::marker const> marker_cache::find(std::string const& uri,
                 }
             }
             //svg.arrange_orientations();
-            double lox,loy,hix,hiy;
-            svg.bounding_rect(&lox, &loy, &hix, &hiy);
-            marker_path->set_bounding_box(lox,loy,hix,hiy);
+            marker_path->set_bounding_box(0,0,svg.width(),svg.height());
             marker_path->set_dimensions(svg.width(),svg.height());
             if (update_cache)
             {
@@ -227,9 +225,7 @@ std::shared_ptr<mapnik::marker const> marker_cache::find(std::string const& uri,
                     }
                 }
                 //svg.arrange_orientations();
-                double lox,loy,hix,hiy;
-                svg.bounding_rect(&lox, &loy, &hix, &hiy);
-                marker_path->set_bounding_box(lox,loy,hix,hiy);
+                marker_path->set_bounding_box(0,0,svg.width(),svg.height());
                 marker_path->set_dimensions(svg.width(),svg.height());
                 if (update_cache)
                 {
--- a/src/marker_cache.cpp
+++ b/src/marker_cache.cpp
@@ -187,9 +187,8 @@ std::shared_ptr<mapnik::marker const> marker_cache::find(std::string const& uri,
                 }
             }
             //svg.arrange_orientations();
-            double width, height;
-            width = svg.width();
-            height = svg.height();
+            const double width = svg.width();
+            const double height = svg.height();
             marker_path->set_dimensions(width,height);
             if (width != 0 && height != 0) {
                 marker_path->set_bounding_box(0,0,width,height);
@@ -237,9 +236,8 @@ std::shared_ptr<mapnik::marker const> marker_cache::find(std::string const& uri,
                     }
                 }
                 //svg.arrange_orientations();
-                double width, height;
-                width = svg.width();
-                height = svg.height();
+                const double width = svg.width();
+                const double height = svg.height();
                 marker_path->set_dimensions(width,height);
                 if (width != 0 && height != 0) {
                     marker_path->set_bounding_box(0,0,width,height);
--- a/src/marker_cache.cpp
+++ b/src/marker_cache.cpp
@@ -187,8 +187,20 @@ std::shared_ptr<mapnik::marker const> marker_cache::find(std::string const& uri,
                 }
             }
             //svg.arrange_orientations();
-            marker_path->set_bounding_box(0,0,svg.width(),svg.height());
-            marker_path->set_dimensions(svg.width(),svg.height());
+            double width, height;
+            width = svg.width();
+            height = svg.height();
+            marker_path->set_dimensions(width,height);
+            if (width != 0 && height != 0) {
+                marker_path->set_bounding_box(0,0,width,height);
+            } else {
+                // falback in case the SVG has a zero width, height, or has
+                // those as percentages: determine the smallest rectangle
+                // in which all objects fit.
+                double lox,loy,hix,hiy;
+                svg.bounding_rect(&lox, &loy, &hix, &hiy);
+                marker_path->set_bounding_box(lox,loy,hix,hiy);
+            }
             if (update_cache)
             {
                 auto emplace_result = marker_cache_.emplace(uri,std::make_shared<mapnik::marker const>(mapnik::marker_svg(marker_path)));
@@ -225,8 +237,20 @@ std::shared_ptr<mapnik::marker const> marker_cache::find(std::string const& uri,
                     }
                 }
                 //svg.arrange_orientations();
-                marker_path->set_bounding_box(0,0,svg.width(),svg.height());
-                marker_path->set_dimensions(svg.width(),svg.height());
+                double width, height;
+                width = svg.width();
+                height = svg.height();
+                marker_path->set_dimensions(width,height);
+                if (width != 0 && height != 0) {
+                    marker_path->set_bounding_box(0,0,width,height);
+                } else {
+                    // falback in case the SVG has a zero width, height, or has
+                    // those as percentages: determine the smallest rectangle
+                    // in which all objects fit.
+                    double lox,loy,hix,hiy;
+                    svg.bounding_rect(&lox, &loy, &hix, &hiy);
+                    marker_path->set_bounding_box(lox,loy,hix,hiy);
+                }
                 if (update_cache)
                 {
                     auto emplace_result = marker_cache_.emplace(uri,std::make_shared<mapnik::marker const>(mapnik::marker_svg(marker_path)));
--- a/test/unit/svg/svg_parser_test.cpp
+++ b/test/unit/svg/svg_parser_test.cpp
@@ -847,4 +847,15 @@ TEST_CASE("SVG parser") {
         CHECK(bbox.width() == Approx(100));
         CHECK(bbox.height() == Approx(100));
     }
+
+    SECTION("SVG contents outside <viewBox>")
+    {
+        std::string svg_name("./test/data/svg/shape_outside_viewbox.svg");
+        std::shared_ptr<mapnik::marker const> marker = mapnik::marker_cache::instance().find(svg_name, false);
+        REQUIRE(marker);
+        REQUIRE(marker->is<mapnik::marker_svg>());
+        mapnik::marker_svg const& svg = mapnik::util::get<mapnik::marker_svg>(*marker);
+        auto bbox = svg.bounding_box();
+        REQUIRE(bbox == mapnik::box2d<double>(0, 0, 10, 10));
+    }
 }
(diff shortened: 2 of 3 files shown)


artemp:
@pitdicker Thanks for looking into this and providing context.  `svg.bounding_rect(...)` method is used (if I remember correctly) because some SVGs don't have `width` and/or `height` specified or quite often have `width="100% height="100%"`. In these cases calling `bounding_rect` (https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement/getBBox) is the only reasonable way to determine bounds for rasterization.  I agree, if `width` and `height` are present they should be used but having a fallback is important, too. 
Perhaps, something like https://github.com/mapnik/mapnik/blob/master/utils/svg2png/svg2png.cpp#L77-L88 ?

pitdicker:
@artemp Thank you for the explanation, that makes sense. I'll read up on how mapnik handles SVGs that have `width="100% height="100%"`, and add the fallback.

pitdicker:
Added the fallback. Didn't add extra logging, as there is already a warning on parsing the image in https://github.com/mapnik/mapnik/blob/master/src/svg/svg_parser.cpp#L973-L982.

artemp:
> Added the fallback. Didn't add extra logging, as there is already a warning on parsing the image in https://github.com/mapnik/mapnik/blob/master/src/svg/svg_parser.cpp#L973-L982.
Looks good, thanks. Could you add an unit test to capture this change, pls?

pitdicker:
Github now recognized the extra commit, and the PR is up in https://github.com/mapnik/test-data/pull/24.

pitdicker:
Thank you for the patience, I should have run the visual tests. And this is the kind of change that could cause problems that have to be communicated...

What is the best way to run those tests so that they are not all marked as having minuscule changes?

artemp:
> Thank you for the patience, I should have run the visual tests. And this is the kind of change that could cause problems that have to be communicated...

No worries :) Yes, this change requires careful approach and communication. 
> 
> What is the best way to run those tests so that they are not all marked as having minuscule changes?

There's no command option I can see - visual tests are expected to pixel match (`expected` <--> `actual` images) exactly. Having said that, number of different pixels is reported for each test so should be straightforward to change while developing. Let me know if you need help. You can list available options with  `./test/visual/run -h`

pitdicker:
I did an ugly workaround to run the visual tests: first run them on master and overwrite all images, and than compage the results of this branch. Also the tests for some reason run the installed copy of mapnik. Commands:
```
git checkout master
make
sudo make install
./test/visual/run -j16 -o
git checkout svg_patterns
make
sudo make install
./test/visual/run -j16
```

Failing tests, sorted by used svg:

#### circle-mm.svg
- marker-svg-circle-mm-width

`circle-mm.svg` relies on a width and height in mm corresponding to px in 90 DPI (once the default in Inkscape), while SVG now defines it as 96 DPI. Before this PR the test passed by accident.
It also has a viewBox of `0 0 24 0`. This should disable rendering the element, but I can imagine it is good for compatibility that mapnik still renders it somehow (currently by forgetting all dimensions and using the fallback).

Can be fixed by using a width and height that rely on 96 DPI, and removing the wrong viewBox.

#### shield-multiple-command-parts.svg
- shield-svg-multiple-command-parts

The shield shape inside this svg is 36 high, while the height of the SVG is 40. The `shield-svg-multiple-command-parts` test also defines the height to 40. Before this PR the height of the SVG was ignored, and 36 was detected. Than the image was stretched to be 40 high (breaking the aspect ratio) because of the size in the test.

Can be fixed by setting the height of the SVG to 36.

#### airfield-12-nobox.svg
- marker-on-line-svg-with-box
- marker-on-line-angled-point
- marker-on-line-svg-no-box

The plane shape in `airfield-12-nobox.svg` is 9 wide instead of 12 as defined on the svg. Easiest solution is to remove all dimensions on the svg (or set them to percentages), so it uses the fallback.

#### octocat.svg
- marker-svg-octocat-test
- polygon-pattern-alignment-global
- line-pattern-smooth-and-offset
- polygon-pattern-alignment-local-clip
- geometry-transform-scale-patterns
- functional-expressions
- marker-direction
- polygon-pattern-alignment-local
- geometry-transform-translate-patterns-svg
- functional-expressions-length

The shapes in `octocat.svg` are a very little bit smaller than the viewBox. Before this PR it was stretched to the width and height in the viewBox (breaking the aspect ratio).

As the difference is small, it seems best to update the reference images. I couldn't find an easy way to adjust the viewBox, width and height to produce identical results.

#### transparent_circle.svg
- marker-vs-point
Again in the existing reference image the width and height of the SVG weren't used. The calculated size ignored the stroke-width of the circle, scaling them so they stick outside the red square.

It is probably best to update the reference image, so all circles fit within their squares. Maybe also adjust the width and height in the `marker-vs-point` to 76 so the image is symmetrical again.

The SVG-files are changed in https://github.com/mapnik/test-data/pull/24, but I can't update the reference images as I can't reproduce the current ones pixel-perfect yet.

pitdicker:
The marker-vs-point test showed there is  another problem (which I should have noticed before): an svg currently doesn't get clipped to it's bounding box. Would there be anything against adding a clipping rectangle in cairo and agg?

artemp:
@pitdicker - thanks for detailed info, I'll re-read and reply on case by case, hopefully tomorrow.

artemp:
I'm testing SVGs in question (unmodified) by rasterizing them with `inkscape` and `svg2png` e.g

```
inkscape airfield-12-nobox.svg -o airfield-12-nobox-inkscape.png
```
```
./utils/svg2png/svg2png -v  test/data/svg/airfield-12-nobox.svg 
```

I'm getting consistent visual results and matching sizes (96dpi) 

|SVG| Inkscape | svg2png | image size |
--- | --- | --- | --- |
|airfield-12-nobox.svg |airfield-12-nobox-inkscape|airfield-12-nobox | 12x12 |
|circle-mm.svg| circle-mm-inkscape|circle-mm| 51x51|
|octocat.svg |octocat-inkscape|octocat| 379x334|
|shield-multiple-command-parts.svg | shield-multiple-command-parts-inkscape |shield-multiple-command-parts| 36x40|

Could you try rasterising updated (as per your suggestion) SVGs and verify `inkscape` and `svg2png` producing matching images?

I'm trying to figure out if updating references or modifying SVGs is the best approach here. Cheers.

src/marker_cache.cpp
```diff
@@ -187,10 +187,20 @@ std::shared_ptr<mapnik::marker const> marker_cache::find(std::string const& uri,
                 }
             }
             //svg.arrange_orientations();
-            double lox,loy,hix,hiy;
-            svg.bounding_rect(&lox, &loy, &hix, &hiy);
-            marker_path->set_bounding_box(lox,loy,hix,hiy);
-            marker_path->set_dimensions(svg.width(),svg.height());
+            double width, height;
+            width = svg.width();
```
mathisloge, line 191: could this be `const width = ...; const height = ...;` ? 
I do not see any other changes and in my opinion this would be better.

(And same below.)

artemp: :+1:

pitdicker:
Sorry for the slow reply.

> I'm trying to figure out if updating references or modifying SVGs is the best approach here.

Let me try to write a bit better why I think some SVGs should be modified or not... (in multiple comments)

#### circle-mm.svg
The width and height in the test file were 48×25.4/90=13.546667mm.
This should be 48×25.4/96=12.7mm, because the svg specification defines the DPI to be 96.
This test only passed by accident, because the width and height were ignored.

| Inkscape | svg2png | svg2png current | image size |
|------------|----------|-------------------|-------------|
| circle-mm_new_inkscape  | circle-mm_new |  circle-mm_master | 48x48 (was 51x51) |

src/marker_cache.cpp
pitdicker: Thank you, I never leaned C++ properly.

pitdicker:
While looking again at airfield-12-nobox.svg I noticed `svg2png` doesn't use the x and y of the bounding box in the fallback path. Fixed to keep it consistent with the rendering in mapnik.

pitdicker:
#### airfield-12-nobox.svg
The difference between airfield-12.svg and airfield-12-nobox.svg is that the first one has an invisible square. The existance of these two tests shows that ignoring the width and height of an svg is somewhat intentional, with the square being a hacky way to set dimensions.

To get the same behavior while following the SVG spec, airfield-12.svg should remain as it is with a width and height. The fake rectangle can stay.
airfield-12-nobox.svg should now rely on the fallback, by including no dimensions, or setting the dimensions to 100%.

New airfield-12-nobox.svg rendered with the patched `svg2png`:
airfield-12-nobox
I rendered it at 10× and added a pink background to show the dimensions. Notice the partially transparent border around the plane, and how it is not included in the bounding box.

Inkscape does not really render it comparable. With missing or 100% dimensions, it seems to default to a viewBox of `0 0 300 150`.

pitdicker:
#### shield-multiple-command-parts.svg
Not sure what to say here. The SVG set its dimensions to 36x40px, and the test set them to 36x40px. But the dimensions of the content inside the SVG are only 36x36, cousing it to be stretched vertically .

| Inkscape | svg2png | svg2png current | stretched in test | image size |
|------------|----------|-------------------|-------------------|-----------|
| shield-multiple-command-parts_inkscape | shield-multiple-command-parts | shield-multiple-command-parts | shield-multiple-command-parts | 36x36 (was 36x40) |

(Added a pink background to show the dimensions.)

I don't think it really matters whether the SVG is changed, the reference image, or both. Going on the name, it is testing a complex path, not something involving dimensions.

To keep the test purely about the path, it would be better to only update the reference image to remove the stretched rendering.

pitdicker:
#### transparent_circle.svg
I did not intent to change this SVG. The size that used to be detected was 60x60px, while it should be 76x76, which also are the dimensions of the SVG.

Because strokes can extend beyond the bounding box, it was still rendered as 76x76, causing it to extend outside the red square in the reference tests.

| Inkscape | svg2png | before in test | image size |
|------------|----------|-------------------|-------------|
| transparent_circle_inkscape | transparent_circle_new | transparent_circle_auto | 76x76 (was 60x60)

I am not sure about the intention of the `marker-vs-point` test, but guess it is not this behavior.

pitdicker:
#### octocat.svg
Between the top of the image and its ears is 1px empty space, and between the blue circle and the bottom about 0.5px. Now that the the dimensions of the SVG are used instead of the contents, this leads to tiny differences in the reference images.

octocat

pitdicker:
Feel free to reopen when there is interest in improving svg support.

PR #4244, opened 2021-09-14
[CMake] make mapnik::core a public dependency of mapnik::mapnik

mathisloge:
This PR adds mostly a convenience to the target `mapnik::mapnik` so that it has a public dependency on `mapnik::core`. All targets which have a dependency of `mapnik::mapnik` (which is the most common use case) won't need to add `mapnik::core`.

PR tested and ready to merge.


artemp:
:+1:

Issue #4235, opened 2021-07-21
boost regression 


mathisloge:
This issue is pulled out of #4191 to split the discussion a bit.

The minimum boost version in SCons is 1.61. However in https://github.com/mapnik/mapnik/pull/4191#issuecomment-875053928 were build failures reported (with boost version  1.66.0) 

I've tested on Ubuntu 20.04 again and had linkage errors with boost 1.71. The latest boost version that compiled successfully was 1.73. 

Linkage error with boost 1.71: 
[boost-err.txt](https://github.com/mapnik/mapnik/files/6858320/boost-err.txt)

I think we should at least support 1.65, since this is the latest version on ubuntu 18.04

mathisloge:
#4211 is related

craigds:
I also was getting build failures with boost 1.65 on Ubuntu 18.04 LTS (the same failures mentioned in https://github.com/mapnik/mapnik/pull/4191#issuecomment-875053928 )

Using libboost1.74 from https://launchpad.net/~mhier/+archive/ubuntu/libboost-latest/ fixed this problem for me

craigds:
This issue makes it quite hard to build a python-mapnik on Ubuntu 18.04

I tried using a libboost1.74 from [this PPA](https://launchpad.net/~mhier/+archive/ubuntu/libboost-latest) which works for mapnik itself if you get the packaging right, but because there's no boost-python included there's no way to build python-mapnik with that.

b00ga:
For reference, I'm trying to build for RHEL/CentOS 8 and your packaged options are Boost 1.66 or a parallel-installable 1.69.

PR #4248, opened 2021-09-22
[CMake] Simplify install, add min versions and PkgConfig

mathisloge:
This PR mostly refactors the installation and should provide users with a much better experience.

Changed/Added: 
- Simplified install commands
- PROJ Include should not be present in the installed target (now as a build interface)
- Add components so that packagers could install only the runtime components. 
- All build utility programs are now exported as a target, so that these could directly be used in `execute_command` like calls. (without the need to call `find_program`)

Todos: 
- [x]  split `mapnik_install(target, options...)` to `mapnik_install(target)` and `mapnik_install_plugin(target)`
- [x] Check if the component installation works correctly
- [x] ~Explore ways so that the plugins are component targets for `find_package`. Blocker: currently there are not really nice ways to add the needed dependencies to the component targets. Currently just the installed plugin path gets published, this however could cause loading errors when using with a multiconfiguration enviroment.~ exporting the plugins as targets would be a bit of an overkill. However to address the multiconfiguration installations there is now a own export file for each configuration which has the correct plugin installation path. Added ad convinient function to get the path for consumers `mapnik_find_plugin_dir`. 
- [x] utility executables should be components for `find_package`
- [x] add *.pc files for pkg-config so that users don't have to use cmake when building their own projects. 
- [x] add Version and SOVersion to all base libraries (wkt, json and mapnik). ~How does the ABI combability looks? Windows would probably need the full version since we can't really reach a ABI stable interface (due to stl exports on interfaces). On linux however it should be the same major?~ => libname.so.MAJOR[.MINOR[.PATCH]]
- [x] test pkg-config

I think this PR should be squashed when merging.


mathisloge:
@artemp which ABI compability on linux/macos do you expect. Only major or major and minor?
So the resulting `mapnik.so` would be `mapnik.so.MAJOR[.MINOR]`

artemp:
> @artemp which ABI compability on linux/macos do you expect. Only major or major and minor?
> So the resulting `mapnik.so` would be `mapnik.so.MAJOR[.MINOR]`

`libname.so.MAJOR[.MINOR[.PATCH]]`  e.g `libmapnik.so.4.0.0`, `libboost_python38.so.1.76.0`

mathisloge:
adds basic pkg-config support. I still need to test it, since I'm normally not using pkg-config. 
The following is the generated `libmapnik.pc` file:
`libmapnikwkt` and `libmapnikjson` have their own *.pc files. 
```
prefix=D:/dev/install
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${exec_prefix}/lib

Name: libmapnik
Description: mapnik library
Version: 4.0.0
Requires: libmapnikwkt libmapnikjson
Libs: -L"${libdir}" -llibmapnik -llibmapnikwkt -llibmapnikjson
Cflags: -I"${includedir}" -DMAPNIK_THREADSAFE -DBIGINT -DMAPNIK_MEMORY_MAPPED_FILE -DHAVE_LIBXML2 -DHAVE_PNG -DHAVE_JPEG -DHAVE_TIFF -DHAVE_WEBP -DHAVE_CAIRO -DMAPNIK_USE_PROJ -DMAPNIK_PROJ_VERSION=70200 -DGRID_RENDERER -DSVG_RENDERER
```

mathisloge:
@artemp is the boost min version now 1.73 as in https://github.com/mapnik/mapnik/issues/4235 or do you plan to support boost 1.61 again?

mathisloge:
The pkg-config file is missing the other dependencies (so users currently have to manually add the dependency libraries to the build system). But the important things are getting exported (libmapnik, libmapnikwkt, libmapnikjson and the Cflags). 

So the currently minimum versions are
```
BOOST_MIN_VERSION 1.61
HARFBUZZ_MIN_VERSION 0.9.34
PROJ_MIN_VERSION 7.2.0
```

But at least for boost this isn't true currently, so we might change it in a different PR. 


TLDR:
@artemp PR is ready to merge.

mathisloge:
@artemp I would propose to discuss the boost min version in #4235

So ubuntu 20.04 which is the latest LTS release has boost 1.71 but that doesn't compile currently. But since mapnik v4 won't land in ubuntu 20.04 and ubuntu 21.04 has already 1.74 I think it would be fine to bump to 1.73

PR #4251, opened 2021-10-06
[CMake] fix `mapnik_find_plugin_dir`

mathisloge:
Sorry @artemp 
found a bug in the cmake export logic. My test project had the used variable `_build_config` defined, which normally does not exists. This fixes the bug.


artemp:
:+1:

PR #4262, opened 2021-10-25
[CI] add github workflow for tests

mathisloge:
This is a initial step towards testing with github actions. 

There are some needed steps. For the whole dependency management I'm using vcpkg. So most of the dependencies are very up to date. But this needs some setup steps. vcpkg is using binary caching via nuget to not build all dependencies each run. If we ever should use vcpkg in another repo, this might confilct. The trick is to add one central repo where all binaries gets uploaded. (I've an empty repo https://github.com/mathisloge/vcpkg-nuget where all my binaries are landing. But for this PR we need to add a repo to mapnik e.g. mapnik/vcpkg-cache. Additionally you will need to add a github personal access token with these scopes:
image

Afterwards you will need to add this newly created PAT to this repository so that the workflow can upload the binaries to mapnik/vcpkg-cache

Currently the tests can't be runned on windows since I need to rework the DLL copy mechanism for windows. It works well on a local machine, but not very reliable on a CI machine. 

The built mapnik binaries will be uploaded at the end of all runs as an artifact so it can be downloaded and used on a local machine. Those will be deleted after two days.

### DIFF
--- a/test/visual/run.cpp
+++ b/test/visual/run.cpp
@@ -44,10 +44,6 @@ log_levels_map log_levels{{"debug", mapnik::logger::severity_type::debug},
                           {"none", mapnik::logger::severity_type::none}};
 #endif
 
-#if defined(_WIN32)
-#include <windows.h>
-#endif
-
 using namespace visual_tests;
 namespace po = boost::program_options;
 
@@ -113,6 +109,7 @@ int main(int argc, char** argv)
     SetConsoleCP(CP_UTF8);
     SetConsoleOutputCP(CP_UTF8);
 #endif
+
     po::options_description desc("visual test runner");
     // clang-format off
     desc.add_options()


zy6p:
This action can support M1/arm macOS?

mathisloge:
There shouldn't be a problem. But since github actions don't have instances of arm m1 Macs, I can't test it.

mathisloge:
I'm having some strange visual test artifacts. 
e.g.:
image
image

or some font tests: 
image
base: 
image


@artemp Any  idea why this is happening? 

The test results (currently osx and ubuntu) https://github.com/mathisloge/mapnik/actions/runs/1579164405 could be downloaded here, at the bottom of the page.

artemp:
Text rendering (visual tests) is very sensitive to dependencies versions (`harfbuzz`, `freetype2`, `icu`).  The only way to have consistent output from visual tests is to use predefined set of dependencies (not system). Having said that, first two images ^^ look like something else .. hmm I'll need to take a look, thanks for highlighting!

mathisloge:
Ah okay. Since I'm using vcpkg for the dependency management, usually the most recent versions are used. 

The CI also uploads the build artifacts. So you could download `ubuntu-mapnik-artifact` and attach a debugger to the visual test. The CI builds in release mode but with debug information. The dependencies are linked static (on linux and macos).

artemp:
ok, the first image (and second, I'm guessing) is from `svg-renderer` (line-offset-900-250-1.0-svg.svg). Not a proper solution, but given `svg-renderer` haven't been used/tested vigorously, perhaps disabling `SVG_RENDERER=0` is an option? It looks like initialisation problem or 'close_path' not implemented correctly.. I'll see if there's a quick fix
@mathisloge

mathisloge:
@artemp I could leave it enabled since I've marked the visual test output as `continue-on-error` (so even if there are errors with the visual tests, it will mark the CI as OK but note that there were errors in the flow).

For the CI the general error handling is: 

* if one of the tests of catch2 is failing: Fail CI
* if the visual tests are failing: Continue with the CI flow but note that there were errors. 

Example were MacOS and Linux tests were successful but the visual tests weren't and windows tests are failed:
image

mathisloge:
All oses have now 100% test coverage🎉 Screenshot_20220126-012929

Remaining steps:

- [] tweak Windows coverage
- [] Split this PR into multiple smaller PRs for better Review

PR #4266, opened 2021-11-14
[CMake] improve build structure

mathisloge:
This PR improves the build output structure by outputting all executables and libraries into the same dir. Therefore no plugins, fonts and data needs to be copied across the build dirs. Furthermore this reduces long dll resolution times on windows since all needed deps are outputted into the same build dir (except for the plugin dlls, there we need to copy them into the main build dir).

- [x] Needs testing on linux
- [x] Needs install testing and pkg-conf testing


mathisloge:
@artemp ready to merge

PR #4153, opened 2020-06-14
Add Dockerfile and installation instructions

hiddewie:
Based on https://github.com/jawg/docker-mapnik3 and https://github.com/jawg/docker-mapnik3/pull/8

This should fix #3704.

Travis integration is still needed: https://docs.travis-ci.com/user/docker/#building-a-docker-image-from-a-dockerfile

The Python and Node bindings could have their own Docker image which also compile those bindings into the image.

Issue #3704, opened 2017-06-19
Official Docker images for Mapnik?


knutole:
Will there be offical Docker builds for Mapnik? There are [31 unofficial](https://github.com/search?utf8=%E2%9C%93&q=docker+mapnik&type=) ones, would be nice with a stable build of Mapnik 3.

artemp:
@knutole - 👍 
/cc @springmeyer

gisupc:
is docker image finished?

hiddewie:
Some good candidates (Apache License): https://github.com/jawg/docker-mapnik3.

Travis can also build the Dockerfile on each commit: https://docs.travis-ci.com/user/docker/#building-a-docker-image-from-a-dockerfile.

I can make a pull request which adds a Dockerfile which can be build automatically by Dockerhub for each release (and the latest commits of branches like master and `v3.0.x`). Would this be acceptable?

Furthermore, `python-mapnik` and `node-mapnik` can also use Docker images which would build their bindings to the Mapnik Docker image.

PR #4267, opened 2021-11-15
[help-needed] Docker

mathisloge:
related https://github.com/mapnik/mapnik/issues/3704 and https://github.com/mapnik/mapnik/pull/4153

What does this PR make?

* builds with cmake and leaves the dev packages in so that e.g. node-mapnik can directly build with the docker image. I have made an example here: https://github.com/mapnik/node-mapnik/pull/976
* github workflow: publishes the docker images into the github registry.  See: https://github.com/mathisloge/mapnik/pkgs/container/mapnik . So this branch is labeled `ghcr.io/mathisloge/mapnik:docker`

Pinging some people that were interested in the past: @knutole @hiddewie @jawg @joxit

maybe we can work together or leave some reviews. I'm not a mapnik docker user myself

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+build
+.vs
+.vscode
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,54 @@
+FROM ubuntu:21.04
+LABEL description="mapnik" 
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN mkdir /app
+COPY . /app
+WORKDIR /app
+
+RUN apt update
+RUN apt install -y gpg wget
+RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null 
+RUN echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ focal main' | tee /etc/apt/sources.list.d/kitware.list >/dev/null
+RUN apt update
+RUN apt install -y \
+    cmake \
+    ninja-build \
+    build-essential
+
+ENV BUILD_DEPENDENCIES="libicu-dev \
+    libfreetype6-dev \
+    libharfbuzz-dev \
+    libxml2-dev \
+    libjpeg-dev \
+    libtiff-dev \
+    libwebp-dev \
+    libcairo2-dev \
+    libproj-dev \
+    libgdal-dev \
+    libboost-filesystem-dev \
+    libboost-program-options-dev \
+    libboost-regex-dev \
+    "
+
+ENV RUNTIME_DEPENDENCIES="libfreetype6 \
+    libharfbuzz-bin \
+    libxml2 \
+    libjpeg8 \
+    libtiff5 \
+    libwebp6 \
+    libcairo2 \
+    libproj19 \
+    libgdal28 \
+    libboost-filesystem1.74-dev \
+    libboost-program-options1.74.0 \
+    libboost-regex1.74.0 \
+    "
+
+RUN apt install -y $BUILD_DEPENDENCIES $RUNTIME_DEPENDENCIES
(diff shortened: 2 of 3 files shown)


Joxit:
Hi there,
Are you sure this will work with GA ? Is there no time limit for a running workflow ? IIRC, building mapnik takes time :sweat_smile:

mathisloge:
hi @Joxit 
it has already runned :D https://github.com/mathisloge/mapnik/runs/4214019688?check_suite_focus=true 

But the limits for open source projects are actually quite high. each job can take up to 6 hours and a workflow up to 72 hours https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration 

But the workflow only takes half a hour. so it is quite ok

mathisloge:
@artemp this should be ready to merge. There shouldn't be any actions needed other than a merge. The workflow would use the github context to get the secrets / repo names. 

Each branch will have its own docker tag
Each tag labeled with `v*.*.*` will have its own docker tag. 
The latest semver will get the `latest` docker tag

You can see some example tags here: 
https://github.com/mathisloge/mapnik/pkgs/container/mapnik

Note, that this uses the cmake build. For node-mapnik or python-mapnik I will now find a way to use pkg-config instead of `mapnik-config`. 

Or is there a desire to use SCons to build mapnik on the Docker image?

.dockerignore
```diff
@@ -0,0 +1,3 @@
+build
+.vs
+.vscode
```
hiddewie, line 3: It would be good to add the `.git` directory here. It contains many files that do not influence the Docker build context or the build result. 

On my machine the size of the Docker build context went from 234MB to 48 MB with this change. 

Maybe more files/directories can be excluded. See `.gitignore`.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN apt-get update && apt-get install -y gpg wget \
+    && wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null \
+    && echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ focal main' | tee /etc/apt/sources.list.d/kitware.list >/dev/null \
+    && apt-get update \
+    && apt-get install -y \
+        cmake \
+        ninja-build \
+        build-essential \
+        $BUILD_DEPENDENCIES \
+    && cmake --preset linux-gcc-release -DBUILD_DEMO_VIEWER=OFF -DBUILD_TESTING=OFF -DBUILD_DEMO_CPP=OFF -DBUILD_BENCHMARK=OFF \
```
hiddewie, line 31: It might be a good idea to split this command into 2 parts: 
- installing the system dependencies
- building the project

This makes it possible to cache the Docker layer with the system dependencies. When the system dependencies do not change, the Docker build will start with the existing cached image.

The `COPY . .` should go after the system depedencies, because otherwise any change in any file in this project will invalidate the Docker layer cache and rebuild everything.

mathisloge:
@hiddewie thanks for your reviews. I've added more dirs to the ignore list. So it is reduced to 33MB. 
and splitted the commands as suggested.

zy6p:
Looking forward to see this PR merged. This will help with some custom plugins we created.

mathisloge:
@zy6p you could actually use the containers already. https://github.com/mathisloge/mapnik/pkgs/container/mapnik/10503672?tag=docker

zy6p:
> @zy6p you could actually use the containers already. [mathisloge/mapnik/pkgs/container/mapnik/10503672?tag=docker](https://github.com/mathisloge/mapnik/pkgs/container/mapnik/10503672?tag=docker)

Great! My laptop is M1 mac, and pre-build mapnik doesnot support arm. But these containers can work 😃

gdey:
Any movement on this?

artemp:
> Any movement on this?

I'm planning to take a look at this when I have time, I know it was a long time.

Meanwhile, is there any reasons you can't simply

```
npm install @mapnik/mapnik

```
?

PR #4270, opened 2021-11-19
fix proj version detection / harfbuzz pkg-config for cmake

mathisloge:
Some proj deployments can't access the version in the `find_package` call. 
With this change it will check the found version separately and errors out if the version is less then the `PROJ_MIN_VERSION` (currently 7.2.0). The version is however available after the package was found. 

And it drops the custom FindHarfbuzz and uses pkg-config as a backup, if the cmake version could not be found.
If users like in #4265 need the pkg-config behavior, they can pass `CMAKE_DISABLE_FIND_PACKAGE_harfbuzz=OFF` to the configure call to disable the `find_package`  call and force pkg-config

@artemp this is ready.


mathisloge:
@artemp can we merge this?

artemp:
:+1:

manthey:
I think you also need to remove the reference to `FindHarfBuzz.cmake` in `cmake/MapnikExport.cmake`.

PR #4282, opened 2022-01-26
Feature clang format

mathisloge:
Proposal for a clang-format file. 

I've tried to stay as close as possible to the current style (5615c2b4368da2fb0c00ae72c2a0f3231ab921a7). But sometimes the style has changed with the years passing by :D 

So I've formatted _some_ files to showcase the differences. 

If you have any other preferences, I'll try to add them to the clang format style file.

After you are happy with the new style, I would apply the style to all files in this PR as a single commit.


artemp:
@mathisloge I just had a brief look, one thing stands out 
 
`const&` -> `const &`

 I prefer original formatting style in this case ^^ which is how boost formats `c++`. I know this is a personal preference but the logic for me is: 

*  `&var` - address of `var`
*  `const& var` - const ref and `& var` non-const reference 

Thanks for starting looking into all this!

mathisloge:
@artemp thanks for reviewing! 

this makes sense. Have changed it so pointer and references 

pointer should be the same correct? 
so currently all pointer are formatted like this: 
```
Type* var; 
Type* var = &x;
Type& y = *var;

```

artemp:
Looks good to me👍

mathisloge:
for the boost::x3 rules I've turned off the formatting because the manual style can't be reflected with clang-format. 
via:
```c++
// clang-format off
... rules
// clang-format on
```
But I need to go through them all again, there might be some misses.

mathisloge:
I've formatted all files, looked through them and run the tests locally to test if there are some hidden issues. But it seems good to go.

mathisloge:
@artemp will resolve the merge conflicts 👍

mathisloge:
@artemp resolved

PR #4286, opened 2022-01-28
add github actions

mathisloge:
There are some needed steps. For the whole dependency management I'm using vcpkg. So most of the dependencies are very up to date. But this needs some setup steps. vcpkg is using binary caching via nuget to not build all dependencies each run. If you ever should use vcpkg in another repo, this might confilct. The trick is to add one central repo (e.g. named `vcpkg-cache`) where all binaries gets uploaded. (I've an empty repo https://github.com/mathisloge/vcpkg-nuget where all my binaries are landing. But for this PR we need to add a repo to mapnik e.g. mapnik/vcpkg-cache. Additionally you will need to add a github personal access token with these scopes to each repo using the binary cache
image

Afterwards you will need to add this newly created PAT to this repository so that the workflow can upload the binaries to mapnik/vcpkg-cache

Latest run from the #4262 branch (where the usernames are not changed) 
https://github.com/mathisloge/mapnik/actions/runs/1760952679

CodeCov:
https://app.codecov.io/gh/mathisloge/mapnik/branch/gh_ci/commits?page=1

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -1,0 +1,76 @@
+name: Build and Test
+
+on:
+  push:
+    branches-ignore:
+      - "no-ci-*"
+  pull_request:
+    branches:
+      - "**"
+
+jobs:
+  windows:
+    name: Windows memory mapped
+    uses: mapnik/mapnik/.github/workflows/windows.yml@gh_ci
+    with:
+      VCPKG_SHA: 2e34a79a814551424d9b1a981b3b110e0a8fbdfd
+      NUGET_REGISTRY: https://nuget.pkg.github.com/mapnik/index.json
+      NUGET_USERNAME: mapnik
+      USE_MEMORY_MAPPED_FILE: 'ON'
+    secrets:
+      NUGET_REGISTRY_PAT: ${{ secrets.VCPKG_NUGET_CI }}
+  
+  windows-mmf-off:
+    name: Windows file based
+    uses: mapnik/mapnik/.github/workflows/windows.yml@gh_ci
+    with:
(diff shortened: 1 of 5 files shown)


mathisloge:
This should work out of the box now, after the above steps have been taken

@artemp how do we proceed now?

mathisloge:
@artemp perfect! 

For the PAT: you would need to create one for the organization `mapnik` with the right shown in the picture above and add this PAT to the `mapnik/mapnik` repo.  You only need to tell me the name of it, so I can pass it into the actions. Name could be e.g.: `VCPKG_CACHE_PAT` . 

(https://docs.github.com/en/actions/security-guides/encrypted-secrets for usage of secrets in repositories and actions)

artemp:
> `VCPKG_CACHE_PAT`

:green_circle:

mathisloge:
Uoh, this was a bit early. :D I need to adjust the name of the token. Wasn't sure which name you would choose. 

I'll make a followup PR

artemp:
@mathisloge - sorry :)

mathisloge:
@artemp the first CI run might take some hours to finish, since all dependencies needs to be build.

The CI also might fail, since the runners only have 14gb of memory. If a lot of dependencies need to be built, the memory might run empty.
You can then just restart the run, after all runners have finished or failed.

artemp:
> @artemp the first CI run might take some hours to finish, since all dependencies needs to be build.
> 
> The CI also might fail, since the runners only have 14gb of memory. If a lot of dependencies need to be built, the memory might run empty. You can then just restart the run, after all runners have finished or failed.

Sure, no probs!

PR #4291, opened 2022-02-02
Feature/cmake static plugins

### LINKED ISSUE #4287: question about static plugins
I wanted to add the static plugin functionality to cmake, too. But https://github.com/mapnik/mapnik/blob/master/src/datasource_cache_static.cpp declares either plugins, which are not present anymore  or currently available plugins not. 

Missing plugin definitions: geobuf, pgraster, topojson

So my question is: Is this feature deprecated/not used by anyone or should it be maintained? (add missing definitions)

There are also double declared symbols when using ogr and gdal as a static plugin (`on_plugin_load`)

mathisloge:
fixes #4287 

@artemp I needed to refactor the plugin logic a bit. I've introduced a new class `datasource_plugin` which has a common interface. So multiple static plugins with the same `on_plugin_load` logic can exist. 

BUT as it now builds correctly and most tests are both in dynamic and static configuration good, the static configuration still has some flaws: the problem with singeltons then arises. e.g. the `image_reader` registration won't work correctly, since the compiler optimizes those calls away.


mathisloge:
So I've now added a `mapnik::setup()` function which should be called before any other method of mapnik. Currently this only registers the image_readers. 

This would mean a very big breaking change, since users have to add this function to their source code. IMHO the global initialisation with functions a little opaque, as it is not clear what happens when or why something might not happen. 

The plugins are working now in static and dynamic configurations. 

But I don't know how i can express the additional things in SCons.

mathisloge:
@artemp thoughts on this?

src/plugin.cpp
```diff
@@ -130,21 +118,21 @@ void* PluginInfo::get_symbol(std::string const& sym_name) const
...
+    if (module_ && module_->dl && !module_->name.empty())
         return true;
 #endif
     return false;
 }
 
 std::string PluginInfo::get_error() const
 {
-    return std::string("could not open: '") + name_ + "'";
+    return std::string("could not open: '") + module_->name + "'";
```
mathisloge, line 135: todo: possible nullptr

artemp:
@mathisloge I pushed `Scons` related changes into https://github.com/mapnik/mapnik/tree/mathisloge-feature/cmake-static 
Could you rebase your PR onto the branch ^^ given this is breaking change?

mathisloge:
@artemp done. 
But I would like to debug and test this with node mapnik etc. before merging (if you want to merge :D )

mathisloge:
I've added static CI builds. There are some issues on the different platforms: 

* Windows: somehow no tests are generated
* linux: link error with icu dat symbols.
* mac: double defined symbol `unesc_char` in  `csv_grammar_x3_def.hpp` and `expression_grammar_x3_def.hpp`

alsaleem00:
@artemp, is cmake/static ready to build? I am using it in a c++ app.
What is the timeline to merge cmake/static?

mathisloge:
@alsaleem00 hey sorry. Currently it is not ready. I still need to figure out how to fix some linking issues, when you build all dependencies static, too.

Currently I'm not that productive since I've begun a new Job. But it should get better in july/August...

If this time frame is too long for you, feel free to take on the problems.

mathisloge:
@artemp yey!
I've got the necessary bits finally together.
Could you review this PR?
From the functionality it is mergable.

mathisloge:
@artemp do you had any chance to review this? :)

artemp:
> @artemp do you had any chance to review this? :)
Happy New Year @mathisloge :) 

Q. How to ensure static build favours linking to static (*.a) dependecies?  
E.g `RUNTIME_LINK: Set preference for linking dependencies (shared|static)` ?

mathisloge:
Happy New Year @artemp 🥳 

> How to ensure static build favours linking to static (*.a) dependecies?

I don't know of any reliable way of doing it. I would leave that to the user and the user has to make sure to pass the correct config file locations or paths to cmake config / pkg-config files when configuring the project. E.g. you can set https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH.html#variable:CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH to disable the system path and make sure that only the passed paths are used.

artemp (dismissed):

mathisloge:
@artemp  needed to rebase. 
You can merge this now :)

artemp (approved):

PR #4295, opened 2022-02-08
[CI] disable vcpkg debug mode

mathisloge:
@artemp this just disables the vcpkg debug output and is ready to merge


mathisloge:
Saw a thing with the readonly token. Will fix

PR #4304, opened 2022-04-05
fix test for newer cairo versions

mathisloge:
Cairo has released a new version. Unfortunately they removed the version number in the svg tag. So the test will always fail with newer versions of cairo

see https://gitlab.freedesktop.org/cairo/cairo/-/issues/553


mathisloge:
@artemp yes

As soon as cairo upstream adds the version back, I'll extend the cairo version check with the max version

PR #4256, opened 2021-10-20
[CMake] add mapnik-config

mathisloge:
- [ ] Squash commits together

PR #4333, opened 2022-08-03
Fix font installation logic

artemp:
@mathisloge - font installation to match expected layout e.g fonts in
 `/mapnik-install-dir/lib/mapnik/fonts/`


springmeyer (commented):
To review questions:

- Can a whitespace linter be run over all the cmake files so that future changes like this don't have lots of whitespace changes? (all changes will have already been done)
- Can tests be run that confirm this change is correct/working?

mathisloge:
@springmeyer I'll look into this when I'm back from my vacation (in 2 weeks)

@artemp all other changes are looking good at first glance.

mathisloge:
@artemp could you check if the PAT for the CI is still valid? 
The authentication fails again when uploading the built packages.

artemp:
> @artemp could you check if the PAT for the CI is still valid? The authentication fails again when uploading the built packages.

@mathisloge I checked, PAT is valid e.g `Expires on Fri, Sep 30 2022` Hmm..can you share the message you're seeing?

artemp:
@mathisloge  - I'm trying to make `mapnik-render` use input plugins and font dirs relative to the executable. On Linux the following approach works: 

```diff
diff --git a/utils/mapnik-render/mapnik-render.cpp b/utils/mapnik-render/mapnik-render.cpp
index 608dea754..bf81b31d2 100644
--- a/utils/mapnik-render/mapnik-render.cpp
+++ b/utils/mapnik-render/mapnik-render.cpp
@@ -16,6 +16,7 @@ MAPNIK_DISABLE_WARNING_PUSH
 MAPNIK_DISABLE_WARNING_POP
 
 #include <string>
+#include <filesystem>
 
 int main(int argc, char** argv)
 {
@@ -32,7 +33,6 @@ int main(int argc, char** argv)
     logger.set_severity(mapnik::logger::error);
     int map_width = 600;
     int map_height = 400;
-
     try
     {
         po::options_description desc("mapnik-render utility");
@@ -118,9 +118,12 @@ int main(int argc, char** argv)
         {
             map_height = vm["map-height"].as<int>();
         }
-
-        mapnik::datasource_cache::instance().register_datasources("./plugins/input/");
-        mapnik::freetype_engine::register_fonts("./fonts", true);
+        std::string exe_path = "";
+#ifdef __linux__
+        exe_path = std::filesystem::canonical("/proc/self/exe").remove_filename().string();
+#endif
+        mapnik::datasource_cache::instance().register_datasources(exe_path + "../lib/mapnik/input/");
+        mapnik::freetype_engine::register_fonts(exe_path + "../lib/mapnik/fonts/", true);
         mapnik::Map map(map_width, map_height);
         mapnik::load_map(map, xml_file, true);
         map.zoom_all();
```

Do you know how to access exe path at runtime on Windows?

mathisloge:
> Do you know how to access exe path at runtime on Windows?

I think https://docs.microsoft.com/de-de/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamea would give you the path of the executable, if you pass `null`. 

However, don't you think it would be enough to use the current working directory and make it relative to that? 
At least in cmake I set the current working directory for the tests. And the renderer would normally run from its own directory too?

artemp:
> However, don't you think it would be enough to use the current working directory and make it relative to that? At least in cmake I set the current working directory for the tests. And the renderer would normally run from its own directory too?

Agree lets keep it simple! 

Remaining issue is that `mapnik-render` gets installed into `<install-dir>/bin` but registration paths are set for running from `<mapnik-src>/build/out`. I see two options:  Perhaps, we should add command options to set custom paths

mathisloge:
> Remaining issue is that mapnik-render gets installed into <install-dir>/bin but registration paths are set for running from <mapnik-src>/build/out

maybe we could add a program argument rootdir or something like that? And from that dir all relative paths are resolved? But in the end it would be nearly the same as using the current working dir. 

> but registration paths are set for running from <mapnik-src>/build/out

Where is this set? I can't find it in the source code. 

> Also, a big picture question: I think we should promote CMake as a default build system and deprecate scons in next Mapnik 4.x.x release. What do you think?

I think it would make it easier. Since maintaining two build systems is a good amount of work and I see a lot of build failures related to unresolved symbols with scons in the issues. But if we do this, I need to take up #4256 again. so that node-mapnik and python-mapnik can be build. But I don't know if it will be working. Need to check it again.

artemp:
> Where is this set? I can't find it in the source code.

Sorry, it wasn't clear:) 

Currently paths for plugins and fonts are set to
`./plugins/input`
`./fonts` in `mapnik-render.cpp`  but executable installed into `<mapnik-dir>/bin`. To make it work, dirs  
should be - `../lib/mapnik/input` and `../lib/mapnik/fonts` to match installation layout. 

> I think it would make it easier. Since maintaining two build systems is a good amount of work and I see a lot of build failures related to unresolved symbols with scons in the issues. But if we do this, I need to take up #4256 again. so that node-mapnik and python-mapnik can be build. But I don't know if it will be working. Need to check it again.

:+1:

mathisloge:
> Currently paths for plugins and fonts are set to `./plugins/input` `./fonts` in `mapnik-render.cpp` but executable installed into `<mapnik-dir>/bin`. To make it work, dirs should be - `../lib/mapnik/input` and `../lib/mapnik/fonts` to match installation layout.

Ah ok :) I think it would be good to provide two new program arguments. One to the fonts and one to the plugin dir. Then it would be independent from any installation layout and it would default to the standard installation layout.

mathisloge:
@springmeyer @artemp 

https://github.com/mapnik/mapnik/pull/4334 is adding the linting and a CI workflow to fix those whitespaces and more in each PR.

mathisloge:
@artemp maybe rebase on current master?

artemp:
> @artemp maybe rebase on current master?

@mathisloge Sure, shall I merge this PR into master 

> @artemp maybe rebase on current master?

:+1:  Let me know if all is good

artemp:
> > @artemp maybe rebase on current master?
> 
> @mathisloge Sure, shall I merge this PR into master
> 
> > @artemp maybe rebase on current master?
> 
> +1 Let me know if all is good

Yes

mathisloge (approved):

mathisloge:
> Sure, shall I merge this PR into master

yes

PR #4334, opened 2022-08-11
adding auto formatting

mathisloge:
Adds pre-commit config file, which strips whitespaces from cmake files and runs clang-format on every commit, if the developer runs `pre-commit install` locally in the repository.


artemp (approved):
Looking good. Personally, I'd prefer `clang-format` be a bit more forgiving regarding `namespace` and ` external "C"` formatting. And also forcing function arguments declarations on one line..  but we can amend `.clang-format` later :) Cheers!

artemp:
@mathisloge - do you know how to fix Windows builds? It complains about "Ninja" toolset ^

mathisloge:
@artemp 

> And also forcing function arguments declarations on one line.. but we can amend .clang-format later

we can fix clang in this PR to, as it touches so many files again :) 

>  do you know how to fix Windows builds? It complains about "Ninja" toolset

I need to check why the cache isn't reused. The ninja error is a bit misleading as all the builds consume the whole HDD memory of the CI runner.

mathisloge:
@artemp 
I've changed clang-format so that the extern C block is intended as follows: 
```
extern "C" {
... code ....
}
```

instead of 
```
extern "C" 
{
...code...
}
```

Could you clarify what you mean by `forgiving regarding namespace`? Unfortunatly I can only use one style for the whole project. But I could exclude agg source code.

artemp:
> Could you clarify what you mean by `forgiving regarding namespace`? Unfortunatly I can only use 
one style for the whole project.

I was hoping we can have flexible formatting for namespaces e.g both 

```
namespace A {
namespace B {
} //
} //
```
and

```
namespace A { namespace B {
}}
```

are accepted but I recalling clang-format is not that flexible. 

>  But I could exclude agg source code.
Sounds good, lets disable clang-format on `agg` and any third-party deps.

mathisloge:
> Sounds good, lets disable clang-format on `agg` and any third-party deps.

agg and deps are now excluded

> I was hoping we can have flexible formatting for namespaces e.g both

we could rewrite such namespaces as nested namespace as soon as we are switching to c++17 (as boost x3 is switching in 1.81). But then we can't support c++14 builds anymore

mathisloge:
If you are OK with this PR, you could merge it.

artemp (approved):

PR #4338, opened 2022-08-17
[CI] only run on master pushes and all PRs

mathisloge:
currently the CI is running on every push and each PR . 
As a result, for each commit in a mapnik/mapnik based branch, the CI is running twice. Now it is only running on master pushes and on PRs which aren't beginning with `no-ci-*`.


artemp (approved):
great call!

PR #4357, opened 2022-10-18
SVG: Fix reflection points for smooth curves after arcs

kkaefer:
I found that the patch in #4113, which attempts to fix the shorthand “smooth curve” commands, has a defect when interacting with “elliptical arc” commands. Internally, Agg/Mapnik converts arcs to curves. This means that the detection logic in the “smooth curve” command, which looks at the previous points for calculating the reflection point, mistakenly assumes that the previous commands were curve commands, even though they were in reality arc commands.

An example where this occurs is the following SVG path:

```
M 11.676 30.682 a 4.5 4.5 0 0 1 -6.364 -6.364 s 1.149 0.133 3.69 2.674 s 2.674 3.69 2.674 3.69 z
```

### DIFF
--- a/include/mapnik/svg/svg_path_adapter.hpp
+++ b/include/mapnik/svg/svg_path_adapter.hpp
@@ -382,10 +382,12 @@ void path_adapter<VC>::arc_to(double rx,
         {
             join_path(a);
         }
-        else
-        {
-            line_to(x, y);
-        }
+
+        // We are adding an explicit line_to, even if we've already add the
+        // bezier arc to the current path. This is to prevent subsequent smooth
+        // bezier curves from accidentally assuming that the previous command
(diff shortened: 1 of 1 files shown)


kkaefer:
This fix could probably be applied to `agg::path_storage` as well, which suffers from the same problem.

artemp:
@kkaefer - thanks, great fix!

kkaefer:
@artemp this doesn't actually fix the SVG rendering bug because the code is duplicated there. SVG rendering doesn’t actually use `agg::path_storage` AFAIK.

artemp:
@kkaefer - ah, ok :rofl:  I guess it doesn't harm to have this fix in case someone is using AGG directly. 

In any case, I'm going to run `svg2png` with a bunch of SVGs and post here if any artifacts ..

Issue #4363, opened 2022-11-16
SVG - `<g>` opacity is not handled correctly 


artemp:
The **opacity** attribute specifies the transparency of an object or of a group of objects, that is, the degree to which the background behind the element is overlaid.

 (https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/opacity)

`opacity` attribute requires special considerations when processing SVG document. 

### Example
```svg
<svg viewBox="0 0 320 320" version="1.1" xmlns="http://www.w3.org/2000/svg">
  <g opacity="0.5">
    <rect x="120" y="120" width="160" height="160" fill="blue"/>
    <circle cx="120" cy="120" r="80" fill="magenta" opacity="1.0" fill-opacity="0.666"/>
  </g>
</svg>
```
#### Current (incorrect) rendering 

![test](https://user-images.githubusercontent.com/59056/202208133-7af20856-8517-43bb-b277-aa08b00d63f3.png)

#### Correct rendering - applying `<g>` `opacity` when blending group elements into parent canvas.
 
![test](https://user-images.githubusercontent.com/59056/202207758-756cb468-0f34-4c4f-b01b-dba4d98c1083.png)

artemp:
https://github.com/mapnik/mapnik/tree/svg-group-render - initial group/element `opacity` implementation

TODO:
* Fix unit + visual tests 
* Better names

artemp:
```svg
<svg xmlns="http://www.w3.org/2000/svg" width="300" height="300">
  <g stroke-width="20" fill="#ffff00">
    <circle cx="150" cy="125" r="100" fill="#0000ff" stroke="#ff00ff"/>
    <rect x="25" y="100" width="250" height="50" rx="10" ry="10" stroke="#008000"/>
  </g>
</svg>
```

![image](https://user-images.githubusercontent.com/59056/202741567-8ef41d42-5937-4be4-90b7-11893025127b.png)


```svg
<svg xmlns="http://www.w3.org/2000/svg" width="300" height="300">
  <g opacity="0.5" stroke-width="20" fill="#ffff00">
    <circle cx="150" cy="125" r="100" fill="#0000ff" stroke="#ff00ff"/>
    <rect x="25" y="100" width="250" height="50" rx="10" ry="10" stroke="#008000"/>
  </g>
</svg>
```
![image](https://user-images.githubusercontent.com/59056/202740662-38994c63-64ba-418f-b142-72e1e00255c3.png)

```svg
[...]

PR #4364, opened 2022-11-21
Svg group renderer [WIP]

artemp:
See https://github.com/mapnik/mapnik/issues/4363 for details


include/mapnik/svg/svg_group.hpp
```diff
@@ -0,0 +1,48 @@
+/*****************************************************************************
```
mathisloge, line 1: Maybe start using SPDX License headers? 
Instead the whole text using: 
`// SPDX-License-Identifier: LGPL-2.1-only`

See https://spdx.dev/ids/#how and https://spdx.org/licenses/LGPL-2.1-only.html

artemp: Good call @mathisloge! I'll take a look, thanks.

mathisloge:
@artemp I think you will need to rebase from master so that the windows-ci will work again

PR #4365, opened 2022-11-21
[CI] update vcpkg

### LINKED ISSUE #4362: update vcpkg in ci

mathisloge:
closes #4362

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -23,7 +23,7 @@ jobs:
     name: Windows memory mapped
     uses: ./.github/workflows/windows.yml
     with:
-      VCPKG_SHA: f6af75acc923c833a5620943e3fc7d5e4930f0df
+      VCPKG_SHA: 34d2cf7e62d781f4bcb9c7f44f4d2389f568e92b
       NUGET_REGISTRY: https://nuget.pkg.github.com/mapnik/index.json
       NUGET_USERNAME: mapnik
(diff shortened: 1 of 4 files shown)


artemp:
> @artemp this would be ready to merge. But I've updated the llvm toolchain to 15. This results in some other formatting. See the diff in the action: https://github.com/mapnik/mapnik/actions/runs/3518341002/jobs/5897145563

👍 thanks, I'll take a look tomorrow.

artemp:
@mathisloge couple ideas - 

* should we disable formatting on third-party code e.g [catch.hpp](https://github.com/mapnik/mapnik/blob/master/test/catch.hpp) ? 

* would it be possible to update `.clang-format` to preserve original behaviour? New formatting is looking fine to me but I'd like it to be less strict.

mathisloge:
> should we disable formatting on third-party code e.g [catch.hpp](https://github.com/mapnik/mapnik/blob/master/test/catch.hpp) ? 

Typically yes. but as it is in the source dir, I need to look into it how to exclude it. Generally I want to remove it and aquire it via cmake fetch content directly from github.  But I can only do that, as soon as the scons build is removed. 

> would it be possible to update .clang-format to preserve original behaviour? New formatting is looking fine to me but I'd like it to be less strict.

I need to look into it. Have reverted the change for now. So that the vcpkg pipeline can be run again

mathisloge:
@artemp let us merge this first and I will then look into updating clang-format

artemp (approved):

PR #4366, opened 2022-11-21
[CI] refactor github actions

mathisloge:
I've refactored the existing github ci actions. 
Furthermore I've added a new pipeline `linux_release` which runs on ubuntu 22.04 which builds against the packages of the apt package manager. At the end it will upload a debian package which can be used to install mapnik on the local system.


mathisloge:
@artemp this would be ready to merge

PR #4368, opened 2022-11-24
Fix windows build error

mathisloge:
Latest msvc compiler can't correctly use find the correct overloaded function: 

```
D:\a\mapnik\mapnik\test\unit\vertex_adapter\simplify_converters_test.cpp(40): error C2668: 'mapnik::util::to_wkt': ambiguous call to overloaded function
D:\a\mapnik\mapnik\include\mapnik/util/geometry_to_wkt.hpp(36): note: could be 'bool mapnik::util::to_wkt(std::string &,const mapnik::geometry::geometry<int[64](https://github.com/mapnik/mapnik/actions/runs/3526352727/jobs/5914192007#step:17:65)_t> &)'
D:\a\mapnik\mapnik\include\mapnik/util/geometry_to_wkt.hpp(34): note: or       'bool mapnik::util::to_wkt(std::string &,const mapnik::geometry::geometry<double> &)'
D:\a\mapnik\mapnik\test\unit\vertex_adapter\simplify_converters_test.cpp(40): note: while trying to match the argument list '(std::string, mapbox::geometry::line_string<T,std::vector>)'
        with
        [
            T=double
        ]
```

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -101,7 +101,7 @@ jobs:
         continue-on-error: true
         working-directory: build/${{ env.preset }}/out
         env:
-          PROJ_LIB: ${{ github.workspace }}/build/${{ env.preset }}/vcpkg_installed/x64-osx/share/proj/data
+          PROJ_LIB: ${{ github.workspace }}/build/${{ env.preset }}/vcpkg_installed/x64-osx/share/proj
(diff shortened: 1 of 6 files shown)


mathisloge:
@artemp this is ready to merge. I had to fix one thing in the tests. Now all CIs are running again

Issue #4359, opened 2022-10-21
fatal error: mapbox/variant.hpp: No such file or directory 


the-dias:
Got an error after doing all the steps from the installation [guide](https://github.com/mapnik/mapnik/blob/master/INSTALL.md)  when run demo/c++ program. 

I found a similar problem, did what was suggested there, but it did not help #3246

Issue #4243, opened 2021-09-01
Fresh install:  undefined reference to `mapnik::feature_style_processor<mapnik::cairo_renderer<std::shared_ptr<_cairo> > >::apply(double)


yohanboniface:
When I try to build Mapnik, after a fresh clone, it fails with this message:

```
/usr/bin/ld: demo/c++/rundemo.o: in function `main':
rundemo.cpp:(.text.startup+0x2699): undefined reference to `mapnik::feature_style_processor<mapnik::agg_renderer<mapnik::image<mapnik::rgba8_t>, mapnik::label_collision_detector4> >::apply(double)'
/usr/bin/ld: rundemo.cpp:(.text.startup+0x2aba): undefined reference to `mapnik::feature_style_processor<mapnik::cairo_renderer<std::shared_ptr<_cairo> > >::apply(double)'
collect2: error: ld returned 1 exit status
scons: *** [demo/c++/rundemo] Error 1
scons: building terminated because of errors.
make: *** [Makefile:48: mapnik] Error 2
```

Here is `./configure` output:

```
leonardo:~/C/c/mapnik.new (master=) ./configure                                                                                                                                  48m 12s 541ms
scons: Reading SConscript files ...

Welcome to Mapnik...

Configuring build environment...
Configuring on Linux in *release mode*...
C++ compiler: c++ (GCC) 11.1.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Checking for pkg-config... yes
Checking for freetype2... yes
Checking for dlfcn.h support ... yes
Checking if compiler (c++) supports -std=c++14 flag... yes
Checking for C library z... yes
Checking for C++ library icuuc... yes
Checking for ICU version >= 4.0... yes (found ICU 69.1)
Checking for C++ library harfbuzz... yes
Checking for HarfBuzz version >= 0.9.34... yes (found HarfBuzz 2.9.0)
Checking for HarfBuzz with freetype support... yes
Searching for boost libs and headers... (cached) 
Found boost libs: /usr/lib
Found boost headers: /usr/include
Checking for C++ header file boost/version.hpp... yes
Checking for Boost version >= 1.61... yes
Found boost lib version... 1_76
Checking for C++ library boost_system... yes
Checking for C++ library boost_filesystem... yes
Checking for C++ library boost_regex... yes
[...]

PR #4370, opened 2022-12-09
Fix compiler errors

robert-werner:
Several issues like this (https://github.com/mapnik/mapnik/issues/4359#issue-1418892371) were opened, and fixes were proposed, but no PRs were created to manage these errors without patches. So I decided to create this PR.


luk-brue (commented):
Are the changes in `SConstruct` really necessary? I read in #4243  that this was the first workaround, but the substantial changes are those in the other three files. At least this is my interpretation. I am currently running a build only with changes in the other three files, let's see if that works as well.

robert-werner:
> Are the changes in `SConstruct` really necessary?

Yes, they _are_ necessary. Otherwise it would fail to build, rendering mapnik un-usable for any usage.
I've built mapnik for usage in OSM server, tiles are rendering fine without errors.

luk-brue:
I was able to compile completely without the removal of `-fvisibility=hidden` flags as default compiler option. For sure this does not mean that it will run well, I did not test any further. Did you try this?  

My scepticism is  due to the thought that the flag `-fvisibility` type flags might have been added with a reason in mind that we don't know. Note how the [official documentation for visibility support](https://gcc.gnu.org/wiki/Visibility) mentions performance reasons. If I am mistaken, please explain to me why it would be beneficial to remove the flag from global compiler defaults.

if we just remove it, it might cause other problems that go unnoticed? The flags purpose is to set the visibility to hidden as a default value unless declared otherwise. ([Explanation on SO](https://stackoverflow.com/questions/3570355/c-fvisibility-hidden-fvisibility-inlines-hidden)). So usually, there is a macro to declare objects as visible, in this case it is `MAPNIK_DECL`. 
@albertas-akistechnologies [fix](https://github.com/mapnik/mapnik/issues/4243#issuecomment-1118604455) mentions that their fix is declaring a template as visible. I don't fully understand this because I am not a programmer but even without that knowledge is think removing the default=hidden visibility feature globally is a more drastic thing to do than to manually set the desired objects as visible. 

If this is not right, please help me understand!

But if my reasoning is not wrong, my suggestion would be that you or someone else should confirm whether mapnik is building & functioning properly without the changes in `SConstruct`, and then we could merge this.

luk-brue:
Because this has gone stale unfortunately, could you maybe review these changes @mathisloge ? It would fix some issues, and make building easier again if you are a newbie like me and don't know the workarounds. Or do you have other plans coming in for fixing the compilation errors?

PR #4382, opened 2023-03-01
Add `fonts_dir` & `plugins_dir` variables to `libmapnik.pc`

hummeltech:
In order to provide some of the functionality previously provided by `mapnik-config`

* `mapnik-config --fonts` => `pkg-config libmapnik --variable=fonts_dir`
* `mapnik-config --input-plugins` => `pkg-config libmapnik --variable=plugins_dir`


artemp:
@hummeltech - looks good to me
@mathisloge ?

mathisloge:
Should be good. 
There could be one special case when setting `FONTS_INSTALL_DIR` to a different location then the install dir. But I think we don't tinker on this now and see if anyone runs into problems with it.

mathisloge (approved):
Tested it locally

PR #4383, opened 2023-03-09
Use `std::filesystem` when CXX_STD >= 17 and not `USE_BOOST_FILESYSTEM`

artemp:


src/fs.cpp
```diff
@@ -25,11 +25,18 @@
...
+#include <filesystem>
+namespace fs = std::filesystem;
+#else
 MAPNIK_DISABLE_WARNING_PUSH
 #include <mapnik/warning_ignore.hpp>
 #include <boost/filesystem/operations.hpp> // for absolute, exists, etc
 #include <boost/filesystem/path.hpp>       // for path, operator/
 MAPNIK_DISABLE_WARNING_POP
+namespace fs = boost::filesystem;
+#endif
```
mathisloge, line 39: maybe we should add a file in `include/mapnik/fs.hpp` with 
```cpp
#if defined(__cpp_lib_filesystem) && !defined(USE_BOOST_FILESYSTEM)
#include <filesystem>
#else
#include <boost/filesystem/operations.hpp> // for absolute, exists, etc
#include <boost/filesystem/path.hpp>       // for path, operator/
#endif

namespace mapnik {
#if defined(__cpp_lib_filesystem) && !defined(USE_BOOST_FILESYSTEM)
namespace fs = std::filesystem;
#else
namespace fs = boost::filesystem;
#endif
}
```
and then just include it everywhere where we need the filesystem?

artemp: https://github.com/mapnik/mapnik/pull/4383/commits/a244effa91c04f9e46e506e72213fbd1080516af

CMakeLists.txt
```diff
@@ -154,7 +155,14 @@ mapnik_find_package(PkgConfig REQUIRED)
 mapnik_find_threads()
 mapnik_find_package(ICU REQUIRED COMPONENTS uc i18n data)
 
-mapnik_find_package(Boost ${BOOST_MIN_VERSION} REQUIRED COMPONENTS filesystem system regex)
+mapnik_find_package(Boost ${BOOST_MIN_VERSION} REQUIRED COMPONENTS regex)
+if(CMAKE_CXX_STANDARD LESS 17)
+   set(USE_BOOST_FILESYSTEM ON)
+endif()
+if(CMAKE_CXX_STANDARD LESS 17 OR USE_BOOST_FILESYSTEM)
```
mathisloge, line 162:
```suggestion
if(USE_BOOST_FILESYSTEM)
```

test/visual/CMakeLists.txt
```diff
@@ -7,11 +7,17 @@ add_executable(mapnik-test-visual
...
     mapnik::agg
     ICU::data ICU::i18n ICU::uc # needed for the static build (TODO: why isn't this correctly propagated from mapnik::mapnik?)
 )
+
+if(CMAKE_CXX_STANDARD LESS 17 OR USE_BOOST_FILESYSTEM)
+target_link_libraries(mapnik-test-visual PRIVATE
+    Boost::filesystem
+)
+endif()
+
```
mathisloge, line 20:
```suggestion
```
These lines shouldn't be needed, since `mapnik::mapnik` either depends on filesystem or not.

mathisloge (approved):

PR #4384, opened 2023-03-17
Fixed check for `__cpp_lib_filesystem` in `filesystem.hpp`

hummeltech:
`__cpp_lib_filesystem` is only defined within `filesystem` & `version`, therefore `#if defined(__cpp_lib_filesystem)` is always `false` which causes all subsequent `std::filesystem` function calls to fail.

I.E.:
```
mapnik/src/fs.cpp:40:16: error: ‘exists’ is not a member of ‘mapnik::fs’
   40 |     return fs::exists(filepath);
      |                ^~~~~~
```
```shell
$ find /usr/include/c++/12.2.1 | xargs grep -sn '__cpp_lib_filesystem'
/usr/include/c++/12.2.1/filesystem:49:#define __cpp_lib_filesystem 201703L
/usr/include/c++/12.2.1/version:145:#define __cpp_lib_filesystem 201703L
```

### DIFF
--- a/include/mapnik/filesystem.hpp
+++ b/include/mapnik/filesystem.hpp
@@ -23,6 +23,8 @@
 #ifndef MAPNIK_FILESYSTEM_HPP
 #define MAPNIK_FILESYSTEM_HPP
 
+#include <version>
+
 #if defined(__cpp_lib_filesystem) && !defined(USE_BOOST_FILESYSTEM)
 #include <filesystem>
 #else
--- a/include/mapnik/filesystem.hpp
+++ b/include/mapnik/filesystem.hpp
@@ -23,9 +23,7 @@
 #ifndef MAPNIK_FILESYSTEM_HPP
 #define MAPNIK_FILESYSTEM_HPP
 
-#include <version>
-
-#if defined(__cpp_lib_filesystem) && !defined(USE_BOOST_FILESYSTEM)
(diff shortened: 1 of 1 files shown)


mathisloge (approved):
oh yeah, sorry, that was my fault!

include/mapnik/filesystem.hpp
```diff
@@ -23,7 +23,7 @@
 #ifndef MAPNIK_FILESYSTEM_HPP
 #define MAPNIK_FILESYSTEM_HPP
 
-#if defined(__cpp_lib_filesystem) && !defined(USE_BOOST_FILESYSTEM)
+#if (__cplusplus >= 201703L) && !defined(USE_BOOST_FILESYSTEM)
```
hummeltech, line 26: This should allow compiling with MSVC without specifying the `/Zc:__cplusplus` option:
```suggestion
#if ((__cplusplus >= 201703L) || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)) && !defined(USE_BOOST_FILESYSTEM)
```

mathisloge: Let me check it tomorrow. Normally this shouldnt be needed.

hummeltech: No problem, just trying to be thorough here, I mentioned it in case using `/Zc:__cplusplus` with `MSVC` was not something you wanted to do.

PR #4385, opened 2023-03-17
Add build & test for `CXX_STANDARD=17` to GitHub Actions

hummeltech:
This had been failing when using `-DCMAKE_CXX_STANDARD=17` on Linux, but was fixed with #4384


mathisloge (dismissed):

hummeltech:
I rebased this after the merging of https://github.com/mapnik/mapnik/pull/4384 and it is no longer failing to build `fs.cpp.o`:
Before:
https://github.com/mapnik/mapnik/actions/runs/4451141582/jobs/7817610102#step:9:58
After:
https://github.com/mapnik/mapnik/actions/runs/4455985544/jobs/7826141574#step:9:57

hummeltech:
Unfortunately Windows does not seem to have `std::filesystem`:
https://github.com/mapnik/mapnik/actions/runs/4455985544/jobs/7826141628#step:9:76
It appears to be experimental (`std::experimental::filesystem::v1`):
https://learn.microsoft.com/en-us/cpp/standard-library/filesystem?view=msvc-170

mathisloge:
Windows does have a Filesystem. Let me look into it tomorrow.

hummeltech:
Yeah, right after I wrote the last message I read a little bit more and saw that I was mistaken.
:facepalm:
I still don't know what's going on, but I'm not that familiar with Windows development, thanks for the help!

hummeltech:
Alright, it looks like checking `__cplusplus` might not be the best option either as it requires an additional option:
https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170

I believe I may have fixed the issue [here](https://github.com/mapnik/mapnik/pull/4385/commits/f6cd056552f1ab3b90609bf95a3f5b12a76ee8b6) but it might be a good idea to revisit the underlying issue.

mathisloge:
In fact, I don't think we should check the cpp mode at this point. 
In CMake we check for c++17 and if this is not the case, boost filesystem is enforced. What is defined in CMake should be the only single point of truth (the definition will be exported). 
Now when an application consumes mapnik either BOOST_FILESYSTEM=ON so boost filesystem must be used everywhere in the API interfaces. If =OFF, then std::filesystem.

So we should just look at `!defined(USE_BOOST_FILESYSTEM) #include<filesystem> #else #include<boost/filesystem>`

hummeltech:
Yeah, I like that, it sounds good to me! Let CMake handle it (as is already supported,) then it just leaves it up to manual specification with SCons (which is also already supported).

:+1:

mathisloge:
SCons will be deprecated in the future (CMake will be the main build system for 4.0)

hummeltech:
Aww, that didn't work either. I think I figured out partly why it wasn't working though. Since `mapnik_option` is a macro for [`option`](https://cmake.org/cmake/help/latest/command/option.html), it causes a `BOOLEAN CACHE` variable to be created. This cannot be overwritten with a simple [`Set Normal Variable`](https://cmake.org/cmake/help/latest/command/set.html#set-normal-variable), therefore a `FORCE`d [`Set Cache Entry`](https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry) must be used. This was causing `USE_BOOST_FILESYSTEM` to always be set to `OFF`.

Before bb280a2:
```shell
$ rm -rf build; cmake -B build -S . -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON -DCMAKE_CXX_STANDARD=14 -LA | grep USE_BOOST_FILESYSTEM:
USE_BOOST_FILESYSTEM:BOOL=OFF

$ rm -rf build; cmake -B build -S . -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON -DCMAKE_CXX_STANDARD=17 -LA | grep USE_BOOST_FILESYSTEM:
USE_BOOST_FILESYSTEM:BOOL=OFF
```

After bb280a2:
```shell
$ rm -rf build; cmake -B build -S . -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON -DCMAKE_CXX_STANDARD=14 -LA | grep USE_BOOST_FILESYSTEM:
USE_BOOST_FILESYSTEM:BOOL=ON

$ rm -rf build; cmake -B build -S . -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON -DCMAKE_CXX_STANDARD=17 -LA | grep USE_BOOST_FILESYSTEM:
USE_BOOST_FILESYSTEM:BOOL=OFF
```

hummeltech:
Alright, it was just a typo `USE_BOOST_FILESYSYTEM` that was causing all of these issues. I think this should be ready to go now!

artemp (approved):

Issue #4403, opened 2023-09-07
Drop support for GDAL < 2


Nakaner:
I am currently implementing support for the so-called "open options" of OGR. By working on the code, I found that Mapnik still support GDAL versions prior to 2.0.0. This leads to duplicated code with preprocessor macros:

```cpp
#if GDAL_VERSION_MAJOR >= 2
        unsigned int nOpenFlags = GDAL_OF_READONLY | GDAL_OF_VECTOR;
        const char* papszAllowedDrivers[] = {driver.c_str(), nullptr};
        dataset_ = reinterpret_cast<gdal_dataset_type>(
          GDALOpenEx(dataset_name_.c_str(), nOpenFlags, papszAllowedDrivers, nullptr, nullptr));
#else
        OGRSFDriver* ogr_driver = OGRSFDriverRegistrar::GetRegistrar()->GetDriverByName(driver.c_str());
        if (ogr_driver && ogr_driver != nullptr)
        {
            dataset_ = ogr_driver->Open((dataset_name_).c_str(), false);
        }
#endif
```

In order to avoid duplicated implementation, I suggest to drop support for GDAL prior to 2.0.0. GDAL 2.0.0 was released in March 2018 (more than five years ago). Almost all distributions (apart from few long-term support versions) should come with at least GDAL >= 2.0.

PR #4420, opened 2023-11-16
Add support for open_options in OGR input plugin

Nakaner:
Various OGR drivers require so-called *open options*. In my use case, I want to access vector tiles (MVT format) stored in a MBTiles file. OGR [requires](https://gdal.org/drivers/raster/mbtiles.html#opening-options) the open option `ZOOM_LEVEL=<NUMBER>` to indicate which zoom level should be used.

At Geofabrik, we currently use plain files as tile storage (z/x/y.pbf on disk) and the OGR input plugin for our vector to raster tool chain. By replacing it with a MBTiles file all the file system overhead can be saved and tile rendering is as fast as with files on the disk.

This pull request does implement the new option for the OGR input plugin and OGR/GDAL version 2 and newer. I did not implement it for old versions of GDAL. See #4403 for the reasons. I suggest to merge #4419 first. I will rebase afterwards.

### DIFF
--- a/plugins/input/ogr/build.py
+++ b/plugins/input/ogr/build.py
@@ -31,6 +31,7 @@ plugin_sources = Split(
   """
   %(PLUGIN_NAME)s_converter.cpp
   %(PLUGIN_NAME)s_datasource.cpp
+  %(PLUGIN_NAME)s_utils.cpp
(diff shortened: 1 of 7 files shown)


artemp:
@Nakaner - #4419 is merged

PR #4421, opened 2023-11-20
Fix CMake build issues after OGR open_options support addition

hummeltech:
`mapnik` currently has issues building with `CMake` after merging #4420, this resolves those issues.

- The newly added file `plugins/input/ogr/ogr_utils.cpp` was missing from two `CMakeLists.txt` files

Before: https://github.com/mapnik/mapnik/actions/runs/6941668588/job/18883003006#step:9:2862
After: https://github.com/mapnik/mapnik/actions/runs/6934483283/job/18862689256#step:9:2864


mathisloge (approved):
Should be good. Currently out for a longer vacation.

PR #4425, opened 2023-11-24
Fix linking error

Nakaner:
This pull request uses fixes described in https://github.com/mapnik/mapnik/issues/4243#issuecomment-1118604455

It is an alternative to #4370 but with fewer changes (and addresses the [comment by @luk-brue](https://github.com/mapnik/mapnik/pull/4370#pullrequestreview-1629876944) on that pull request.

I can confirm the original bug report on both Arch Linux (same as the reporter) and Ubuntu Jammy.


Nakaner:
Building of the current master branch and 4.0.0-rc1 works without this patch now (due to minimum required C++ version being raised to 17 and . Therefore this pull request can be closed.

PR #4445, opened 2024-04-19
replace majority of boost->std optional and boost->std string_view 

mathisloge:


include/mapnik/evaluate_global_attributes.hpp
```diff
@@ -265,7 +265,7 @@ std::tuple<T, bool> pre_evaluate_expression(expression_ptr const& expr)
 {
     try
     {
-        return std::make_tuple(util::apply_visitor(mapnik::evaluate_expression<T, boost::none_t>(boost::none), *expr),
+        return std::make_tuple(util::apply_visitor(mapnik::evaluate_expression<T, std::nullopt_t>(std::nullopt), *expr),
```
mathisloge, line 268: check if correct

src/mapped_memory_cache.cpp
```diff
@@ -63,19 +63,17 @@ bool mapped_memory_cache::remove(std::string const& key)
...
 
-boost::optional<mapped_region_ptr> mapped_memory_cache::find(std::string const& uri, bool update_cache)
+std::optional<mapped_region_ptr> mapped_memory_cache::find(std::string const& uri, bool update_cache)
 {
 #ifdef MAPNIK_THREADSAFE
     std::lock_guard<std::mutex> lock(mutex_);
 #endif
 
     using iterator_type = std::unordered_map<std::string, mapped_region_ptr>::const_iterator;
-    boost::optional<mapped_region_ptr> result;
```
mathisloge, line 73: revert to previous logic to not get any logic changes

test/unit/datasource/geojson.cpp
```diff
@@ -152,14 +151,15 @@ TEST_CASE("geojson")
...
             }
 
-            auto invalid_empty_geometries = {"{ \"type\": \"Point\", \"coordinates\": [] }",
-                                             "{ \"type\": \"LineString\", \"coordinates\": [[]] }"
-                                             "{ \"type\": \"Polygon\", \"coordinates\": [[[]]] }",
-                                             "{ \"type\": \"MultiPoint\", \"coordinates\": [[]] }",
-                                             "{ \"type\": \"MultiLineString\", \"coordinates\": [[[]]] }",
-                                             "{ \"type\": \"MultiPolygon\", \"coordinates\": [[[[]]]] }"};
+            constexpr auto invalid_empty_geometries =
+              std::array{"{ \"type\": \"Point\", \"coordinates\": [] }",
```
mathisloge, line 155: this fails for some compiler. add explicit decuction

include/mapnik/evaluate_global_attributes.hpp
mathisloge: should be good by looking at the code and the test results

mathisloge:
@artemp There are some missing bits, where the optional value type has a reference type. I would do these in a seperate PR, as depending on the approch taken, would be a bit more code change then just renaming.

mathisloge:
@artemp could you review this?

artemp:
@mathisloge looks good, going to merge!

artemp:
> @artemp There are some missing bits, where the optional value type has a reference type. I would do these in a seperate PR, as depending on the approch taken, would be a bit more code change then just renaming.

`std::optional< std::reference_wrapper<T>>`

mathisloge:
Yeah, that would be one way, with the least intrusive change.

PR #4452, opened 2024-05-16
don't pollute include dir with installed submodules

Vedingrot:
When install mapnik with bundled submodules it creates `mapbox` and
`protozero` directories which could conflict with previously installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -455,16 +455,20 @@ file(GLOB TTF_FONT_FILES "fonts/*/*/*.ttf")
 install(FILES ${TTF_FONT_FILES} DESTINATION "${FONTS_INSTALL_DIR}")
 
 if(NOT USE_EXTERNAL_MAPBOX_GEOMETRY)
-    install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/deps/mapbox/geometry/include/" DESTINATION "${MAPNIK_INCLUDE_DIR}")
+    install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/deps/mapbox/geometry/include/"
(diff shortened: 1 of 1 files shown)


hummeltech:
I see what you mean, but if one already has those dependencies installed, wouldn't they just build `Mapnik` with the corresponding `CMake` options?
I.E.:
* `USE_EXTERNAL_MAPBOX_GEOMETRY`
* `USE_EXTERNAL_MAPBOX_POLYLABEL`
* `USE_EXTERNAL_MAPBOX_PROTOZERO`
* `USE_EXTERNAL_MAPBOX_VARIANT`

Vedingrot:
For example, include/mapnik may contains unsupported version of
dependencies that requires by other programs. Using submodules solves
this problem but they should be installed in a directory that "belongs"
only for mapnik because they will be used only by it.

mathisloge:
I don't think that this is really needed, since you should install mapnik in a extra dir, if you have conflicting versions of those, since you should never link those bits together. 

I do have concernce regarding the integration, since the headers now need to be included as <mapnik/protozero/...> etc.

Vedingrot:
There is compiler flag `-I"${includedir}/mapnik` that has been added
in https://github.com/mapnik/mapnik/pull/4441/files so there's no
need to change headers.

landryb:
i have to agree with the PR here, one might have protozero installed systemwide, which isnt the same version as the one forked by mapbox.. osrm-backend also installs some mapbox headers under `include/mapbox` so they would conflict.

my solution is to pass `-DCMAKE_INSTALL_INCLUDEDIR=/usr/local/include/mapnik` so that everything mapnik-related is in its own dir..

PR #4453, opened 2024-05-16
Set proper soversion

Vedingrot:
Soversion should only reflects the major version, since the soname
link cannot be created due to equal name as the library. Library
libmapnik.so.4.0.0 prevents soname libmapnik.so.4.0.0 from being
created.


mathisloge:
As per https://github.com/mapnik/mapnik/pull/4248#issuecomment-925596509 we explicitly set the whole number, as (currently) there isn't a strong focus on ABI stability.

PR #4464, opened 2024-06-19
Use CMAKE_INSTALL_LIBDIR for libdir in pkg-config files

tomhughes:
This ensure that we use `lib64` instead of `lib` when appropriate.

### DIFF
--- a/cmake/MapnikExportPkgConfig.cmake
+++ b/cmake/MapnikExportPkgConfig.cmake
@@ -3,7 +3,7 @@ function(create_pkg_config_file _target _lib_name _description)
 prefix=@CMAKE_INSTALL_PREFIX@
 exec_prefix=${prefix}
 includedir=${prefix}/include
-libdir=${exec_prefix}/lib
+libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
 
 Name: @_lib_name@
 Description: @_description@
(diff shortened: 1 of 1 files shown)


hummeltech:
There is a definition of `MAPNIK_LIB_DIR` [here](https://github.com/mapnik/mapnik/blob/e7d0b85dbc1683efcd21fb117f6ddb5c1776491d/CMakeLists.txt#L102) which may be prefererable.

artemp:
Looks good to me. @mathisloge ?

cmake/MapnikExportPkgConfig.cmake
```diff
@@ -3,7 +3,7 @@ function(create_pkg_config_file _target _lib_name _description)
 prefix=@CMAKE_INSTALL_PREFIX@
 exec_prefix=${prefix}
 includedir=${prefix}/include
-libdir=${exec_prefix}/lib
+libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
```
mathisloge, line 6: as @hummeltech  has written, I would prefer `MAPNIK_LIB_DIR` to be used

artemp: 👍

artemp:
thanks! @tomhughes @hummeltech @mathisloge

PR #4466, opened 2024-06-25
Speed up the GitHub Actions CI jobs

hummeltech:
It is currently taking almost 2 hours for the `Build and Test` workflow to run.
image

The biggest portion of that seems to be that all of the `vcpkg` dependencies are being built every time and not being cached, the  binary cache location doesn't seem to be working as expected.

This pull request puts the `binary` cache into the `GitHub Actions Cache`. After having a "primed" binary cache, the `Build and Test` workflow only takes 35 minutes.
image


mathisloge:
Nice that it's working now.
At the time of the setup, the cache was limited to 5GB

hummeltech:
Ahh, that makes sense. Yes, there is currently a 10GB limit to the `GitHub Actions Cache`. I don't believe it will come close to that.

hummeltech:
My fork of `mapnik/mapnik` is currently using only `2,430,000,543 bytes` with a full binary cache for the 3 tested OSes.

hummeltech:
Before (`1h 55m 26s`):
https://github.com/mapnik/mapnik/actions/runs/9667925553

After (`34m 47s`):
https://github.com/mapnik/mapnik/actions/runs/9669105888

mathisloge (approved):

PR #4493, opened 2025-02-12
Mapmeld cjk

artemp:
TextSymbolizer: `lang` attribute


mapmeld:
🤔  I'm open to `lang` property as an expression, I don't know if that makes this a degree more complicated

artemp:
> 🤔 I'm open to `lang` property as an expression, I don't know if that makes this a degree more complicated

@mapmeld - shouldn't be too complicated but will add expressions evaluation.. Might be still a good idea but should be separate PR anyway. I'm going to merge this PR, thanks again

Issue #4499, opened 2025-03-22
mapnik: build failure with libc++ and `BOOST_REGEX_HAS_ICU` enabled


bengsparks:
When building the library on nixpkgs for darwin platforms, I encounter the following output:

```
mapnik> /nix/store/yd23hbj32bmd7r6ldycf1986n4rx1gh5-libcxx-19.1.7-dev/include/c++/v1/string:820:42: error: implicit instantiation of undefined template 'std::char_traits<int>'
mapnik>   820 |   static_assert(is_same<_CharT, typename traits_type::char_type>::value,
mapnik>       |                                          ^
mapnik> /tmp/nix-build-mapnik-4.0.5.drv-0/source/src/expression_node.cpp:43:41: note: in instantiation of template class 'std::basic_string<int>' requested here
mapnik>    43 |     int32_t len = safe_cast<int32_t>(src.length());
mapnik>       |                                         ^
mapnik> /nix/store/yd23hbj32bmd7r6ldycf1986n4rx1gh5-libcxx-19.1.7-dev/include/c++/v1/__fwd/string.h:23:29: note: template is declared here
mapnik>    23 | struct _LIBCPP_TEMPLATE_VIS char_traits;
mapnik>       |                             ^
mapnik> In file included from /tmp/nix-build-mapnik-4.0.5.drv-0/source/src/expression_node.cpp:23:
mapnik> In file included from /tmp/nix-build-mapnik-4.0.5.drv-0/source/include/mapnik/expression_node.hpp:28:
mapnik> In file included from /tmp/nix-build-mapnik-4.0.5.drv-0/source/include/mapnik/value.hpp:29:
mapnik> In file included from /tmp/nix-build-mapnik-4.0.5.drv-0/source/include/mapnik/value/hash.hpp:27:
mapnik> In file included from /tmp/nix-build-mapnik-4.0.5.drv-0/source/include/mapnik/util/variant.hpp:27:
mapnik> In file included from /tmp/nix-build-mapnik-4.0.5.drv-0/source/deps/mapbox/variant/include/mapbox/variant.hpp:8:
mapnik> In file included from /nix/store/yd23hbj32bmd7r6ldycf1986n4rx1gh5-libcxx-19.1.7-dev/include/c++/v1/string:647:
[...]

PR #4500, opened 2025-03-30
Added fix in order to support `libc++ >= 19`

hummeltech:
Closes https://github.com/mapnik/mapnik/issues/4499

### DIFF
--- a/src/expression_node.cpp
+++ b/src/expression_node.cpp
@@ -38,10 +38,15 @@ MAPNIK_DISABLE_WARNING_POP
 namespace mapnik {
 
 #if defined(BOOST_REGEX_HAS_ICU)
-static void fromUTF32toUTF8(std::basic_string<UChar32> const& src, std::string& dst)
+static void fromUTF32toUTF8(boost::u32regex const& src, std::string& dst)
 {
-    int32_t len = safe_cast<int32_t>(src.length());
-    value_unicode_string::fromUTF32(src.data(), len).toUTF8String(dst);
+    boost::u32regex::iterator src_iterator = src.begin();
+    boost::u32regex::iterator src_end = src.end();
+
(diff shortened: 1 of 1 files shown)


bengsparks:
Ah I had tried a similar patch myself, but failed to get it working; very nice! Do you intend to release a new patch version, or should the PR's commit be picked? I can confirm that `mapnik` now builds again on my `aarch64-darwin` when this patch has been applied

hummeltech:
That depends on the maintainers, I suppose there might be a new release at some point, so using this as a patch until then is what I have done for FreeBSD.
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=285778

artemp (changes requested):
@hummeltech - Great catch, thanks! Creating an intermediate `std::string<UChar32>` was redundant. I have small suggestion to append whole `pattern` as it should be more efficient  

```c++ 
static void fromUTF32toUTF8(boost::u32regex const& src, std::string& dst)
{
    std::int32_t size = safe_cast<std::int32_t>(src.size());
    value_unicode_string::fromUTF32(src.expression(), size).toUTF8String(dst);
}
```

What do you think?

hummeltech:
Sure thing, thanks @artemp for the suggestion, that's much nicer!

bengsparks:
@hummeltech could you squash the commits in this PR for easier cherry-picking?

artemp (approved):

artemp:
> @hummeltech could you squash the commits in this PR for easier cherry-picking?

@bengsparks - sorry, didn't see you comment before merging.

bengsparks:
@artemp No worries; if a 4.0.7 release will be cut soon, then I can simply point to that release label :) do you know when that will be?

Issue #4508, opened 2025-05-16
Remove support for unquoted strings in expressions


artemp:
### Background 

`unquoted` strings in expressions are problematic - 

* https://github.com/mapnik/mapnik/pull/3389
* https://github.com/mapnik/mapnik/pull/3287 

### Solution 

Amend expression grammar to only allow `unquoted` strings as a complete expression e.g

* 🟢 `"Hello"`        - valid expression, equivalent to `"'Hello'"` 
* 🔴 `"Hello + '!'" ` - **invalid**! `Hello` is unquoted string inside `a + b` expression 
* 🟢 `"'Hello' + '!'" ` - OK, both strings are quoted
* 🔴 `"[int]+m"` - **invalid**
* 🟢 `"[int]+'m'"` - OK

PR #4509, opened 2025-05-16
Evaluate 'lang' parameter in `TextSymbolizer` + amend expressions grammar per #4508 [WIP]

artemp:
https://github.com/mapnik/mapnik/pull/4493 
#4508

/cc @mapmeld


mapmeld:
The test makes sense to me, if you could generate the reference images it'd be easier to be sure it's working

PR #4514, opened 2025-08-02
Fix two compile issues with mingw

jia-kai:
Hi,

This PR fixes two compile issues with MINGW:

1. Result type of `dlsym` is `FARPROC` on WIN32, which can not be converted via `static_cast`
2. `std::ifstream(std::wstring)` is not an std interface ([see here](https://en.cppreference.com/w/cpp/io/basic_ifstream/basic_ifstream.html)); we need to use `filesystem::path` to wrap the path

The fixed code now compiles with mingw 15.1.0 by cross-compile from Linux. I don't have an MSVC environment at hand so I don't know if it breaks any existing things. If necessary, I can wrap the fix with MINGW guards.

### DIFF
--- a/include/mapnik/util/utf_conv_win.hpp
+++ b/include/mapnik/util/utf_conv_win.hpp
@@ -28,6 +28,7 @@
 #include <mapnik/config.hpp>
 // stl
 #include <string>
+#include <filesystem>
 
 namespace mapnik {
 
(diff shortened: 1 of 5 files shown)


jia-kai:
macos ARM64 build failed with `CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.`  Seems irrelevant here.

artemp (approved):

artemp:
@jia-kai - looks good to me, thanks for your contribution.

PR #4542, opened 2025-12-11
Combined input plugins

artemp:
Combine multiple data sources (that share the same third party dependencies e.g gdal/ogr, postgis/pgraster) in one shared library to minimise binary size of input plug-ins. 

```
gdal.input + ogr.input =>  gdal+ogr.input 
```


hummeltech:
Nice work @artemp, I love it!

PR #4554, opened 2026-05-04
Rendering complex Unicode sequences across multiple fonts in a fontset

astridx:
There is a rendering issue in [openstreetmap-carto](https://github.com/openstreetmap-carto/openstreetmap-carto/issues/5215) and I think this PR solves it.

With this PR, the cluster containing `ශ්‍රී` is rendered using a single font that fully supports the glyph sequence.

If I understand correctly, without this PR, `ශ්‍රී` is rendered using the first font in the fontset that maybe only partially supported the cluster, resulting in inconsistent shaping.

My Test case `hellosinhala.xml`:

```xml
<?xml version="1.0" encoding="utf-8"?>
<Map srs="EPSG:3857">

  <FontSet name="fallback">
    <Font face-name="Noto Sans Sinhala Regular"/>
  </FontSet>

  <Style name="s">
    <Rule>
      <TextSymbolizer fontset-name="fallback"
                       size="32"
                       fill="black">
        [name]
      </TextSymbolizer>
    </Rule>
  </Style>

  <Layer name="t" srs="EPSG:4326">
    <StyleName>s</StyleName>

    <Datasource>
      <Parameter name="type">geojson</Parameter>
      <Parameter name="inline">
        {
          "type": "FeatureCollection",
          "features": [
            {
              "type": "Feature",
              "properties": { "name": "Hello ශ්‍රී ලංකා" },
              "geometry": { "type": "Point", "coordinates": [0, 0] }
            }
          ]
        }
      </Parameter>
    </Datasource>

  </Layer>
</Map>
```

I rendered it via:

```bash
mapnik-render --bbox -10,-10,10,10 \
  --plugins-dir /usr/local/lib/mapnik/input \
  --fonts-dir /usr/share/fonts/truetype/noto/ \
  hellosinhala.xml output.png
```

### Note

This is my first time with Mapnik, so I am still learning. Because of that, I am not fully confident yet about negative side effects.

I compiled Mapnik on Debian 13 using:

```bash
git clone https://github.com/mapnik/mapnik.git --depth 1 --recursive
cd mapnik

sudo rm -rf /usr/local/lib/mapnik
sudo rm -rf /usr/local/include/mapnik

rm -rf build
mkdir build && cd build

cmake .. \
  -DCMAKE_BUILD_TYPE=Debug \
  -DBUILD_BENCHMARK=OFF \
  -DBUILD_TESTING=OFF \
  -DBUILD_DEMO_VIEWER=OFF

make -j2
sudo make install
```
 
For debugging, I added temporary logging (I was not able to make MAPNIK_LOG_DEBUG visible):

```cpp
if (c.empty()) {
    std::cerr << "Cluster " << cluster << " empty, add glyph font "
              << face->family_name() << std::endl;
    c.push_back({face, glyphs[i], positions[i]});
}
else if (c.front().face != face) {
    std::cerr << "Cluster " << cluster << " switch font "
              << c.front().face->family_name()
              << " -> " << face->family_name() << std::endl;

    c.clear();
    c.push_back({face, glyphs[i], positions[i]});
}
else if (c.front().glyph.codepoint == 0) {
    std::cerr << "Cluster " << cluster << " replace missing glyph font "
              << face->family_name() << std::endl;

    c.front() = {face, glyphs[i], positions[i]};
}
else if (in_cluster) {
    std::cerr << "Cluster " << cluster << " add glyph font "
              << face->family_name() << std::endl;

    c.push_back({face, glyphs[i], positions[i]});
}
```

### DIFF
--- a/include/mapnik/text/harfbuzz_shaper.hpp
+++ b/include/mapnik/text/harfbuzz_shaper.hpp
@@ -345,6 +345,11 @@ struct harfbuzz_shaper
                     {
                         c.push_back({face, glyphs[i], positions[i]});
                     }
+                    else if (c.front().face != face)
+                    {
+                        c.clear();
+                        c.push_back({face, glyphs[i], positions[i]});
+                    }
                     else if (c.front().glyph.codepoint == 0)
                     {
                         c.front() = {face, glyphs[i], positions[i]};


artemp:
@astridx - 👍 

>negative side effects 

Let me see if I can spot anything but otherwise looking good.

Thanks for your contribution!

artemp:
@astridx  I was testing your patch with Sinhala country names (see below how I generated geojson) and I found that "NotoSansSinhalaUI-Regular.ttf" (this is what I had locally) was causing bogus `shaping` (e.g overlapping glyphs). I downloaded latest [NotoSansSinhala](
https://fonts.google.com/noto/specimen/Noto+Sans+Sinhala?preview.layout=grid&lang=si_Sinh&preview.script=Sinh&preview.lang=si_Sinh) and rendering looks OK to me but I'd appreciate if you could verify that's the case, thanks.

sinhala-fixed



* https://github.com/pasan93/Sinhala-Country-List/blob/main/Translated_Sinhala_Country_List.json
* https://github.com/datasets/geo-countries/blob/main/data/countries.geojson 
* make_sinhala_geojson.py 
```python
import json

if __name__ == "__main__":

    sinhala_countries = json.load(open("Translated_Sinhala_Country_List.json"))

    sinhala_names = {}
    for c in sinhala_countries:
        name = c['name']
        code = c['code']
        sinhala_names[c['code']] = c['name']

    world = json.load(open("countries.geojson"))
    for c in world['features']:
        code = c['properties']['ISO3166-1-Alpha-2']
        if code in sinhala_names:
            print(code)
            c['properties']['name']=sinhala_names[code]
        else:
            c['properties']['name']=""

    sinhala_world_json = open("sinhala_countries.geojson","w")
    json.dump(world, sinhala_world_json)
```

```xml
<?xml version="1.0" encoding="utf-8"?>
<Map srs="EPSG:3857" background-color="skyblue">
  <FontSet name="fallback">
    <Font face-name="Noto Sans Sinhala Regular"/>
  </FontSet>
  <Style name="polygons">
    <Rule>
      <PolygonSymbolizer fill="beige" fill-opacity="0.9"/>
      <LineSymbolizer stroke="black" stroke-opacity="0.2"/>
    </Rule>
  </Style>
  <Style name="s">
    <Rule>
      <TextSymbolizer fontset-name="fallback" size="12" fill="black" halo-fill="rgba(255,255,255,0.3)" halo-radius="1">
        [name]
      </TextSymbolizer>
    </Rule>
  </Style>

  <Layer name="t" srs="EPSG:4326">
    <StyleName>polygons</StyleName>
    <StyleName>s</StyleName>

    <Datasource>
      <Parameter name="type">geojson</Parameter>
      <Parameter name="file">sinhala_countries.geojson</Parameter>
    </Datasource>
  </Layer>
</Map>
```

astridx:
@artemp Thank you very much for testing this.

I tested your code on a freshly installed Debian 13 system, so these should be the latest available stable fonts for `core` and `ui`.

```
mapnik-render --bbox 3000000,-1000000,17000000,9000000  --plugins-dir /usr/local/lib/mapnik/input --fonts-dir /usr/share/fonts/truetype/noto/  input.xml output.png
```


```bash
sudo apt update
sudo apt install fonts-noto-core fonts-noto-ui-core
````

## 1. I tested everything again using your files

```xml
<FontSet name="fallback">
  <Font face-name="Noto Sans Sinhala Regular"/>
</FontSet>
```

output_1_original

## 2. Then I removed Sinhala completely

This was to make sure that fonts are not being picked up implicitly from the font path even if they are not listed in the fallback set.

```xml
<FontSet name="fallback">
</FontSet>
```

output_2_nur_NotoRegular

## 3. Then I used Sinhala UI in the fallback set

```xml
<FontSet name="fallback">
  <Font face-name="Noto Sans Sinhala UI Regular"/>
</FontSet>
```
output_3_NotoSinhalaUIRegular


The issue with partially displayed or overlapping/combined characters only occurred before the changes in this PR.

output_hello

or 

output_3_NotoSinhalaUIRegular_ohnepatch


If I understand correctly, parts of combined glyphs were previously found in a different font. Therefore, the `c.clear();`, so that fonts which do not contain all required glyphs are removed from the fallback array.

artemp (approved):

ramSeraph:
I ran into the same problem as mentioned in https://github.com/openstreetmap-carto/openstreetmap-carto/issues/5215 elsewhere. 

I think the problem was that  in the default openstreetmap-carto font-set, Arabic font was being picked for the first Sinhalese cluster because the check to see if a cluster is covered( in the existing code ) was to see if the first glyph ~~index~~ codepoint is valid and in the case of the ශ්‍රී(Sri) cluster.. the Arabic font which is ahead of Sinhalese font in the font-set provides a valid glyph ~~index~~ codepoint for the first glyph in the cluster( though it does not completely cover it ) and gets picked. 

The solution would be to check if all the glyph ~~indices~~ codepoints in the cluster are valid before considering a cluster as covered, I think. 

That said, the assumption that the glyph clusters have stable boundaries across fonts for a given text might itself be invalid, although the fact that we use the default HB_BUFFER_CLUSTER_LEVEL_MONOTONE_GRAPHEMES as the cluster level for the buffer does mitigate some of that risk. 

[Added]
This [comment](https://github.com/harfbuzz/harfbuzz/issues/1957#issuecomment-529130896) from the harfbuzz folks seems to suggest that font coverage should be checked at the character level and not at the glyph cluster level though font assignments should be done at glyph cluster level.  It does feel complicated and there is chance I have confused myself.

ramSeraph:
@artemp Is there any appetite for a revamp of the font fallback code using the 2nd suggested method from the [harfbuzz issue](https://github.com/harfbuzz/harfbuzz/issues/1957#issuecomment-529130896)? I think I have the code.

artemp:
> @artemp Is there any appetite for a revamp of the font fallback code using the 2nd suggested method from the [harfbuzz issue](https://github.com/harfbuzz/harfbuzz/issues/1957#issuecomment-529130896)? I think I have the code.

@ramSeraph - Yes, sure. My understanding is that @astridx solved this in https://github.com/mapnik/mapnik/pull/4556, no? In any case improvements/fixes most welcome!

ramSeraph:
Yep.. I think @astridx has already solved this particular issue in the latest PR.

What I plan to add would be a re-imagination of the fallback setup with coverage checks done at character level and not depending on cluster id stability across shaping calls. Very similar to how Chromium/Blink implements it.  We can discuss on the PR when I send it.

Issue #4555, opened 2026-05-08
Font Rendering Issues with Bracket Layers


artemp:
Mismatched "brackets" rendering (glyphs chosen from different fonts) after https://github.com/mapnik/mapnik/pull/4554 

<img width="387" height="242" alt="Image" src="https://github.com/user-attachments/assets/6ca274d1-7f8e-43b9-bdc2-8b2132750ac1" />

/cc @astridx  - interesting artifact after https://github.com/mapnik/mapnik/pull/4554  

```xml
<?xml version="1.0" encoding="utf-8"?>
<Map srs="EPSG:3857">

  <FontSet name="fallback">
    <Font face-name="Noto Sans Regular"/>
    <Font face-name="Noto Serif Tibetan Regular"/>
    <Font face-name="Noto Sans Sinhala Regular"/>
  </FontSet>

  <Style name="s">
    <Rule>
      <TextSymbolizer fontset-name="fallback"
                       size="32"
                       fill="black" allow-overlap="true">
        [name]
      </TextSymbolizer>
    </Rule>
  </Style>

  <Layer name="t" srs="EPSG:4326">
    <StyleName>s</StyleName>

    <Datasource>
      <Parameter name="type">geojson</Parameter>
      <Parameter name="inline">
        {
          "type": "FeatureCollection",
          "features": [
          {
              "type": "Feature",
              "properties": { "name": "སྤུ་ཧྲེང (abc)" },
              "geometry": { "type": "Point", "coordinates": [0, 0] }
           },
           {
              "type": "Feature",
              "properties": { "name": "Hello ශ්‍රී ලංකා (abc)" },
              "geometry": { "type": "Point", "coordinates": [0, -4] }
            }
          ]
        }
      </Parameter>
    </Datasource>

  </Layer>
</Map>
```

astridx:
I am not sure, so I want to ask first. My idea is that the issue might be related to shared Unicode characters in a single run:

https://github.com/mapnik/mapnik/compare/master...astridx:mapnik:itemizercpp

https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/uchar_8h.html#add0409f1e6cbbc84dda50c45cc3e7302
https://www.unicode.org/reports/tr44/#General_Category_Values
https://www.unicode.org/Public/UCD/latest/ucd/Scripts.txt

artemp:
[...]

PR #4556, opened 2026-05-12
Correction for "Rendering complex Unicode sequences across multiple fonts in a fontset"

astridx:
After thinking, I believe this solution is much better:

> The solution would be- to check if all the glyph indices codepoints in the cluster are valid before considering a cluster as covered, I think.  https://github.com/mapnik/mapnik/pull/4554#issuecomment-4428077057

### DIFF
--- a/include/mapnik/text/harfbuzz_shaper.hpp
+++ b/include/mapnik/text/harfbuzz_shaper.hpp
@@ -345,6 +345,28 @@ struct harfbuzz_shaper
                     current_clusters[cluster]
                         .push_back({face, glyphs[i], positions[i]});
                 }
+                for (unsigned cluster_id = 0;
+                    cluster_id < current_clusters.size();
+                    ++cluster_id)
+                {
+                    auto const& cluster_glyphs = current_clusters[cluster_id];
+
+                    if (cluster_glyphs.empty())
+                        continue;
+                    bool valid = true;
+                    for (auto const& info : cluster_glyphs)
+                    {
+                        if (info.glyph.codepoint == 0)
+                        {
+                            valid = false;
+                            break;
+                        }
+                    }
+                    if (valid)
+                    {
+                        glyphinfos[cluster_id] = cluster_glyphs;
+                    }
+                }
                 bool all_set = true;
                 for (auto c_id : clusters)
                 {
--- a/include/mapnik/text/harfbuzz_shaper.hpp
+++ b/include/mapnik/text/harfbuzz_shaper.hpp
@@ -366,6 +366,10 @@ struct harfbuzz_shaper
                     {
                         glyphinfos[cluster_id] = cluster_glyphs;
                     }
+                    else if (glyphinfos[cluster_id].empty())
+                    {
+                        glyphinfos[cluster_id] = cluster_glyphs;
+                    }
                 }
                 bool all_set = true;
                 for (auto c_id : clusters)
--- a/include/mapnik/text/harfbuzz_shaper.hpp
+++ b/include/mapnik/text/harfbuzz_shaper.hpp
@@ -318,6 +318,8 @@ struct harfbuzz_shaper
                 unsigned cluster = 0;
                 bool in_cluster = false;
                 std::vector<unsigned> clusters;
+                std::vector<std::vector<glyph_face_info>> current_clusters;
+                current_clusters.resize(text.length());
 
                 for (unsigned i = 0; i < num_glyphs; ++i)
                 {
--- a/include/mapnik/text/harfbuzz_shaper.hpp
+++ b/include/mapnik/text/harfbuzz_shaper.hpp
@@ -342,12 +342,9 @@ struct harfbuzz_shaper
                     {
                         glyphinfos.resize(cluster + 1);
                     }
-                    current_clusters[cluster]
-                        .push_back({face, glyphs[i], positions[i]});
+                    current_clusters[cluster].push_back({face, glyphs[i], positions[i]});
                 }
-                for (unsigned cluster_id = 0;
-                    cluster_id < current_clusters.size();
-                    ++cluster_id)
+                for (unsigned cluster_id = 0; cluster_id < current_clusters.size(); ++cluster_id)
                 {
                     auto const& cluster_glyphs = current_clusters[cluster_id];
 
--- a/include/mapnik/text/harfbuzz_shaper.hpp
+++ b/include/mapnik/text/harfbuzz_shaper.hpp
@@ -342,19 +342,8 @@ struct harfbuzz_shaper
                     {
                         glyphinfos.resize(cluster + 1);
                     }
-                    auto& c = glyphinfos[cluster];
-                    if (c.empty())
-                    {
-                        c.push_back({face, glyphs[i], positions[i]});
-                    }
-                    else if (c.front().glyph.codepoint == 0)
-                    {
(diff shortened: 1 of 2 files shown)


artemp (approved):

astridx:
@artemp Thanks.

I’ve read the comment https://github.com/mapnik/mapnik/pull/4557#issuecomment-4487182824, but I don’t see any overriding of fonts that support every part. A wrong font from the fallback stack gets inserted only in cases where no font that supports everything is available —at least if I’ve reasoned it through correctly.

I would like to add a test in https://github.com/mapnik/test-data-visual, but I’m struggling a bit with the testing setup. The `bootstrap.sh` form https://github.com/mapnik/mapnik/blob/master/docs/contributing.md doesn’t exist anymore, does it? I’d appreciate a link to some instructions. 

I keep ending up at the same point.

```
88/89 Test #88: font .....................................................   Passed   19.00 sec
      Start 89: map xml I/O
89/89 Test #89: map xml I/O ..............................................***Failed    0.74 sec

99% tests passed, 1 tests failed out of 89

Total Test time (real) =  66.33 sec

The following tests FAILED:
	 89 - map xml I/O (Failed)
Errors while running CTest
Output from these tests are in: /home/astrid/mapnik/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
make: *** [Makefile:91: test] Fehler 8

```

and

```
  REQUIRE_NOTHROW( mapnik::load_map(m, xml.generic_string(), false, abs_base.generic_string()) )
due to unexpected exception with messages:
  path.generic_string() := "test/data/good_maps/marker_ellipse_transform2.xml"
  Could not create datasource for type: 'csv' (searched for datasource plugins
  in 'plugins/input')  encountered during parsing of layer 'layer' in Layer at
  line 33 of 'test/data/good_maps/marker_ellipse_transform2.xml'

Missing colon in +init
proj_create: cannot expand +init=invalid_srs_value
Missing colon in +init
proj_create: cannot expand +init=invalid_srs_value
===============================================================================
test cases:  1 |  0 passed | 1 failed
assertions: 69 | 68 passed | 1 failed

<end of output>
Test time =   0.74 sec
----------------------------------------------------------
Test Failed.
"map xml I/O" end time: May 31 10:38 CEST
"map xml I/O" time elapsed: 00:00:00
----------------------------------------------------------

End testing: May 31 10:38 CEST

```

artemp:
@astridx  - 

```
Could not create datasource for type: 'csv' (searched for datasource plugins
  in 'plugins/input')  encountered during parsing of layer 'layer' in Layer at
  line 33 of 'test/data/good_maps/marker_ellipse_transform2.xml'
``` 

Have you compiled `csv.input` ^^ ? 


re: bootstrap.sh  - https://github.com/mapnik/mapnik/blob/fb2e45c57981f8a3b071f37a0b27f211bf233081/bootstrap.sh

ramSeraph:
> but I don’t see any overriding of fonts that support every part. 

@astridx That is not what I meant.. 

So, if we have a string "<LangA_str> <LangB_str> <LangC_str>"  and a font stack of 
  * FontA
  * FontB
  * FontF 
 
and the FontA covers only <LangA_str> and FontB only covers <LangB_str> and fallback font( FontF ) covers the whole string, you are overriding <LangA_str> and <LangB_str> with FontF as well. Unless I am mistaken this goes against the contract of font fallback  and why FontA and FontB were put ahead of FontF in the fallback list in the first place. 

And styles like OSM Carto have picked the font stack based on recommendations from Google and I think the recommendation assumes this contract.

astridx:
@ramSeraph Thanks for your explanation. I wasn’t aware of the contract of font fallback. Can you recommend where I can read more about it?

ramSeraph:
@astridx I think this would be piece of documentation for it - https://www.w3.org/TR/css-fonts-3/#font-style-matching
It talks mostly of matching at the character/cluster level. 

This is assuming we are trying to follow CSS semantics as closely as possible given what we can express with the Mapnik style language.

ramSeraph:
## Some notes on the USCRIPT_COMMON changes

### The original problem

The script run segmentation code that Mapnik adapted from ICU, keeps track of brackets and assigns the same script to matching brackets. This was to make sure that they get rendered with the same font. This makes the assumption that  given the same font-set,  whatever text accompanies the bracket, as long as the script set is the same it will be assigned the same font. 

This was previously fine because the first font covering the bracket cluster( which is usually just the bracket ) was being picked and that's stable. But with the current changes even if a font covered a cluster but a later font also covers it( while the whole string is not covered ), we are overriding that cluster's font choice and this breaks the assumption of stability made by the brackets script assignment code. This explains the failures seen in #4555.

I will demonstrate the instability with an example:

The script segmentation( without the USCRIPT_COMMON adjustments ) for `Hello ශ්‍රී ලංකා (abc)` was:

```
[0,6)   Latn  -> "Hello "
[6,18)  Sinh  -> "ශ්‍රී ලංකා ("
[18,21) Latn  -> "abc"
[21,22) Sinh  -> ")"
```

Now given the following font-set
```
  <FontSet name="fallback">
    <Font face-name="Noto Serif Tibetan Regular"/>
    <Font face-name="Noto Sans Sinhala Regular"/>
  </FontSet>
```

The  `[21,22) Sinh  -> ")"` always gets "Noto Sans Regular"

But the `[6,18)  Sinh  -> "ශ්‍රී ලංකා ("` section was previously picking 
```
[6,11)    "ශ්‍රී"       -> Noto Sans Sinhala Regular
[11,12)   " "        -> Noto Sans Regular
[12,14)  "ලං"     -> Noto Sans Sinhala Regular
[14,16)  "කා"     -> Noto Sans Sinhala Regular
[16,17)  " "        -> Noto Sans Regular
[17,18)  "("        -> Noto Sans Regular
```
Opening and closing brackets end up with the same font. 


The current code picks:
```
[6,11)    "ශ්‍රී"      -> Noto Sans Sinhala Regular
[11,12)  " "        -> Noto Sans Sinhala Regular
[12,14) "ලං"     -> Noto Sans Sinhala Regular
[14,16) "කා"     -> Noto Sans Sinhala Regular
[16,17) " "        -> Noto Sans Sinhala Regular
[17,18) "("        -> Noto Sans Sinhala Regular
```
So, the opening and closing brackets end up with different fonts. 

### The USCRIPT_COMMON fix

The fix that went in for this problem( in the current PR ) handled this by post-processing the script runs and assigning USCRIPT_COMMON to the punctuation.. which in this case comes out as:
```
[0,6)   Latn   -> "Hello "
[6,17)  Sinh   -> "ශ්‍රී ලංකා "
[17,18)  Common -> "("
[18,21)  Latn   -> "abc"
[21,22)  Common -> ")"
```
I think this will fix a lot of cases including the one above but this is still incomplete in that, the problem will persist in multi bracket runs like `[(text1)text2]` and fontstack of `fontA; fontB`  if the `[` brackets are not available in `fontA` but available in `fontB`..  and the `(` brackets are available in `fontA` and `fontB`. In this case the `(` will be rendered with `fontB` and `)` will be rendered with `fontA`.  This is a theoretical scenario, I haven't tested it.  But for completeness we could limit the punctuation runs to a length of 1.  

### Conclusions:

In general, if we don't assert a later font on a cluster which was already covered by an earlier one.. the USCRIPT_COMMON change is not required.  
 


I am mostly bringing this up because I am going to undo it in #4557 and wanted to explain why.

PR #4557, opened 2026-05-15
Improve HarfBuzz based font fallback behaviour

ramSeraph:
## Summary
 
 This PR refines Mapnik's HarfBuzz fallback behavior and expands regression
 coverage around complex-script fallback, cluster coverage, and RTL handling.
 
 ## Main changes
 
Rework fallback coverage tracking around UTF-16 source positions inline with suggestions from [Harfbuzz folks](https://github.com/harfbuzz/harfbuzz/issues/1957#issuecomment-529130896).

### Algorithm

 The fallback implementation is shaping-driven and hole-based.
 
 We start with the full `text_item` as one uncovered logical range. For each face
 in the fallback list, we shape only the currently uncovered ranges, inspect the
 result cluster-by-cluster, and accept only clusters that are actually covered by
 that face. Accepted clusters are appended to `glyphinfos`, and a per-UTF-16
 coverage map records which appended cluster covers each source position.
 Rejected clusters are turned back into uncovered logical ranges and retried with
 the next face.
 
 Coverage is stored in logical/text order for both LTR and RTL. ~~For RTL, the
 temporary HarfBuzz cluster list is reversed back into text order before
 coverage is recorded~~ When HarfBuzz emits RTL clusters in reverse/visual order, 
 we build the temporary cluster list directly into logical order as we walk the
 shaped glyphs, so coverage recording uses the same logical-order path in both directions. 
 Final emission then walks the logical coverage map
 forward for LTR and backward for RTL, skipping holes and collapsing repeated
 entries, so output still comes out in visual order. 

 ## Regression coverage
 
 - add deterministic fallback fixtures for Sinhala, Khmer, and Myanmar cases
 - expand `test/unit/text/shaping.cpp` with regressions covering:
   - Sinhala fallback
   - Khmer fallback
   - Myanmar fallback
   - RTL fallback behavior
   - interleaving accepted ranges
   - ligature shaping
   - last-face uncovered/tofu behavior


ramSeraph:
There were two kinds of probably related failures in the visual tests.. 

Screenshot 2026-05-15 at 8 14 07 PM
Screenshot 2026-05-15 at 8 14 35 PM

The difference in the number of tofu markers can probably be explained by the fact that the older code was removing the empty codepoints in front of invalid clusters. And I am guessing the complete disappearance  in the second set of failures might be explainable by it being of a different width. Needs more investigation though.

ramSeraph:
Blink source code was used to compare if I got the implementation right.. 
the deque for uncovered spans was inspired by Blink.  

And as part of the research into the Blink code the following edge cases which Blink handles have come up. 
I haven't implemented them yet as I am not sure if any of them are required for our use case in maps.

## Font fallback related

### U+3000 IDEOGRAPHIC SPACE synthesized from ASCII space

HarfBuzz can synthesize `U+3000 IDEOGRAPHIC SPACE` from a font's ordinary space
glyph. When that happens, the current face appears to "cover" the cluster even
though it does not provide the real CJK glyph metrics or GPOS behavior.

In practice this can:

- hide the correct fullwidth advance of the fallback CJK face
- skip font-specific positioning behavior for `U+3000`
- split shaping/fallback decisions in the wrong place

Blink treats this as an edge case where fallback should continue when all of
the following are true:

- the shaped glyph id is the current face's ordinary space glyph
- the source character at the cluster start is `U+3000`
- ~~the current face is not the last fallback face~~( irrelevant for us ).

Related HarfBuzz issue:

- <https://github.com/harfbuzz/harfbuzz/issues/2889>

Related Blink issue:

- <https://crbug.com/1193282>

### Unmatched variation selector fallback

Some faces cover the base character of a variation sequence but do not honor
the requested variation selector. Accepting that shaping result would stop
fallback too early and lose the intended glyph choice.

Blink treats this as a fallback edge case: a face that shapes the base
character but does not honor the requested variation selector should not
necessarily terminate fallback for that cluster.

Related Blink issue:

- <https://issues.chromium.org/issues/40856329>

### Font fallback hints

Blink uses hints based on heuristics to let it pick preferred fonts for script runs.
This feels very browserland specific and we probably don't need it. 
That said.. we might want to look into the problem below which 
they ran into while picking hints for Myanmar script. 

Blink avoids relying on leading punctuation or other
`Common`/`Inherited` script characters as the primary fallback hint when it can
find a later character with a more specific script.

This matters because a run that starts with punctuation can otherwise choose a
poor fallback font. This is related to a bug they found while handling Myanmar script.

Related Blink issues:

- <https://crbug.com/618178>

## Script runs related

### East Asian punctuation is rewritten to Han-family scripts

Blink's `ScriptRunIterator` rewrites some `Common` punctuation to Han-family
script extensions based on East Asian Width. This keeps consecutive CJK
punctuation in the same run and allows CJK font features to apply across the
sequence.

The motivating problem is that some East Asian punctuation characters,
especially compatibility forms, otherwise look like plain `Common` script and
can split runs incorrectly.

Related Chromium issue:

- <https://issues.chromium.org/issues/40198734>


### Paired brackets are resolved with a bracket stack and Han exceptions there. 

Blink keeps a stack of open brackets during script resolution. This might be relevant for #4555

- opening brackets are pushed and their final script assignment is deferred
- when a matching closing bracket is seen, Blink tries to reuse the opening
  bracket's resolved script. Mapnik has similar code for bracket handling till here.
- for Han-family runs, Blink allows the closing bracket to use the current Han
  script instead of insisting on the exact opener script, which helps avoid
  breaking consecutive CJK punctuation runs

ramSeraph:
ccing @1ec5 for input on the relevance of the above edge cases for map labels. Only if you have the time for it.. ofcourse.

1ec5:
I assume the overall impact of addressing these edge cases is to give designers a wider repertoire of compatible fonts to work with. Fortunately, the most widespread fonts for map labeling don’t necessarily need all these workarounds, but this certainly could benefit less resourced languages and other use cases that require specialty fonts.

> Blink uses hints based on heuristics to let it pick preferred fonts for script runs.
> This feels very browserland specific and we probably don't need it.

I figure some of the special handling is because a browser engine needs to handle a mix of languages in the same run of text. This can also happen in map labels if implementing dual language labels or [OSM’s compound multilingual names](https://wiki.openstreetmap.org/wiki/Multilingual_names), but it’s still a bit more predictable than Web text in general.

> Blink's ScriptRunIterator rewrites some Common punctuation to Han-family
> script extensions based on East Asian Width. This keeps consecutive CJK
> punctuation in the same run and allows CJK font features to apply across the
> sequence.

MapLibre does something similar in vertical text layout out of necessity: vertical text is universally expected to hew very closely to the character grid, but halfwidth glyphs break the grid. I hadn’t considered making it also substitute fullwidth glyphs in horizontal text, because conventions seem to differ by language. `zh-Hans-CN` seems to be more accepting than `ja` of misaligned characters in horizontal text. MapLibre currently only implements [fullwidth for punctuation](https://github.com/maplibre/maplibre-gl-js/blob/9f57bd8e19ac4df4494120eb62589c866b9e968c/src/util/verticalize_punctuation.ts), as _tate-chu-yoko_ (to use the Japanese term) is still unimplemented: maplibre/maplibre-gl-js#5404.

There are different visual forms for the same punctuation mark in horizontal versus vertical text. MapLibre has hard-coded a mapping to certain compatibility forms for brackets and some other punctuation marks in vertical text only. It may be that HarfBuzz already handles those edge cases well without the need for compatibility forms.

ramSeraph:
> > Blink uses hints based on heuristics to let it pick preferred fonts for script runs.
> >This feels very browserland specific and we probably don't need it.

> I figure some of the special handling is because a browser engine needs to handle a mix of languages in the same run 
> of text. This can also happen in map labels if implementing dual language labels or [OSM’s compound multilingual 
> names](https://wiki.openstreetmap.org/wiki/Multilingual_names), but it’s still a bit more predictable than Web text in general.

This felt like it was more about performance than correctness given how complex the font stack browsers have to deal with is. They seem to have a more layered fallback like: 
 * CSS / font-group fonts
 * segmented-face / similar targeted sources
 * preference fonts
 * generic system fallback fonts
 * last-resort / .notdef
 
EDIT: It might be about correctness for the segmented-face case.. not sure though.

ramSeraph:
I have just realized while reviewing the LTR and RTL constants that Mapnik does not support vertical ~~scripts~~ writing. Did I miss something or is this intentional?

ramSeraph:
> I have just realized while reviewing the LTR and RTL constants that Mapnik does not support vertical ~scripts~ writing. Did I miss something or is this intentional?

Will add a separate issue for this as it is clear to me now that vertical writing is indeed not supported.

artemp:
> The difference in the number of tofu markers can probably be explained by the fact that the older code was removing the empty codepoints in front of invalid clusters. And I am guessing the complete disappearance in the second set of failures might be explainable by it being of a different width. Needs more investigation though.


@ramSeraph - Correct ^. Might be a good idea to fix this test so all labels are rendered.

artemp:
@ramSeraph - thanks for PR!

ramSeraph:
> Might be a good idea to fix this test so all labels are rendered.

Will do. Doing some sanity tests against the OSM planet dump and OSM Carto fontstacks.. will update the tests with anything else I find.

ramSeraph:
We now have a concrete failure which this PR fixes.  This is a sample among a couple of thousand I found In OSM name/name:* tags. 

Test Case:
```xml
<?xml version="1.0" encoding="utf-8"?>
<Map srs="EPSG:4326"
     background-color="white"
     font-directory="openstreetmap-carto/fonts">
  <FontSet name="fallback">
    <Font face-name="Noto Sans Bengali UI Regular"/>
    <Font face-name="Noto Sans Symbols Regular"/>
  </FontSet>

  <Style name="s">
    <Rule>
      <TextSymbolizer fontset-name="fallback"
                      size="40"
                      fill="#111111"
                      halo-fill="#ffffff"
                      halo-radius="2"
                      placement="point"
                      allow-overlap="true"
                      horizontal-alignment="middle"
                      vertical-alignment="middle">[name]</TextSymbolizer>
    </Rule>
  </Style>

  <Layer name="t" srs="EPSG:4326">
    <StyleName>s</StyleName>
    <Datasource>
      <Parameter name="inline"><![CDATA[
x,y,name
0,0,"Bénestroff − Centre"
]]></Parameter>
    </Datasource>
  </Layer>
</Map>
```
Command:
```
/Users/ram/Code/mapnik_temp/build-visual-homebrew/out/mapnik-render \
   --bbox=-10,-2,10,2 \
   --map-width 600 \
   --map-height 150 \
   --plugins-dir /Users/ram/Code/mapnik_temp/build-visual-homebrew/out/plugins/input \
   /tmp/benestroff-bus-stop.xml /tmp/benestroff-bus-stop-current.png
```
Output( master~1 ~~392e3a0ee8833262e961405ccd6db6392196d0d8~~ fc669e25500cb696689e4a819c17cd09ff1ffe5e ):
benestroff-bus-stop-master-minus1

Output current PR:
benestroff-bus-stop-current

Source is https://www.openstreetmap.org/node/3890514044 though I don't see any problems in the rendering there.

Would have tested against #4556 but that PR is overriding already covered fonts with later fonts.. so not doing that comparison.  @astridx if this PR doesn't go through, you might want to change that.

ramSeraph:
OK.. now coming to problems found with current approach.

In the samples below #4556  refers to the PR with the some additional changes.. full diff below.

```diff
diff --git a/include/mapnik/text/harfbuzz_shaper.hpp b/include/mapnik/text/harfbuzz_shaper.hpp
index c6b31ad..5fa46fb 100644
--- a/include/mapnik/text/harfbuzz_shaper.hpp
+++ b/include/mapnik/text/harfbuzz_shaper.hpp
@@ -318,6 +318,8 @@ struct harfbuzz_shaper
                 unsigned cluster = 0;
                 bool in_cluster = false;
                 std::vector<unsigned> clusters;
+                std::vector<std::vector<glyph_face_info>> current_clusters;
+                current_clusters.resize(text.length());

                 for (unsigned i = 0; i < num_glyphs; ++i)
                 {
@@ -340,23 +342,26 @@ struct harfbuzz_shaper
                     {
                         glyphinfos.resize(cluster + 1);
                     }
-                    auto& c = glyphinfos[cluster];
-                    if (c.empty())
-                    {
-                        c.push_back({face, glyphs[i], positions[i]});
-                    }
-                    else if (c.front().face != face)
-                    {
-                        c.clear();
-                        c.push_back({face, glyphs[i], positions[i]});
-                    }
-                    else if (c.front().glyph.codepoint == 0)
+                    current_clusters[cluster].push_back({face, glyphs[i], positions[i]});
+                }
+                for (unsigned cluster_id = 0; cluster_id < current_clusters.size(); ++cluster_id)
+                {
+                    auto const& cluster_glyphs = current_clusters[cluster_id];
+
+                    if (cluster_glyphs.empty() || !glyphinfos[cluster_id].empty())
+                        continue;
+                    bool valid = true;
+                    for (auto const& info : cluster_glyphs)
                     {
-                        c.front() = {face, glyphs[i], positions[i]};
+                        if (info.glyph.codepoint == 0)
+                        {
+                            valid = false;
+                            break;
+                        }
                     }
-                    else if (in_cluster)
+                    if (valid)
                     {
-                        c.push_back({face, glyphs[i], positions[i]});
+                        glyphinfos[cluster_id] = cluster_glyphs;
                     }
                 }
                 bool all_set = true;
```

## Mongolian problems

Count: ~800

### Sample:

Label: `᠑᠐ ᠳᠦᠭᠡᠷ ᠪᠠᠭ`

Location in OSM: [way/443892079](https://www.openstreetmap.org/way/443892079)

Current rendering: 
Screenshot 2026-05-19 at 11 17 04 PM

Older rendering( both master~1 and #4556 ):
Screenshot 2026-05-19 at 11 17 45 PM

Current rendering matches browser rendering with the same font stack but older rendering matches browser rendering of the same label with a single "Noto Sans Mongolian Regular" font.

## Bengali problems

Count:  ~20

### Sample:

Label: `দুশান্‌বে`

Location in OSM: [node/27547977](https://www.openstreetmap.org/node/27547977)

Current rendering:
Screenshot 2026-05-19 at 11 48 48 PM

Older rendering( master~1 fc669e25500cb696689e4a819c17cd09ff1ffe5e ):
Screenshot 2026-05-19 at 11 50 08 PM

Older rendering( #4556 ): 
Screenshot 2026-05-19 at 11 53 17 PM

Current rendering matches browser rendering with the same font stack but older rendering with #4556 matches browser rendering of the same label with a single "Noto Sans Bengali Regular" font.


## Others

Count: ~1000

There were a few other differences in advances for CJK font rendered labels and labels rendered in other Indic scripts. None of these are visible to the naked eye but a pixel diff( which I did not do ) might have found them. But the above mentioned problems are the more serious ones. 

## Conclusion

Most of these problems can probably be traced to the fact that we are shaping uncovered ranges in the new approach as opposed shaping the whole text with each font in the older approach.

ramSeraph:
I can check if shaping full text instead of uncovered ranges and still handling coverage at UTF16 codepoint level, fixes these problems without causing more. 

But, in general I am slightly wary of this approach as that would be off the well-trodden path and we might be the first ones to try it. 

I am going to take a small break before coming back to this though. probably a week.

ramSeraph:
> I can check if shaping full text instead of uncovered ranges and still handling coverage at UTF16 codepoint level

The alternate implementation has been added at #4560 . I will add the results from using that approach in a bit.

ramSeraph:
All of the OSM planet dump's name, name:* values were again put through the font fallback code of #4560 and the current implementation for comparison( using OSM Carto's font stack )

And out of the approximately 45 million unique names.. around 1871 had different shaping results compared to the current implementation and out of these 1871, for around 1577 #4560 's results match #4556( as referred to in  https://github.com/mapnik/mapnik/pull/4557#issuecomment-4490854384 ) which means #4560 is correct for these entries. 

That left us with around 294 results with unexpected differences.. the summary of which follows.

## Mongolian Problems again( ~272 ):

Sample: `Зүүнхараа ᠵᠡᢉᠦᠨ ᠬᠠᠷᠠᠭ᠎ᠠ`

Source: [node/13513452878](https://www.openstreetmap.org/node/13513452878)

#4560 render:
Screenshot 2026-05-27 at 11 44 16 PM

Current render:
Screenshot 2026-05-27 at 11 44 43 PM

#4556 render ( the correct one ):
Screenshot 2026-05-27 at 11 45 07 PM

## Indic script problems( ~22 ):

Problems were seen across Devanagairi( Nepali, Marathi ) , Kannada, Malayalam, Khmer etc.  These I understand only happen around ZWNJ usage. Though it is unclear whether the usage is non-standard.  

Sample Kannada: `ಸೇಲಂ‌` 

Source: [node/13326713958](https://www.openstreetmap.org/node/13326713958)

#4560 render:
Screenshot 2026-05-27 at 11 57 36 PM

Current render and #4556 render:
Screenshot 2026-05-27 at 11 58 34 PM


All of these can be explained by the fact that the valid clusters of the fallback font even though they were able to cover the uncovered parts was dropped because of overlaps with already covered areas. I am mostly surprised there were so few of these. Also in general I am not comfortable with the #4560 implementation because of the non-standard nature of it and the fact that I am not sure that fallback cluster overlap is a rare occurrence. 

Coming back to the current implementation, looks like the Mongolian problem in the current PR which also affects Blink has been reported at [crbug.com/1499787](https://crbug.com/1499787) and https://github.com/harfbuzz/harfbuzz/issues/4503. 

And the [recommendation](https://github.com/harfbuzz/harfbuzz/pull/5969#issuecomment-4352553817) for fixing it is to treat things like spaces and [U+200C ZERO WIDTH NON-JOINER](https://unicode-explorer.com/c/200C), [U+180E MONGOLIAN VOWEL SEPARATOR](https://unicode-explorer.com/c/180E) etc when they are in a valid cluster returned by harfbuzz as non-covered unless they are accompanied by proper script characters. This was also the [recommended](https://github.com/harfbuzz/harfbuzz/issues/2889#issuecomment-792634559) solution from harfbuzz folks for handling the [U+3000 IDEOGRAPHIC SPACE](https://unicode-explorer.com/c/3000) problem mentioned in Blink edge cases in https://github.com/mapnik/mapnik/pull/4557#issuecomment-4461602450. Though Blink folks did not implement it that way.


Also worth looking at are these notes from someone who has been trying to fix the problem in Blink: https://resonant-cheesecake-be228e.netlify.app/mongolian-shaping-fix.html

ramSeraph:
So in general I am thinking of sticking to the current implementation and fixing the Mongolian/Ideographic space/ZWNJ problem in a seperate PR after this one lands.

ramSeraph:
If we are ok with taking the PR forward.. the changes are ready for review. I might do some more cleanup( especially assertions ) and some refactoring for readability but otherwise I am happy with the logic part. And I think I have a hang of how to handle the Mongolian problem with the current setup, I will add a PR for it on top of this.

PR #4561, opened 2026-05-28
Add !unbuffered_bbox!

leijurv:
When writing SQL in mapping styles, we don't know what our metatile size is or in other words we don't know how much buffer was applied. We can't hardcode it in the style, because these things are frequently (re)configured at the level of the tile renderer (mod_tile or some such).

This PR allows the SQL to see `!unbuffered_bbox!`, which is intended to be the metatile.

I think this will have lots of potential uses, but the one I have in mind can be found [here](https://redirect.github.com/openstreetmap-carto/openstreetmap-carto/issues/951#issuecomment-4526397288). In short, we would like to alter the geometry of various ways on-the-fly, to avoid segmented rendering. But, if we combine these geometries across metatile boundaries, then we will likely cause metatile boundary seams, because now each of the metatiles has a differing opinion on what is the length of this boundary-spanning way. In order to avoid this, **we want some kind of way to determine if a given way is completely within the metatile being rendered, excluding the buffer region.** Hence, `ST_Covers(!unbuffered_bbox!, way)` is the intended use case.

I hope you will consider this PR, it would be a great help to me 🙏

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ Developers: Please commit along with changes.
 
 For a complete change history, see the git log.
 
+## Unreleased
+
+- Add `!unbuffered_bbox!` SQL token to the `postgis` and `pgraster` plugins.
+  It expands to the metatile boundary — the same extent as `!bbox!` but
+  without `buffer-size` padding. Clipped to `!bbox!` so it is never larger.
+  Containment via `ST_Covers(!unbuffered_bbox!, way)` is guaranteed when
+  the map SRS and the layer SRS are the same.
+
 ## Mapnik 4.2.2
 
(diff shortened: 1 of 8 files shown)


imagico:
More general background:

Buffered rendering is essentially a fairly crude way to deal with the edges in rendering for tiled application where there are supposed to be no edges in the rendering results.  If you want to handle this problem more explicitly or in a more differentiated fashion from the map style side you will often need to know the unbuffered extent of the current tile.  Since this is known to Mapnik exposing that info on the SQL query level would be useful.

leijurv:
Ping, whenever you have time artemp 🙏

artemp (approved):

PR #4572, opened 2026-08-17
Enable NDEBUG by default

### LINKED ISSUE #4571: Performance regression in v4.0 with OpenStreetMap Carto?
Tile rendering on OpenStreetMap servers has gotten slower after an upgrade from Debian 12 (Mapnik v3.1) to Debian 13 (Mapnik v4.0), see [openstreetmap/operations#1375](https://github.com/openstreetmap/operations/issues/1375) (also [openstreetmap-carto#5276](https://github.com/openstreetmap-carto/openstreetmap-carto/issues/5276)).

A few sample numbers from just looking at the data of some monitoring graphs:
- ~46% increased low zoom (0-12) rendering job duration (92 min -> 134 min) on server culebre ([Time by Zoom](https://prometheus.openstreetmap.org/d/wyyzhZKMk/tile-rendering?orgId=1&from=2026-05-08T22:00:00.000Z&to=2026-05-10T02:00:00.000Z&timezone=utc&var-instance=culebre&viewPanel=panel-36)) ¹
- ~48% increased low zoom (0-12) rendering job duration (175 min -> 259 min) on server odin ([Time by Zoom](https://prometheus.openstreetmap.org/d/wyyzhZKMk/tile-rendering?orgId=1&from=2026-05-08T21:00:00.000Z&to=2026-05-10T02:00:00.000Z&timezone=utc&var-instance=odin&viewPanel=panel-36)) ¹
- ~129% increased [time per tile](https://prometheus.openstreetmap.org/d/wyyzhZKMk/tile-rendering?orgId=1&from=2026-04-18T09:25:00.000Z&to=2026-05-30T09:25:00.000Z&timezone=utc&var-instance=culebre&viewPanel=panel-49) for zoom 13 (5.9s -> 13.5s avg) on server culebre ²:
  screenshot time per tile, z13, culebre

As I understand, the regression in #4521 does not apply here, as `srs` is not an uppercase EPSG code, right?
  `srs: &srs "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over"`  ([openstreetmap-carto/project.mml](https://github.com/openstreetmap-carto/openstreetmap-carto/blob/v6.0.0/project.mml#L4))

As the Mapnik upgrade was part of a Debian upgrade I'm not really sure if this is actually a Mapnik regression, although it seems likely. Maybe you have some ways to reproduce this issue?

*[disclaimer: I'm just a messenger, neither a member of the OSMF Operations Working Group nor an openstreetmap-carto contributor]*

----
¹ *time between first entry for z0 and last entry for z12*
² *average of values before and after 2026-05-09 09:30*

leijurv:
Fixes #4571

**TLDR: A packaging issue has caused Debian to ship mapnik without `NDEBUG` causing `__attribute__((noinline))`, which causes a significant slowdown in tile rendering**

When openstreetmap.org updated their tileservers from Debian 12 to Debian 13, they went from Mapnik 3.1 to 4.0. This caused a slowdown, please see https://github.com/openstreetmap/operations/issues/1375 The cause of this is essentially that some flags fell through the cracks of the build system changing from SCons to CMake.

How to tell if a build is affected? Run: `nm -D --undefined-only libmapnik.so.* | grep -c assert_fail`

Debian 12: `nm -D --undefined-only libmapnik.so.3.1.0 | grep -c assert_fail` --> `0`
Debian 13: `nm -D --undefined-only libmapnik.so.4.0 | grep -c assert_fail` --> `1`

I tested this by rendering [this tile](https://tile.openstreetmap.org/10/175/408.png)

Debian 12: 25.4 seconds
Debian 13: 93.7 seconds
Debian 13 with mapnik rebuilt with -DNDEBUG added: 24.9 seconds

Because Debian 13 is stable, it will not take a new version, but, this is just a build flag fix so the maintainers might do it in-place (add `export DEB_CPPFLAGS_MAINT_APPEND = -DNDEBUG` to `debian/rules`)

Or maybe this could be published to `apt.openstreetmap.org`
```
apt-get build-dep -y mapnik                              
apt-get source mapnik                                    
cd mapnik-*/                                             
DEB_CPPFLAGS_APPEND=-DNDEBUG dpkg-buildpackage -B -uc -us
```

Affected:
- Debian 13
- Ubuntu 26.04
- Debian sid (for now)

Not affected (older mapnik prior to CMake packaging):
- Debian 12
- Ubuntu 24.04

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,7 @@ mapnik_option(USE_MULTITHREADED "enables the multithreaded features (threadsafe)
 mapnik_option(USE_NO_ATEXIT "disable atexit" OFF)
 mapnik_option(USE_NO_DLCLOSE "disable dlclose" OFF)
 mapnik_option(USE_DEBUG_OUTPUT "enables some debug messages for development" OFF)
+mapnik_option(USE_NDEBUG "define NDEBUG for non-Debug builds. Turning this off makes mapbox::variant dispatch non-inlinable and is drastically slower - see below." ON)
 mapnik_option(USE_LOG "enables logging output. See log severity level." OFF)
 # 0 = debug
 # 1 = warn
@@ -293,6 +294,43 @@ endif()
 
 if(USE_DEBUG_OUTPUT)
     list(APPEND MAPNIK_COMPILE_DEFS MAPNIK_DEBUG)
+endif()
+
+# NDEBUG does much more for mapnik than disable assert(). mapbox::variant annotates
+# every visit()/dispatcher::apply()/recursive_wrapper::get() with VARIANT_INLINE, which
+# expands to __attribute__((noinline)) when NDEBUG is *not* defined, and to nothing when
+# it is. mapnik's expression AST, mapnik::value and all geometry visitation go through
+# mapbox::variant, so an optimised build without NDEBUG cannot inline its own hottest
+# dispatch path: ~6x slower rule-filter evaluation, and ~3.7x slower on a real
+# openstreetmap-carto z10 tile (25s -> 94s).
+#
+# CMake only supplies -DNDEBUG via CMAKE_<LANG>_FLAGS_<CONFIG>, i.e. for the Release,
+# RelWithDebInfo and MinSizeRel configurations. A single-config build that leaves
+# CMAKE_BUILD_TYPE empty or pins it to None gets no NDEBUG at all - which is exactly
+# what Debian's `dh --buildsystem=cmake` does (it configures -DCMAKE_BUILD_TYPE=None so
+# that dpkg-buildflags owns the compiler flags, and dpkg-buildflags supplies -O2 but
+# never NDEBUG). So define it here rather than relying on the configuration name.
+#
+# This lands in MAPNIK_COMPILE_DEFS, which is also what libmapnik.pc advertises, so
+# consumers compiling their own translation units against mapnik's headers inherit it
+# too - they hit the same VARIANT_INLINE penalty otherwise.
+get_property(MAPNIK_IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+string(TOUPPER "${CMAKE_BUILD_TYPE}" MAPNIK_BUILD_TYPE_UPPER)
+if(MAPNIK_IS_MULTI_CONFIG)
+    # Config is chosen at build time; CMAKE_<LANG>_FLAGS_<CONFIG> already handles NDEBUG
+    # per configuration, and CMAKE_BUILD_TYPE is meaningless here.
+elseif(MAPNIK_BUILD_TYPE_UPPER STREQUAL "DEBUG")
+    message(STATUS "Debug build: leaving NDEBUG undefined (assertions on, mapbox::variant dispatch not inlined)")
+elseif(USE_NDEBUG)
+    list(APPEND MAPNIK_COMPILE_DEFS NDEBUG)
+else()
+    message(WARNING
+        "USE_NDEBUG is OFF for a non-Debug build (CMAKE_BUILD_TYPE='${CMAKE_BUILD_TYPE}'). "
+        "Unless -DNDEBUG is supplied some other way, mapbox::variant's VARIANT_INLINE becomes "
+        "__attribute__((noinline)) and libmapnik will be several times slower on any "
+        "expression-heavy style, regardless of the optimisation level. Packagers: append "
+        "-DNDEBUG to CPPFLAGS, or leave USE_NDEBUG ON."
+    )
 endif()
 
 if(USE_LOG)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,7 +57,7 @@ mapnik_option(USE_MULTITHREADED "enables the multithreaded features (threadsafe)
 mapnik_option(USE_NO_ATEXIT "disable atexit" OFF)
 mapnik_option(USE_NO_DLCLOSE "disable dlclose" OFF)
 mapnik_option(USE_DEBUG_OUTPUT "enables some debug messages for development" OFF)
-mapnik_option(USE_NDEBUG "define NDEBUG for non-Debug builds. Turning this off makes mapbox::variant dispatch non-inlinable and is drastically slower - see below." ON)
+mapnik_option(USE_NDEBUG "defines NDEBUG for non-Debug builds. Turning this off makes mapbox::variant dispatch non-inlinable and much slower" ON)
 mapnik_option(USE_LOG "enables logging output. See log severity level." OFF)
 # 0 = debug
 # 1 = warn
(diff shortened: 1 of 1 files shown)


tomhughes:
Is it really a packaging issue? If it was then surely the fix would be downstream not here...

Looking at this I think (speaking as the Fedora packager for mapnik) that this is going to affect more than just Debian?

tomhughes:
Ah the `%cmake_build` macro that I use for Fedora already defines `-DNDEBUG` using:

```
-DCMAKE_C_FLAGS_RELEASE:STRING="-DNDEBUG"
-DCMAKE_CXX_FLAGS_RELEASE:STRING="-DNDEBUG"
```

leijurv:
Perhaps I was wrong to say packaging issue, sorry, imprecise of me. I just mean that `-DNDEBUG` is not being passed, which is causing debug asserts and noinline to be applied to release builds.

Yes Fedora 44 is also affected but for a different reason than Debian. I focused on Debian because that is what surfaced the issue. Fedora sets `CMAKE_CXX_FLAGS_RELEASE="-DNDEBUG"` however `CMAKE_BUILD_TYPE` is not set.

"the fix would be downstream not here" perhaps, but, mapnik used to set this explicitly by default in the SCons era. That's why I wrote above that this `fell through the cracks of the build system changing from SCons to CMake`. See: [here](https://github.com/mapnik/mapnik/blob/8b70a205c8f751bd880d890148f17bc1cdab8075/SConstruct#L302) and [here](https://github.com/mapnik/mapnik/blob/8b70a205c8f751bd880d890148f17bc1cdab8075/SConstruct#L1910) and [here](https://github.com/mapnik/mapnik/blob/8b70a205c8f751bd880d890148f17bc1cdab8075/SConstruct#L1951-L1955).

leijurv:
To see for yourself:

`docker run --rm fedora:44 sh -c 'dnf -y install mapnik binutils >/dev/null 2>&1; L=$(rpm -ql mapnik | grep -m1 libmapnik.so); echo "$L"; nm -D --undefined-only $L | grep __assert_fail; strings $L | grep mapbox/'`

and you will see it gives:

```
/usr/lib64/libmapnik.so.4.2
                 U __assert_fail@GLIBC_2.2.5
/usr/include/mapbox/variant.hpp
/usr/include/mapbox/recursive_wrapper.hpp
```

As versus on, say, Ubuntu 24.04:

`docker run --rm ubuntu:24.04 sh -c 'apt-get update -qq >/dev/null 2>&1; DEBIAN_FRONTEND=noninteractive apt-get install -y -qq libmapnik-dev binutils >/dev/null 2>&1; L=$(find / -name "libmapnik.so.*" -not -type l | head -1); echo "$L"; nm -D  --undefined-only $L | grep __assert_fail; strings $L | grep mapbox/'` just prints `/usr/lib/libmapnik.so.3.1.0`

To see the build type is not set in mapnik.spec:

`docker run --rm fedora:44 sh -c 'dnf -y install cmake-rpm-macros redhat-rpm-config >/dev/null 2>&1; rpm --eval %cmake | grep -c CMAKE_BUILD_TYPE'` prints `0`.

`docker run --rm fedora:44 sh -c 'dnf -y install cmake gcc-c++ >/dev/null 2>&1; mkdir /t; cd /t; { echo "cmake_minimum_required(VERSION 3.20)"; echo "project(t CXX)"; echo "add_library(t t.cpp)"; } > CMakeLists.txt; echo "int f(){return 0;}" > t.cpp; for bt in "" "-DCMAKE_BUILD_TYPE=Release"; do rm -rf b; cmake -S . -B b -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_EXPORT_COMPILE_COMMANDS=ON $bt >/dev/null 2>&1; echo "build type [${bt:-none}] -> NDEBUG on compile line: $(grep -c -- -DNDEBUG b/compile_commands.json)"; done'` prints:

```
build type [none] -> NDEBUG on compile line: 0
build type [-DCMAKE_BUILD_TYPE=Release] -> NDEBUG on compile line: 1
```

joto:
This PR seems rather complicated. Generally CMake will manage the `(N)DEBUG` flag for you depending on the `CMAKE_BUILD_TYPE`. Unfortunately the default for that seems to be the empty string which translates to the `Debug` setting. I have a setting in `CMakeLists.txt` in all my projects that sets `CMAKE_BUILD_TYPE=ReleaseWithDebInfo` if it is unset to counter that. Using the CMake built-in mechanism seems to me is a better solution than your complex code.

I had always assumed that Debian and other distributions simply set `CMAKE_BUILD_TYPE` correctly, although I have never checked that and I can't find anything in the [code](https://github.com/Debian/debhelper/blob/master/lib/Debian/Debhelper/Buildsystem/cmake.pm). Either I am looking in the wrong place or my assumption was wrong.

leijurv:
That is fair. My apologies. I wasn't even really sure where to make a PR - here, or try to talk to distro maintainers. But I made the PR here because this repo used to default to release mode (NDEBUG) but now it does not. I am not sure if it's fair to say that is to "blame" because I am very unfamiliar with packaging in general.

tomhughes:
I certainly thought that the build type defaulted to RelWithDebInfo either because that was the cmake default or because the Fedora macros did that but it seems not.

tomhughes:
So is the mapnik setup managing to ignore that C flag that Fedora sets and substituting it's own NDEBUG value based on the build type?

leijurv:
> I thought the build type defaulted to RelWithDebInfo

I think in CMake`CMAKE_BUILD_TYPE` defaults to empty string, and Fedora never sets it to anything else. That's what I was demonstrating [above](https://github.com/mapnik/mapnik/pull/4572#issuecomment-5319994092), showing that `rpm --eval %cmake | grep -c CMAKE_BUILD_TYPE` gives `0`.

> Is mapnik ignoring that C flag and substituting its own NDEBUG based on the build type?

Nothing in mapnik reads, overrides or substitutes anything here. Prior to this PR, mapnik's CMake did nothing with NDEBUG. It's not receiving any NDEBUG flag *because* the CMAKE_BUILD_TYPE is *not* set. `CMAKE_CXX_FLAGS_RELEASE` is only appended to the compile line when the configuration is `Release`, which it isn't.

leijurv:
> the empty string which translates to the Debug setting

Unfortunately, the empty string seems to result in no per-config flags (like `-g` `-O0` `-DNDEBUG`). 

And in the file you linked, actually see [here](https://github.com/Debian/debhelper/blob/master/lib/Debian/Debhelper/Buildsystem/cmake.pm#L16) it does `-DCMAKE_BUILD_TYPE=None`. So your approach would only fix Fedora not Debian.

For the complication concern, I have pushed a new commit.

tomhughes:
I've rebuilt the Fedora packages with `-DCMAKE_BUILD_TYPE=Release` now.

joto:
You can see the effect on the OSMF tile servers [here](https://prometheus.openstreetmap.org/d/wyyzhZKMk/tile-rendering?orgId=1&from=now-30d&to=now&timezone=utc&var-instance=culebre&var-instance=nidhogg&var-instance=odin&var-instance=ysera&viewPanel=panel-28&refresh=1m).

There is a distinctive pattern of overloaded servers on weekdays, after the switch to the build with `-DNDEBUG` on Wednesday this was fixed.

\cc @sebastic for Debian packaging

sebastic:
> \cc @sebastic for Debian packaging

https://salsa.debian.org/debian-gis-team/mapnik/-/commit/8016a39c6c762364e58497e71dd6a323552915f6

artemp (approved):

sebastic:
This leaks private compile flags in the pkg-config file as reported by lintian
```
E: libmapnik-dev: pkg-config-bad-directive -DNDEBUG [usr/lib/x86_64-linux-gnu/pkgconfig/libmapnik.pc]
```

Either the flag needs to be stripped from `m_compile_defs` in `cmake/MapnikExportPkgConfig.cmake`, or consider reverting this change and let users set `CMAKE_BUILD_TYPE`.

artemp:
I'm in favour of reverting and using `CMAKE_BUILD_TYPE`  
On balance seems best approach here ? 
/cc @leijurv @tomhughes @sebastic @joto

leijurv:
My apologies!

Reverting would be fine for Debian and Fedora users (so, fine for tile.osm.org). But other users would still have the issue of NDEBUG off by default, when it was on by default back with SCons. So every other distro/packager would also need to know to set `CMAKE_BUILD_TYPE`. Maybe that's reasonable, if so, go ahead and revert. But we can actually preserve the faster default without exporting the flag like this. The following diff should work I think:


```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 913984bef5..0000000000 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,14 +302,11 @@ if(USE_DEBUG_OUTPUT)
 # and all geometry visitation go through mapbox::variant, so an optimised build without NDEBUG
 # cannot inline its own hottest dispatch path (ref #4571).
 #
-# CMake only supplies -DNDEBUG via CMAKE_<LANG>_FLAGS_<CONFIG>, so a build that leaves
-# CMAKE_BUILD_TYPE empty or pins it to None - as Debian's `dh --buildsystem=cmake` does - never
-# gets it. Define it here instead of relying on the configuration name. Multi-config generators
-# pick the configuration at build time, where CMAKE_<LANG>_FLAGS_<CONFIG> already handles this.
-get_property(MAPNIK_IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
-string(TOUPPER "${CMAKE_BUILD_TYPE}" MAPNIK_BUILD_TYPE_UPPER)
-if(USE_NDEBUG AND NOT MAPNIK_IS_MULTI_CONFIG AND NOT MAPNIK_BUILD_TYPE_UPPER STREQUAL "DEBUG")
-    list(APPEND MAPNIK_COMPILE_DEFS NDEBUG)
+# Preserve the historical SCons default for Mapnik's own build without making
+# NDEBUG a usage requirement for downstream consumers. Directory compile
+# definitions apply to Mapnik's targets and subdirectories, but are not
+# exported through the installed CMake targets or pkg-config file.
+if(USE_NDEBUG)
+    add_compile_definitions("$<$<NOT:$<CONFIG:Debug>>:NDEBUG>")
 endif()

 if(USE_LOG)
```

sebastic:
That is definitively a better implementation than what's in 4.3.1.

leijurv:
Sorry 🙏

leijurv:
#4577

joto:
As mentioned above using `CMAKE_BUILD_TYPE` and also setting that in the CMake config to a reasonable default has worked for my software for years and never created any problems with Debian or generated any other complaints. So I'd recommend that.

PR #4573, opened 2026-08-21
Performance improvements

leijurv:
Hello, I would like to share some performance improvements to Mapnik.

The motivation was profiling OSM Carto, for which this PR delivers significant speedups to the mid zoom levels.

Please see the following table of before/after Mapnik CPU seconds (not wall-clock time, so we exclude postgres, because that is controlled by the style's queries & indexes). This is measured as 256x256 tiles with 128px buffer, OSM Carto style [v6.0.0](https://github.com/openstreetmap-carto/openstreetmap-carto/releases/tag/v6.0.0), centered on [this](https://www.openstreetmap.org/#map=19/34.056089/-118.234794) coordinate (same as I measured in #4572).

Particularly note the rows around Z10. This was primarily caused by a significant number of `planet_osm_point` features that are style filtered out (Z10 is the first zoom level with POIs (e.g. airports)).

|Z|Before|After|
|---|---|---|
| 8| 0.946|0.525|
| 9| 0.887|0.556|
|<b>10</b>|<b>11.185</b>|<b>1.522</b>|
|11| 5.507|0.886|
|12| 2.394|0.484|
|13| 1.093|0.392|
|14| 0.390|0.221|
|15| 0.205|0.138|
|16| 0.125|0.099|
|17| 0.102|0.088|
|18| 0.076|0.070|
|19| 0.067|0.064|
|20| 0.068|0.067|

Explanation of the changes, recommend to read one at a time. (I could instead make 7 separate PRs if you want to pick and choose)

|Commit|Z10 time|Incremental reduction|Cumulative reduction|Explanation|
|---|---|---|---|---|
|Master   |11.185| | |Nothing changed yet baseline speed|
|https://github.com/leijurv/mapnik/commit/e1e6d31808877171d14c74d5974b923cce423595|9.400|16.0%|16.0%|Expression evaluation `mapnik::value`s were passed by value for most operations. Where possible, instead borrow and compare by reference|
|https://github.com/leijurv/mapnik/commit/d0ed0964ad9d6c23cee6171656b002d1304fff40|7.653|18.6%|31.6%|Template specialize string comparison to short-circuit so that it does not become a `mapnik::value`|
|https://github.com/leijurv/mapnik/commit/d744e094b451a12b070173547f638b896d1386d8|6.520|14.8%|41.7%|Feature value position lookups are faster using a new `std::unordered_map`|
|https://github.com/leijurv/mapnik/commit/e981e4b43ed57aa209175626777735036f97cc88|5.905| 9.4%|47.2%|Reorder expression nodes so that the most common ones are at the top, which is faster since they are tested in declaration order|
|https://github.com/leijurv/mapnik/commit/d59329887dc708612a670117e562070f15634b63|3.951|33.1%|64.7%|A fast path for evaluating style filters as booleans rather than `mapnik::value`|
|https://github.com/leijurv/mapnik/commit/508dd835252b5a523bffef076b6779b0c21e4588|1.572|60.2%|85.9%|An index to lookup rules if they follow the `[attribute] = 'literal'` pattern|
|https://github.com/leijurv/mapnik/commit/40c36ee0125a7056af617f607bda61e9eb21bf08|1.524| 3.1%|86.4%|Store PostGIS attributes keyed by index rather than string name|


include/mapnik/expression_evaluator.hpp
```diff
@@ -74,32 +175,83 @@ struct evaluate
...
+                  x.right.template get_unchecked<value_unicode_string>());
+            }
+            if (rhs && !lhs && rhs->template is<value_unicode_string>() && x.left.template is<value_unicode_string>())
+            {
+                return detail::ustring_compare<Tag>::apply(x.left.template get_unchecked<value_unicode_string>(),
+                                                           rhs->template get_unchecked<value_unicode_string>());
+            }
+        }
+
+        if (lhs)
```
babbaj, line 208: can be simplified to
```
operation(
    lhs ? util::apply_visitor(*this, x.left) : *lhs,
    rhs ? util::apply_visitor(*this, x.right) : *rhs,
);
```
but code gen may be different

include/mapnik/expression_evaluator.hpp
```diff
@@ -74,32 +175,83 @@ struct evaluate
...
+        if constexpr (detail::is_comparison_tag<Tag>::value)
+        {
+            return value_type(compare(x));
+        }
+        else
+        {
+            typename make_op<Tag>::type operation;
+            value_type const* lhs = borrow(x.left);
+            value_type const* rhs = borrow(x.right);
+            if (lhs)
```
babbaj, line 231: same as above

include/mapnik/expression_evaluator.hpp
```diff
@@ -45,6 +121,31 @@ struct evaluate
           vars_(v)
     {}
 
+    using attribute_reference = decltype(std::declval<feature_type const&>().get(std::declval<std::string const&>()));
```
babbaj, line 124: this is cute but std::result_of exists

include/mapnik/feature.hpp
```diff
@@ -125,23 +146,35 @@ class MAPNIK_DECL feature_impl : private util::noncopyable
...
 
+    inline void put(std::size_t index, value&& val)
+    {
+        if (index < data_.size())
+        {
+            data_[index] = std::move(val);
+        }
+        else
+        {
+            throw std::out_of_range("Attribute index does not exist");
```
babbaj: Error messages may potentially be worse now that the key itself is no longer in the exception message

include/mapnik/expression_evaluator.hpp
```diff
@@ -74,32 +175,83 @@ struct evaluate
...
+
+        if constexpr (detail::ustring_compare<Tag>::enabled)
+        {
+            if (lhs && !rhs && x.right.template is<value_unicode_string>())
+            {
+                return detail::ustring_compare<Tag>::apply_value(
+                  *lhs,
+                  x.right.template get_unchecked<value_unicode_string>());
+            }
+            if (rhs && !lhs && rhs->template is<value_unicode_string>() && x.left.template is<value_unicode_string>())
```
babbaj, line 201: Why is this branch different? the branch above will use `*lhs` if it's not a nullptr but this one always calls get_unchecked on `rhs` even though it checked that `rhs` is not a nullptr

include/mapnik/rule_cache.hpp
```diff
@@ -79,10 +149,16 @@ class rule_cache : private util::noncopyable
 
     rule_ptrs const& get_also_rules() const { return also_rules_; }
 
+    rule_indices const& get_rules_without_precondition() const { return rules_without_precondition_; }
+
+    std::vector<precondition_group> const& get_precondition_groups() const { return precondition_groups_; }
```
babbaj: std::span should be preferred here

artemp:
@leijurv - great stuff!

artemp (approved):

PR #4582, opened 2026-09-08
Cache rasterized SVG polygon patterns

leijurv:
Polygon patterns for landcover/landuse are used in OSM Carto. I wanted to optimize them so that we could fully move from PNG patterns to SVG patterns, so that HiDPI rendering becomes more consistent (see: https://github.com/openstreetmap-carto/openstreetmap-carto/pull/5302). However, I found that for some of the patterns it would cause a significant performance regression. The reason is that some of them, in particular rock https://github.com/openstreetmap-carto/openstreetmap-carto/blob/master/symbols/rock_overlay.png, are rather heavy as SVGs. The issue is that mapnik is currently re-rasterizing the pattern SVGs for each polygon in the metatile. The XML parsing is cached, but the actual drawing could be cached too since it gets path clipped from the overall pattern per polygon (for pattern consistency across landuse/landcover boundaries I imagine?). Therefore, this can be optimized with a per-thread LRU cache, where the key is the SVG source identity and the final affine transform. This will use up to 256kb per pattern per thread, up to 16mb total per thread if there are many distinct tiles. The key in that case will always be `(1,0,0,1,0,0)`. For retina / 2x rendering, it will be `(2,0,0,2,0,0)` and memory usage is up to 64mb per thread. Note that opacity, gamma, and compositing is unaffected (they are further downstream). OSM Carto has 41 polygon patterns, of which currently 27 are SVG and 14 are PNG, so in the tile.osm.org case, the RAM usage increase is a little under ten megabytes (even if all remaining PNGs are converted to SVG patterns). In my prior performance PRs I had picked urban areas which have some land cover patterns but not that many, leaving this change having no measurable effect except by chance that the Z13 selection had a lot of landcover pattern polygons. So I have changed the measurement area here to be two locations with such landcover, while still being urban. Measurements still taken mostly the same as #4578 (as similar to tile.osm.org as I can get, Mapnik CPU seconds, average of 5 runs, etc). Starting at Z13 since that's when the patterns appear. The only difference from #4578 methodology is that I am now measuring the tiles warm with mapnik having rendered tiles beforehand in the same process. This is more faithful to tile.osm.org mod_tile renderd processes, and it also helps in this case because of the caching. Note that the before and after numbers below are both warm of course.

As you can see, the speedup for such metatiles is realistic and nontrivial. "Why is London Z13 sped up so much???" Because it has **2,731 distinct forest pattern polygons**, all of which previously caused the entire forest pattern to be redrawn. Now the forest pattern is only drawn once and cached as a bitmap. The resulting PNGs are identical.

|Zoom|London|NYC|
|---|---|---|
|13|<b>−29.77%</b> (2.073s → 1.456s)|−16.78% (2.831s → 2.356s)|
|14|−19.26% (1.790s → 1.445s)|−5.65% (2.294s → 2.165s)|
|15|−14.51% (1.335s → 1.142s)|−3.84% (1.908s → 1.835s)|
|16|−14.69% (0.821s → 0.700s)|−6.32% (1.051s → 0.985s)|
|17|−8.92% (0.420s → 0.383s)|−2.36% (0.792s → 0.773s)|
|18|−8.77% (0.337s → 0.308s)|−3.48% (0.388s → 0.374s)|
|19|−6.54% (0.241s → 0.225s)|−8.95% (0.229s → 0.208s)|

<details>
<summary><b>See the metatiles in question (click to expand)</b></summary>

london_z13
london_z14
london_z15
london_z16
london_z17
london_z18
london_z19

nyc_z13
nyc_z14
nyc_z15
nyc_z16
nyc_z17
nyc_z18
nyc_z19
</details>

### DIFF
--- a/include/mapnik/renderer_common/render_pattern.hpp
+++ b/include/mapnik/renderer_common/render_pattern.hpp
@@ -40,6 +40,11 @@ struct marker_svg;
 template<typename T>
 void render_pattern(marker_svg const& marker, agg::trans_affine const& tr, double opacity, T& image);
 
+// SVG pattern pixels depend on the source and evaluated image transform, not
+// on the polygon that will be filled with them. Images use opacity 1 here;
+// polygon opacity and compositing are applied afterward.
+std::shared_ptr<image_rgba8 const> rasterized_pattern(marker_svg const& marker, agg::trans_affine const& tr);
+
 } // namespace mapnik
 
 #endif // MAPNIK_RENDER_PATTERN_HPP
--- a/src/agg/process_polygon_pattern_symbolizer.cpp
+++ b/src/agg/process_polygon_pattern_symbolizer.cpp
@@ -86,10 +86,8 @@ struct agg_renderer_process_visitor_p
         auto image_transform = get_optional<transform_type>(sym_, keys::image_transform);
         if (image_transform)
             evaluate_transform(image_tr, feature_, common_.vars_, *image_transform, common_.scale_factor_);
-        mapnik::box2d<double> const& bbox_image = marker.get_data()->bounding_box() * image_tr;
-        mapnik::image_rgba8 image(bbox_image.width(), bbox_image.height());
-        render_pattern<buffer_type>(marker, image_tr, 1.0, image);
-        render(image);
+        auto image = rasterized_pattern(marker, image_tr);
+        render(*image);
     }
 
     void operator()(marker_rgba8 const& marker) const { render(marker.get_data()); }
--- a/src/renderer_common/render_pattern.cpp
+++ b/src/renderer_common/render_pattern.cpp
@@ -19,6 +19,9 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
  *
  *****************************************************************************/
+// stl
+#include <array>
+#include <list>
 // mapnik
 #include <mapnik/renderer_common/render_pattern.hpp>
 #include <mapnik/geometry/box2d.hpp>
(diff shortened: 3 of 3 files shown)


artemp:
@leijurv - 👍  
what are your thoughts about implementing similar caching logic for SVG markers?

artemp (approved):

leijurv:
@artemp It's interesting actually! In short, yes you can do it, but it's awkward in the code and it's a very small win, maybe 1%, for dense POI tiles. The pattern polygons by comparison were very simple and controlled in how they're drawn.

However, I was unable to get the metatiles to be exactly equal to master. The reason turns out to be that pixels on the exact edge of the metatile would vary by sometimes a tiny amount (plus or minus 1).

https://github.com/leijurv/mapnik/commit/678178eac42ffad0e4373fbab641989fb5edec7d

This is due to a few different reasons, which are essentially that drawing off the edge of the metatile is very very similar but sometimes has subpixel differences to drawing at full size then cropping. (and that cache commit is going to always draw at full size then translate+crop+over). This happens in all directions, but a bit more in the negative direction because once the pixel coordinate goes negative, rounding also flips direction.

I also have a separate additional commit that would address those differences. In short it makes it so that no matter where you draw a SVG, whether the coords are positive or negative, whether they are partially in the metatile or fully in the metatile, you get an identical draw. Translation invariant. However, this negates most of the performance win of the prior commit unfortunately, takes it down to maybe 0.5% win at best for high zoom dense areas, a small loss in lower zoom: https://github.com/leijurv/mapnik/commit/94c568cdd9c79ecfab169e658c0eb8a6bd93d9c7

But it's unclear if that 2nd commit is really warranted, because arguably the 1st commit is more correct than master, not less correct? The 2nd commit makes the translation invariance global, to all SVGs (even ones that aren't cacheable). I'm really not sure and it's rather complicated and the win is unclear.

artemp:
@leijurv - 👍 for detailed response ^ 

> maybe 1%, for dense POI tiles 

I think the gain could be much greater for more complex SVG markers and also more "marker intensive" styles.. but needs testing. I'll have more time to look into all this in a week or so. Thanks again for your efforts!

